PR #10, opened 2016-10-27
new String extension added

h44f33z:
Added new extension for String to convert to Date based on entered format. Refer below for example use

```swift
let dateString = "2016/10/20"
let format = "yyyy/MM/dd"
        
let date = dateString.toDate(withFormat: format)
        
print("The date is ---- \(date)")
```


omaralbeik:
@h44f33z Thank you for your contribution, I updated your function to return optional value.

h44f33z:
Sure. You library very useful. Thanks

PR #16, opened 2016-12-11
use of reduce for remove duplicates in array

sairamkotha:
Reduce can be used for remove duplicates in array rather than use of for loop in array. 
Thanks for the beautiful extensions.


omaralbeik:
@sairamkotha Updated in #v1.3.1, Thank you!

sairamkotha:
Welcome :)

PR #71, opened 2017-01-05
Add String slice method and tests

pawurb:
Hello. I am a Ruby developer playing around with Swift for a while now. One of the things I keep missing is **String#slice** method from Ruby stdlib. This is implementation that replicates the behaviour of a Ruby **String#slice** for integer argument types https://ruby-doc.org/core-2.2.0/String.html#method-i-slice . Kind of bombarded it with tests but wanted to cover all the edge cases. Hope that can be useful.

## Checklist

- [x] New extensions were created in **development branch**.
- [ ] Pull request was created to **development head branch**.

### DIFF
--- a/Source/Extensions/StringExtensions.swift
+++ b/Source/Extensions/StringExtensions.swift
@@ -410,9 +410,11 @@ public extension String {
 	///   - length: Amount of characters to be sliced after given index.
 	/// - Returns: Sliced substring of length number of characters (if applicable) (example: "Hello world".slicing(from: 6, length: 5) -> "world")
 	public func slicing(from i: Int, length: Int) -> String? {
-		guard length >= 0, i >= 0, i < characters.count,
-			i.advanced(by: length) <= characters.count  else {
-				return nil
+		guard length >= 0, i >= 0, i < characters.count  else {
+			return nil
+		}
+		guard i.advanced(by: length) <= characters.count else {
+			return slice(at: i)
 		}
 		guard length > 0 else {
 			return ""
--- a/Source/Extensions/StringExtensions.swift
+++ b/Source/Extensions/StringExtensions.swift
@@ -414,7 +414,7 @@ public extension String {
 			return nil
 		}
 		guard i.advanced(by: length) <= characters.count else {
-			return slice(at: i)
+			return slicing(at: i)
 		}
 		guard length > 0 else {
 			return ""
--- a/Source/Extensions/StringExtensions.swift
+++ b/Source/Extensions/StringExtensions.swift
@@ -402,6 +402,61 @@ public extension String {
 	public mutating func reverse() {
 		self = String(characters.reversed())
 	}
+
+	/// SwifterSwift: Slice string.
+	/// - Parameters:
+	///   - at: string index the slicing should start from.
+	///   - length: amount of characters to be sliced.
+	/// - Returns: sliced substring of length number of characters or nil
+
+	public func slice(at: Int, length: Int) -> String? {
+		if length < 0 {
+			return nil
+		}
+
+		if length == 0 {
+			if at > characters.count {
+				return nil
+			} else {
+				return ""
+			}
+		}
+
+		if at < 0  {
+			let fromIndex = index(endIndex, offsetBy: at)
+			let toIndex = index(fromIndex, offsetBy: (length-1))
+
+			return self[fromIndex...toIndex]
+		} else {
+			let fromIndex = index(startIndex, offsetBy: at)
+			let toIndex = index(startIndex, offsetBy: (length+at-1))
+
+			return self[fromIndex...toIndex]
+		}
+	}
+
+	/// SwifterSwift: Slice string.
+	/// - Parameters:
+	///   - at: index of the character that should be sliced
+	/// - Returns: returns a substring of a character at a given index or nil
+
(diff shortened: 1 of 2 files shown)


muescha:
with the `slice(at: int)` i get this result:

```swift
let str1 = "The morning is upon us."
print(str1.slice(at: 12))  // -> Optional("i")
```

this makes sense because the `slice` is also notated in brackets `[]` in ruby:

 ```ruby
"abcd"[3] #-> 'd'
```

but somehow in swift i expect the `slice` more like the javascript slice (see [Javascript MDN | String.prototype.slice](https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/String/slice) ):

```swift
let str1 = "The morning is upon us."
print(str1.slice(at: 12))  // -> Optional("is upon us.")
```

pawurb:
If I change behaviour of only one argument function call, then it will not be compatible with neither JS nor Ruby implementation. Do you suggest going all for JS version then? Obviously I am biased but I prefer the Ruby one.

muescha:
1) the 1 argument call can also done with the `"abced"[3]` call from SwifterSwift
2) i think the 2 argument call is now already JS compatible 

@omaralbeik what is your favorite?

omaralbeik:
I was trying to avoid adding slice method, because it is implemented differently in different programming languages, however I think there is no escape from it :)

I refactored @pawurb code to the following
```swift
public func slice(from index: Int, length: Int) -> String? {
	guard length >= 0, index >= 0, index < characters.count,
		index.advanced(by: length) < characters.count  else {
		return nil
	}
	guard length > 0 else {
		return ""
	}
	let start = self.index(startIndex, offsetBy: index)
	let end = self.index(startIndex, offsetBy: index + length)
	return self[start...end]
}

public func slice(from startIndex: Int, to endIndex: Int) -> String? {
	return slice(from: startIndex, length: endIndex - startIndex)
}

public func slice(at index: Int) -> String? {
	return slice(from: index, to: characters.count - 1)
}
```

Example:
```swift
let s = "Hello world"
s.slice(from: 6, length: 4) // -> "world"
s.slice(from: 3, to: 10) // -> "lo world"
s.slice(at: 6) // -> "world"
```


@pawurb @muescha Your feedback is important,  what do you think of the above code, does it make sense to you?

pawurb:
By calling the function parameter `index` you are shadowing internal String method `index`. Looks a bit confusing, even if you use `self.index` to call it later on. The same goes for `startIndex` and `endIndex` that are also native String methods.

Also: `s.slice(from: 6, length: 4)` returns the string that is 5 characters long. It is not compatible with neither Ruby nor JS implementation. I think you should change:

`let end = self.index(startIndex, offsetBy: index + length)` to
`let end = self.index(startIndex, offsetBy: index + length - 1)`

Apart from it looks ok.

muescha:
thx for the hint about shadowing 

to avoid confusion we can rename the parameter to `sliceStartIndex` and `sliceEndIndex`

```
return slice(from: startIndex, length: endIndex - startIndex)
```



this can be done by:
```
return self[sliceStartIndex... sliceEndIndex]
```


and 
```
return slice(from: index, to: characters.count - 1)
```

can use the internal endIndex:
```
return self[sliceIndex...endIndex]
````

**Comment by omaralbeik** [comment]
@pawurb @muescha How about now

```swift
public func slice(from i: Int, length: Int) -> String? {
	guard length >= 0, i >= 0, i < characters.count,
		i.advanced(by: length) <= characters.count  else {
		return nil
	}
	guard length > 0 else {
		return ""
	}
	return self[i..<i.advanced(by: length)]
}

public func slice(from start: Int, to end: Int) -> String? {
	return self[start..<end]
}

public func slice(at i: Int) -> String? {
	guard i < characters.count else {
		return nil
	}
	return self[i..<characters.count]
}
```

Example
```swift
let s = "Hello world"
s.slice(from: 6, length: 5) // -> "world"
s.slice(from: 6, to: 11) // -> "world"
s.slice(at: 6) // -> "world"
```

pawurb:
👍 as long as the specs are green ;)

muescha:
1) i think you can safe use `endIndex` instead of `characters.count` because its an native method and maybe faster 
2) i would expect in the case of `i.advanced(by: length) <= characters.count` then it should be all until the end:
> If intEnd < 0, let to be max(len + intEnd,0); otherwise let to be min(intEnd, len)

https://tc39.github.io/ecma262/#sec-string.prototype.slice

muescha:
@pawurb reading the specs i see what you mean with the difference to ruby and JS. in javascript it is `slice(start[,end]) `

omaralbeik:
@muescha 
1. endIndex is not of type Int it is of type (String.Index), I was not even able to cast it to Int
2. correct, how about this this:

```swift
public func slice(from i: Int, length: Int) -> String? {
	guard length >= 0, i >= 0, i < characters.count  else {
		return nil
	}
	guard i.advanced(by: length) <= characters.count else {
		return slice(at: i)
	}
	guard length > 0 else {
		return ""
	}
	return self[i..<i.advanced(by: length)]
}
```

muescha:
thats ok

the edge cases should also go into the unit tests :)

omaralbeik:
@pawurb Please re-create this request from development head branch to merge it.

Issue #81, opened 2017-01-09
watchOS support


omaralbeik:

omaralbeik:
@pawurb Could you help please.

pawurb:
Sure, will look into it. AFAIK testing watchOS is not as straightforward because out of the box it does not support test targets, but alamofire does it somehow so will try to copy their approach.

omaralbeik:
@pawurb That would be great, thank you!

omaralbeik:
@pawurb watchOS support added in v1.4.6, thanks for the great work!

PR #83, opened 2017-01-10
[#81] add watchOS target

pawurb:
Looks like currently there is no way to test watchOS targets. What is recommended is to test the code in iOS target what we already do. I just did not include files with code not compatible with watchOS into the target.

I tested it with Carthage and imported into a dummy watchOS app. I managed to import framework module but looks like types are not extended because I cannot use any of the SwifterSwift methods. Could you take a look at it? Maybe some lib linking setting that I forgot to tick or sth.

## Checklist


omaralbeik:
@pawurb Working on it, Thank you!

PR #90, opened 2017-01-11
Add travis support for all the targets

pawurb:
Should be fine now. I removed xcpretty because I think that it breaks the return value of test script and could cause false positive build statuses.

## Checklist


pawurb:
I think we should wait with merging until build passes. Btw there is workaround around build statuses, will apply it in a moment.

PR #97, opened 2017-01-15
Enum helper functions

evermeer:
You do have to extend your enums with the EnumExtensions protocol

## Checklist

- [ ] All comments headers have the word **SwifterSwift:** before description.

### DIFF
--- a/Source/Extensions/EnumExtension.swift
+++ b/Source/Extensions/EnumExtension.swift
@@ -1,0 +1,58 @@
+//
+//  EnumExtensions.swift
+//  SSTests
+//
+//  Created by Omar Albeik on 07/12/2016.
+//  Copyright © 2016 Omar Albeik. All rights reserved.
+//
+
+protocol EnumExtensions {  
+    // Returns an array of all enum values
+    static var allValues: [Self] { get }
+    
+    // Returns a .label for the enum value and a .value for the associated value
+    public var associated: (label: String, value: Any?) { get }
+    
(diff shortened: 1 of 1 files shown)


omaralbeik:
@evermeer Nice work, however this doesn't extend a specific type, it creates a new protocol, which is not a supported style in this project.
SwifterSwift is about extending types already defined by Apple for now.
Thank you again!

muescha:
because i searched for `list of all enum values` a long time:

how about to make `SwifterExtensions` or `SwifterProtocols` as a new repo with this kind of extensions/protocols?

muescha:
@evermeer you should also some unit tests.

evermeer:
@muescha There are various projects available that contain Swift helper functions. Since it's impossible to extend an enum and the only way to give an enum more functionality is by extending it with a protocol I was hoping it could still be part of this project. 

If a new repo was made, then SwifterExtension would not ba a good name. This repo is already about extensions. SwifterProtocols would be better. But then it would be difficult to define the borders of the project. For instance my entire library of https://github.com/evermeer/EVReflection can be used as an extension. 

Since these enum function won't be art of this repo, it's also not needed to add any unit tests here. Just point me to a repo where this belongs and I will add it there.

evermeer:
@muescha I just created a repository for 'Stuff' like this.

You can find it at:
https://github.com/evermeer/Stuff

And you can use it with:
`pod "Stuff/Enum"`

PR #110, opened 2017-02-06
Use reduce for string randomizer.

basememara:


Source/Extensions/StringExtensions.swift
```diff
@@ -614,14 +614,21 @@ public extension String {
...
 			self.init(str)
 			return
 		}
 		return nil
 	}
 	
+	/// SwifterSwift: Create a new random string of given length.
+	///
+	/// - Parameter length: number of characters in string.
+	public init(randomOfLength length: Int) {
```
basememara: Nice, good idea 👍

PR #133, opened 2017-03-01
Update UITableView+UITextField test

SD10:
## Checklist

- [X] I checked the [**Contributing Guidelines**](https://github.com/omaralbeik/SwifterSwift/blob/master/CONTRIBUTING.md) before creating this request.
- [X] I have **only one commit** in this pull request.
- [] New extensions support iOS 8 or later.
- [] New extensions are written in Swift 3.
- [] I have added tests for new extensions, and they passed.
- [X] Pull request was created to [**master head branch**](https://github.com/omaralbeik/SwifterSwift/tree/master).
- [] All extensions have a **clear** comments explaining their functionality, all parameters and return type in English.
- [] All comments headers have the word **SwifterSwift:** before description.

Summary of Pull Request:
- Completes testing for UITableViewExtensions
- Removes unused indexPath parameter from UITableViewExtension
```swift
dequeReusableHeaderFooterView(withClass:)
```
- Tests  method in UICollectionViewExtensions
```swift
func reloadData(completion:)
```
- Completes testing for UITextFieldExtensions
- Adds 3 new files for testing: TestImage.png, UITableViewHeaderFooterView.xib, UITableViewCell.xib


SD10:
Guess I have to wait for the AWS issues to resolve to see the build logs for why this failed 😂

omaralbeik:
@SD10 according to https://www.traviscistatus.com/incidents/hmwq9yy5dh9d the problem has been fixed 7 minutes ago
I restarted the build

SD10:
I was able to check the prev build log. It said the UITableView `reloadData(completion:)` test failed.

It looks fine to me, I was thinking the timeInterval could be too short with the poor performance of TravisCI lately. 

It looks like they're having more problems today, so we'll just sit on this.

SD10:
I just modified the test not to depend on time. So close to 90% coverage.

SD10:
Codecov didn't register a hit for the completion block.
Thus, I didn't solve the problem I just avoided it 😭

Problem is in `UIView.animateWithDuration()`
Quote Apple:
"If the duration of the animation is 0, this block is performed at the beginning of the next run loop cycle."

Note to self that I need to fix this.

omaralbeik:
@SD10 hmm, what about [`performWithoutAnimation(_:)`](https://developer.apple.com/reference/uikit/uiview/1622484-performwithoutanimation)

SD10:
It would be nice if that method had another completion parameter. 
I would like the completion block to be called after `reloadData()` is completed.
I've been thinking we could possibly have a value of 0.01 TimeInterval.
However, this could result in some unexpected behavior for the view.
Best solution could involve something with GCD.

The other question is, does the completion need to occur after reloadData() is complete?
If not, this is an option:
```swift
func reloadData(completion: () -> Void) {
    tableView.reloadData()
    completion()
}
```

PR #134, opened 2017-03-05
Reduce shuffle()/shuffled() complexity and minor updates

SD10:
@omaralbeik don't merge this yet it's in progress

- Updated `shuffle()` / `shuffled()` methods to use Fisher-Yates Algorithm
Replaces the previous **O(n^2)** complexity with **O(n)**
The output is now completely random. I've removed the check against the original array.

- Converted `shuffled` and `withoutDuplicates` properties to methods
`withoutDuplicates` has been renamed to `duplicatesRemoved()`

- Updating `pop()` to just return the result of `popLast()` so we can rely on the standard libraries implementation of this method

I'm considering putting the shuffle() / shuffled() methods on an extension to MutableCollection
This way they will work for not only Arrays but ArraySlices as well.
This can be refactored at a later date.

The main reason I'm holding off on this is because having an extension on MutableCollection may not be inherently obvious that it can be used on Arrays.

We would have to acknowledge in the SwifterSwift Docs that Array conforms to MutableCollection. 

(although people would see the method with autocomplete or could view the official Apple docs to see what conforms to MutableCollection)

Ultimately, however, I'd like to share as much functionality as possible between Collections by extending their composing protocols.

### DIFF
--- a/Source/Extensions/ArrayExtensions.swift
+++ b/Source/Extensions/ArrayExtensions.swift
@@ -38,26 +38,6 @@ public extension Array where Element: FloatingPoint {
 	
 }
 
-
-// MARK: - Properties (Equatable)
-public extension Array where Element: Equatable {
-	
-	/// SwifterSwift: Shuffled version of array.
-	public var shuffled: [Element] {
-		var arr = self
-		arr.shuffle()
-		return arr
-	}
-	
-	/// SwifterSwift: Array with all duplicates removed from it.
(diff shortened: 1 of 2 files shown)


SD10:
I was going to say we should wait but okay! These changes are fine for now 😂

omaralbeik:
@SD10 sorry I didn't see your message about merging 😬

PR #139, opened 2017-03-26
Swap index position array extension.

LucianoPAlmeida:
## Checklist

### DIFF
--- a/Source/Extensions/ArrayExtensions.swift
+++ b/Source/Extensions/ArrayExtensions.swift
@@ -148,6 +148,23 @@ public extension Array {
 		append(newElement)
 	}
 	
+    /// SwifterSwift: Swap values at index positions.
+    ///
+    /// - Parameters:
+    ///   - from: index of first element.
+    ///   - to: index of other element.
+    public mutating func swap(from index: Int, to otherIndex: Int)  {
+        guard
+            index != otherIndex &&
+                index >= 0 && index < count &&
+                otherIndex >= 0 && otherIndex < count else {
+                    return
+        }
+        let aux = self[index]
+        self[index] = self[otherIndex]
+        self[otherIndex] = aux
+    }
+    
 }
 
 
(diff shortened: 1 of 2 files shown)


SD10:
Hey @LucianoPAlmeida,
Thank you for the PR!

I think this functionality already be accomplished with the use of:
```swift
var array = [1, 2, 3]

swap(&array[0], &array[2])
```
Was your intent to give this action a new interface or make it a fail safe operation?

I prefer operations that require Index math to fail as you should have already thought carefully about the logic so that your index is never out of range. 

I think this is why Array's have the behavior they do.

Please let me know your thoughts!

omaralbeik:
@LucianoPAlmeida @SD10 How about we change the name to `safeSwap` ?
and @LucianoPAlmeida maybe consider separating the condition in the guard statement to multiple guards, it reads better

Thank you for your contribution, keep the great work!

SD10:
I agree @omaralbeik `safeSwap` would be a great name for this.

I'd also like to base the implementation off using the `startIndex` rather than 0 in case we refactor the extensions to support other types such as `ArraySlice` in the future.

Possibly something like this:

```swift
extension Array {

    public mutating func safeSwap(from index: Int, to otherIndex: Int) {
    
        guard index != otherIndex,
              startIndex...endIndex ~= index,
              startIndex...endIndex ~= otherIndex else { return }
        
        Swift.swap(&self[index], &self[otherIndex])
    
    }
} 
```

LucianoPAlmeida:
Hello @omaralbeik @SD10 
Thank you for the feedback :)

Answering @SD10 's both new interface and make it safe were the intentions for this, but the main one was to give a new interface since i think it's more readable, simple and more swift 3 like API. 

I agree with the name change to safeSafe since make swap operation safe was one of the intentions when i thought this extension. But i also think in keep implementation of swap without safe check just as a simpler interface for ` Swift.swap(&array[index], &array[otherIndex]) `. 

Please let me know your thoughts :)

omaralbeik:
That makes sense to me
Thank you @LucianoPAlmeida!

SD10:
Great work, thank you @LucianoPAlmeida

PR #155, opened 2017-04-19
Add optional string extension for checking if nil or empty

basememara:
Swift 3.1 opens a new world of extensions by allowing to target concrete types. This is an optional string extension for checking for `isNilOrEmpty` stolen from `C#`. Unit test included.

Instead of this:
```
if optionalStr == nil || optionalStr?.isEmpty == true {
    blah.. blah.. show warning.. etc..
}
```
Do this:
```
if optionalStr.isNilOrEmpty {
    blah.. blah.. show warning.. etc..
}
```

### DIFF
--- a/Source/Extensions/StringExtensions.swift
+++ b/Source/Extensions/StringExtensions.swift
@@ -748,3 +748,11 @@ public extension String {
 	}
 	
 }
+
+public extension Optional where Wrapped == String {
+
+    /// A Boolean value indicating whether a string is `nil` or has no characters.
+    public var isNilOrEmpty: Bool {
+        return self?.isEmpty ?? true
+    }
+}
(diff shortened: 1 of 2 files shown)


SD10:
@basememara Thank you for your contributions 🎉

If you could please fix this breaking build due to specializing a generic parameter I would appreciate it.
This may be better suited as a `String` extension rather than on `Optional`.

I don't think there is a good protocol to constrain `Wrapped` to.
It doesn't make sense to use something like `ExpressibleByStringLiteral`.

If you decide to keep it in `Optional` then please move it to the `Optional` file.

basememara:
I think the build is failing because the `.travis` is targeting `Xcode 8.2`. I have another PR for updating the `.travis` config.

basememara:
I just saw your comments in the other PR, makes sense. We might have to hold off on this extension if this is something we want to add. Thanks for the feedback 👍

SD10:
@basememara thank you! I'm hoping we can update to Xcode 8.3 and remove this limitation.

PR #156, opened 2017-04-19
Add new locale extension with POSIX static property

basememara:
Provides a static property for POSIX locale. Great on performance too since statically cached for reuse. Added new locale unit tests.

Instead of this:
```
dateFormatter.locale = Locale(identifier: "en_US_POSIX")
```
Do this:
```
dateFormatter.locale = .posix
```


SD10:
@basememara Thank you for the contribution 👍

PR #161, opened 2017-04-25
ArrayExtensions last method codecoverage fix

tapanprakasht:
ArrayExtensions has code coverage issue where **last** method code coverage report says **return nil** is not executed even though all the test passes. The exact reason for why this is happening is not clear.

image

But we have a similar method like **lastIndex** in same extensions which is returning nil. Here code coverage report says return nil executed. The only difference I see is the use of where condition in with for loop between lastIndex and last method. 

By this pull request, I made **last** method code similar as **lastIndex**.

## Checklist
- [ ] New extensions support iOS 8 or later.
- [ ] New extensions are written in Swift 3.

### DIFF
--- a/Source/Extensions/ArrayExtensions.swift
+++ b/Source/Extensions/ArrayExtensions.swift
@@ -154,7 +154,9 @@ public extension Array {
     /// - Parameter condition: condition to evaluate each element against.
     /// - Returns: the last element in the array matching the specified condition. (optional)
     public func last(where condition: (Element) -> Bool) -> Element? {
-        for element in reversed() where condition(element) { return element }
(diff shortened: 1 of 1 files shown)


SD10:
@tapanprakasht this is interesting, nice fix 👍

tapanprakasht:
Thanks, @SD10. But I really don't understand why the previous code is not covered the code coverage tool. I am just trying to find the real reason behind it. If anybody knows it would be helpful.

SD10:
@tapanprakasht I have no idea really because as you said the test for nil passes

omaralbeik:
Thank you @tapanprakasht!

PR #165, opened 2017-04-30
Remove drop(while:) Array extension

SD10:
### Summary of Pull Request:
- Removes `drop(while:)` extension on `Array` that already exists in the Swift Standard Library


omaralbeik:
Thank you @SD10

PR #194, opened 2017-07-12
For each slice implementation.

LucianoPAlmeida:
## Checklist


Tests/SwifterSwiftTests/ArrayExtensionsTests.swift
```diff
@@ -277,4 +277,64 @@ class ArrayExtensionsTests: XCTestCase {
         XCTAssertEqual(grouped["j"] ?? [], [ "james", "jordan", "jonshon" ])
         XCTAssertEqual(grouped["i"] ?? [], [ "irving", "iverson" ])
     }
+    
+    func testForEachSliceEqualIterations() {
```
SD10: Let's condense all this into 1 test. For all SwifterSwift extensions we do 1 extension : 1 unit test.
We don't need to check the number of iterations or their values in every test, only for one.
A slice of zero should just have no side effects.
A slice greater than the array count should just perform the body on everything at once.
A slice that does not divide the total evenly, here you should check the iteration count + output of each iteration.

PR #199, opened 2017-07-17
Adding array split(size:) method

LucianoPAlmeida:
## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,7 +34,7 @@ N/A
 - New **Array** extensions
     - added `groupByKey` to group the elements of the array by key in a dictionary. [#181](https://github.com/SwifterSwift/SwifterSwift/pull/181) by [@LucianoPAlmeida](https://github.com/LucianoPAlmeida)
     - added `forEach(slice:body:)` to iterate by specified slice size and call a closure. [#194](https://github.com/SwifterSwift/SwifterSwift/pull/194/files) by [@LucianoPAlmeida](https://github.com/LucianoPAlmeida)
-
+    - added `split(size:)` to split in an array of slices of a size. [#199](https://github.com/SwifterSwift/SwifterSwift/pull/199) by [@LucianoPAlmeida](https://github.com/LucianoPAlmeida)
 - New **Dictionary** extensions
     - add `count(where:)` to count dictionary elements where the condition returns true. [#193](https://github.com/SwifterSwift/SwifterSwift/pull/193) by [@LucianoPAlmeida](https://github.com/LucianoPAlmeida)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,8 @@ All notable changes to this project will be documented in this file.
 N/A
 
 ### Enhancements
-N/A
+- New **Array** extensions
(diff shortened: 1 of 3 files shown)


Sources/Extensions/Foundation/ArrayExtensions.swift
```diff
@@ -280,6 +280,23 @@ public extension Array {
             value += slice
         }
     }
+    
+    
+    /// SwifterSwift: Returns an array of slices of lenght "size" from the array.  If array can't be split evenly, the final slice will be the remaining elements.
+    ///
+    /// - Parameters:
+    ///   - size: The size of the slices to be returned.
+    public func split(size: Int) -> [ArraySlice<Element>]? {
```
SD10: I'm a little concerned about the memory implications of returning an ArraySlice here. I think returning an Array would be a safer API for the user.

Sources/Extensions/Foundation/ArrayExtensions.swift
```diff
@@ -280,6 +280,23 @@ public extension Array {
...
+        //Inspired by: https://lodash.com/docs/4.17.4#chunk
+        guard size > 0, !isEmpty else { return nil }
+        var value : Int = 0
```
SD10: Type inference 🤔

Tests/FoundationTests/ArrayExtensionsTests.swift
```diff
@@ -333,4 +333,31 @@ class ArrayExtensionsTests: XCTestCase {
         
     }
     
+    func testSplitSize() {
+        
+        // A slice with value zero
+        var array : [String] = [ "james", "irving", "jordan", "jonshon", "iverson", "shaq"]
```
SD10: Type inference 😅

CHANGELOG.md
```diff
@@ -34,7 +34,7 @@ N/A
 - New **Array** extensions
     - added `groupByKey` to group the elements of the array by key in a dictionary. [#181](https://github.com/SwifterSwift/SwifterSwift/pull/181) by [@LucianoPAlmeida](https://github.com/LucianoPAlmeida)
     - added `forEach(slice:body:)` to iterate by specified slice size and call a closure. [#194](https://github.com/SwifterSwift/SwifterSwift/pull/194/files) by [@LucianoPAlmeida](https://github.com/LucianoPAlmeida)
-
+    - added `split(size:)` to split in an array of slices of a size. [#199](https://github.com/SwifterSwift/SwifterSwift/pull/199) by [@LucianoPAlmeida](https://github.com/LucianoPAlmeida)
```
SD10: Thank you 👍

Sources/Extensions/Foundation/ArrayExtensions.swift
SD10: @LucianoPAlmeida I'm also thinking this method may be better named as `group(by size: Int)`. I vaguely remember discussing this when adding the `groupByKey` extension recently.

LucianoPAlmeida: @SD10 Hey man :)
About the memory implications, I think it's the kind of extension that the people who use knows the concerns about ArraySlice. But i don't know, may it's something to consider. 
I just return the ArraySlice because I wanted to follow the same, pattern of the native mehtods that do similar stuff like `split(whereSeparator: (Element) throws -> Bool)-> [ArraySlice<Element>]` and by the way the name split is also to follow the pattern of this methods.
I thought about `slices(ofSize:)` or `split(ofSize:)`, let me know what you think?

PR #210, opened 2017-07-21
Add the limit length of the UITextField

xingfuqiu:
## Checklist

### DIFF
--- a/Sources/Extensions/UIKit/UITextFieldExtensions.swift
+++ b/Sources/Extensions/UIKit/UITextFieldExtensions.swift
@@ -100,6 +100,19 @@ public extension UITextField {
         self.leftView?.frame.size = CGSize(width: image.size.width + padding, height: image.size.height)
         self.leftViewMode = UITextFieldViewMode.always
     }
+    
+    /// SwifterSwift: Add the limit length
+    ///
+    /// - Parameter length: length limit
+    public func limit(_ length: Int) {
+        NotificationCenter.default.addObserver(forName: NSNotification.Name.UITextFieldTextDidChange, object: nil, queue: OperationQueue.main) { (notification) in
+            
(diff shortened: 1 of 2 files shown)


Sources/Extensions/UIKit/UITextFieldExtensions.swift
```diff
@@ -100,6 +100,19 @@ public extension UITextField {
         self.leftView?.frame.size = CGSize(width: image.size.width + padding, height: image.size.height)
         self.leftViewMode = UITextFieldViewMode.always
     }
+    
+    /// SwifterSwift: Add the limit length
+    ///
+    /// - Parameter padding: length limit
```
SD10: The parameter name is wrong here, it should be length

xingfuqiu: sorry, my mistake

omaralbeik:
@xingfuqiu Thank you for your handy extension, however I don't think it is a good fit for this library for the following reasons:
1. The extension is adding an observer to NotificationCenter but never removes it which might lead to some memory issues
2. The observer is added from  the function without asking users for permission, or at least letting them know, what happens if they have another notification with the same name?
3. Most importantly Apple offers a better and shorter way for doing this using `UITextFieldDelegate`, here is how:


```swift
class ViewController: UIViewController, UITextFieldDelegate {
	
	@IBOutlet weak var textField: UITextField!
	
	override func viewDidLoad() {
		super.viewDidLoad()
		textField.delegate = self
	}
	
	func textField(_ textField: UITextField, shouldChangeCharactersIn range: NSRange, replacementString string: String) -> Bool {
		guard let text = textField.text else {
			return true
		}
		let newLength = text.characters.count + string.characters.count - range.length
		return newLength <= 10
	}
}
```

Issue #214, opened 2017-07-25
CoreImage not available on watchOS


tache:
[x] I've read, understood, and done my best to follow the [Contributing guidelines](https://github.com/SwifterSwift/SwifterSwift/tree/master/CONTRIBUTING.md) before opening this issue.

In the latest commit, I added `CoreImage` accessor extension, but in testing it further, I discovered that it does not have watchOS support. https://developer.apple.com/documentation/coreimage

In order to resolve this issue, there are two options ... please let me know and I will correct the issue.

Should I remove it? 
or 
Should I wrap it?

```
    #if os(iOS) || os(tvOS)
        /// SwifterSwift: CoreImage.CIColor (read-only). Only available on iOS and tvOS
         public var coreImageColor: CoreImage.CIColor? {
             return CoreImage.CIColor(color: self)  // The resulting Core Image color, or nil
        }
    #endif
```

omaralbeik:
@tache I'd suggest you wrap it 👍

omaralbeik:
@tache Could you pleas submit a PR with the functions wrapped only for not watchOS :)

PR #217, opened 2017-07-27
Fixed the CoreImage accessor in UIColor to be restricted to iOS and tvOS

tache:
This PR addresses issue #214 

Issue: CoreImage is not available on watchOS, resulting in a compilation error. 
Fix: Restricts the accessor to be only available on iOS and tvOS via wrapping it in a conditional compilation block.

### DIFF
--- a/Tests/UIKitTests/UIColorExtensionsTests.swift
+++ b/Tests/UIKitTests/UIColorExtensionsTests.swift
@@ -124,6 +124,32 @@ class UIColorExtensionsTests: XCTestCase {
 
     }
 
+    func testCoreImageColor() {
+        var color = UIColor(hex: 0xFF0000, transparency: 1.0)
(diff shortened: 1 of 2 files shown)


omaralbeik:
Thank you @tache!

Tests/UIKitTests/UIColorExtensionsTests.swift
```diff
@@ -124,6 +124,32 @@ class UIColorExtensionsTests: XCTestCase {
 
     }
 
+    func testCoreImageColor() {
```
SD10: Doesn't this need to be wrapped as well? I don't think we're running watchOS tests on Travis which allows this to pass.

omaralbeik: We're not running watchOS tests currently, but ideally, of course this should be wrapped, @SD10 thanks for bringing this up, didn't notice 😬.
@tache Could you please wrap the test function as well 😅

PR #225, opened 2017-08-13
Improve sum & average extensions by not relying on reduce

SD10:
Sometimes the smallest code isn't always the most performant. I reckon `removeDuplicates()` and `duplicatesRemoved()` would also benefit from this approach. I was a bit too lazy though 😅

I'm using Travis to run my tests for this because my simulator was too slow 😐

### DIFF
--- a/Sources/Extensions/Foundation/ArrayExtensions.swift
+++ b/Sources/Extensions/Foundation/ArrayExtensions.swift
@@ -18,8 +18,9 @@ public extension Array where Element: Integer {
 	///
 	/// - Returns: sum of the array's elements.
 	public func sum() -> Element {
-		// http://stackoverflow.com/questions/28288148/making-my-function-calculate-average-of-array-swift
-		return reduce(0, +)
+        var total: Element = 0
+        forEach { total += $0 }
+        return total
 	}
 	
 }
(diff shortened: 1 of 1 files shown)


Sources/Extensions/Foundation/ArrayExtensions.swift
```diff
@@ -34,8 +35,10 @@ public extension Array where Element: FloatingPoint {
 	///
 	/// - Returns: average of the array's elements.
 	public func average() -> Element {
-		// http://stackoverflow.com/questions/28288148/making-my-function-calculate-average-of-array-swift
-		return isEmpty ? 0 : reduce(0, +) / Element(count)
+        guard isEmpty == false else { return 0 }
```
omaralbeik: I think using !isEmpty makes it more readable here 😃 
```swift
guard !isEmpty else { return 0 }
```

SD10: I actually specifically don't use `!` because I don't think its readable at all 🤔 
I only use it to invert `Booleans` when setting values. Not when writing conditionals.
We can still change it though

SD10: Better to even change it to:
```Swift
if isEmpty { return 0 }
```
But you are supposed to use guards for exiting the scope early :P

omaralbeik: @SD10, I see, that makes sense as well 🤔
I'm cool with both styles, feel free to do whatever you feel is more readable 👍

Issue #190, opened 2017-07-09
UIColor.hexString not supporting alpha


SuperWomble:
Issue: hexString does not honour the color's alpha component.

Although the color's alpha is retrieved as part of the getting of the components, it is ignored, and  the output string is formatted to #RRGGBB instead of #RRGGBBAA

Desired outcome: if alpha is 1, then the current form should be returned. Otherwise, the string should be formatted to #(8 chars)
Cheers.

SD10:
@SuperWomble Thank you for reporting this. Do you want to submit a pull request with this change?

SuperWomble:
TBH, and I am shamed about this, but I'm a total github nub. I barely know how to read the site, let alone contribute anything more than issues. ::deep, deep shame::

SD10:
@SuperWomble There's nothing to be ashamed about! I'm _still_ a git noob 😐

pvtmert:
okay so i saw this and decided to close, but need review on following snippet:

```swift
extension UIColor {
	@available(*, deprecated, message: "Variable format no longer supported, use 'hexString(withAlpha:)' instead")
	public var hexString: String {
		return hexString(withAlpha: false)
	}

	public func hexString(withAlpha: Bool = false) -> String {
		let alphaShift = withAlpha ? 8 : 0
		var red:	CGFloat = 0
		var green:	CGFloat = 0
		var blue:	CGFloat = 0
		var alpha:	CGFloat = 0

		getRed(&red, green: &green, blue: &blue, alpha: &alpha)
		let rgb: Int = (withAlpha ? (Int)(alpha*255) : 0)
			| (Int)(red*255)   << (16+alphaShift)
			| (Int)(green*255) << (8+alphaShift)
			| (Int)(blue*255)  << (0+alphaShift)
		return NSString(format: (withAlpha ? "#%08x" : "#%06x"), rgb).uppercased as String
	}
}

let color = UIColor.init(red: 0.5, green: 0.5, blue: 0.5, alpha: 0.5)
print(color.hexString, color.hexString(withAlpha: true))
```

LucianoPAlmeida:
@pvtmert Just one thing, the swift String class also has a constructor String(format: ), so you don't need to use NSString then cast it to string :)

pvtmert:
[...]

PR #228, opened 2017-08-15
UIColor.hexString alpha support #190

pvtmert:
## Checklist

also without `continue` @omaralbeik ;)

also little question here:
shouldn't `UIColor(hex: 0xABC)` generate `0xAABBCC` rather than `0x000ABC` ?

### DIFF
--- a/Sources/Extensions/UIKit/UIColorExtensions.swift
+++ b/Sources/Extensions/UIKit/UIColorExtensions.swift
@@ -70,19 +70,9 @@ public extension UIColor {
 		return UInt(colorAsUInt32)
 	}
 	
-	/// SwifterSwift: Short hexadecimal value string (read-only, if applicable).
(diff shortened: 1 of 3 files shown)


Sources/Extensions/UIKit/UIColorExtensions.swift
```diff
@@ -70,18 +70,12 @@ public extension UIColor {
 		return UInt(colorAsUInt32)
 	}
 	
-	/// SwifterSwift: Hexadecimal value string (read-only).
+	/// SwifterSwift: -Deprecated- Hexadecimal value string (read-only).
+	@available(*, deprecated, message: "Variable format no longer supported, use 'hexString(withAlpha:)' instead")
```
SD10: I think we're putting all deprecated extensions in separate file

PR #230, opened 2017-08-16
3.1.1

omaralbeik:
[v3.1.1](https://github.com/SwifterSwift/SwifterSwift/blob/3.1.1/CHANGELOG.md#v311)


Sources/Extensions/Foundation/DoubleExtensions.swift
```diff
@@ -7,6 +7,11 @@
 //
 
 import Foundation
+#if os(macOS)
+	import Cocoa
+#else
+	import UIKit
+#endif
```
SD10: Just curious why we need Cocoa or UIKit here? Some work around?

omaralbeik: I tried, `pod lib lint` is complaining because of `CGFloat` not in `Foundation` ¯\_(ツ)_/¯

SD10: Wow, didn't know that 😱

PR #233, opened 2017-09-01
IBOutlet connectable ‘nextField’

drewpitchford:
**IBOutlet Connectable 'nextField'**
This is useful for forms. By assigning the 'nextField' for each textField in a form, the UITextFieldDelegate can do the following to move to the next active field when the user taps the return key on the keyboard.

`textField.nextField?.becomeFirstResponder()`

**Blur background view of UIView**
Blur the background of any UIView using a UIVisualBlurEffectStyle.
Example:
```
let view = UIView(frame: frame)
view.blurBackground(using: .dark) // Blurs the view's background with a Apple's UIVisualBlurEffectStyle.dark
```

## Checklist

- [x ] New extensions are written in Swift 3.
- [x ] New extensions support iOS 8.0+ / tvOS 9.0+ / macOS 10.10+ / watchOS 2.0+.
- [ x] I have added tests for new extensions, and they passed.

### DIFF
--- a/Sources/Extensions/UIKit/UITextFieldExtensions.swift
+++ b/Sources/Extensions/UIKit/UITextFieldExtensions.swift
@@ -127,7 +127,10 @@ public extension UITextField {
 	}
     
     /// SwifterSwift: Next text field to become first responder
-    /// textField.nextField?.becomeFirstResponder()
+    ///
+    ///     textField.nextField = someOtherTextField
+    ///     textField.nextField?.becomeFirstResponder() -> 'someOtherTextField` is now first responder
+    ///
     /// https://stackoverflow.com/a/27030181
     @IBOutlet public var nextField: UITextField? {
         
--- a/Sources/Extensions/UIKit/UITextFieldExtensions.swift
+++ b/Sources/Extensions/UIKit/UITextFieldExtensions.swift
@@ -8,6 +8,8 @@
 
 #if os(iOS) || os(tvOS)
 import UIKit
+    
+fileprivate var kAssociationKeyNextField: UInt8 = 0
 
 // MARK: - Enums
 public extension UITextField {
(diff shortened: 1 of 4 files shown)


Sources/Extensions/UIKit/UIViewExtensions.swift
```diff
@@ -525,6 +525,42 @@ public extension UIView {
 		anchorCenterXToSuperview()
 		anchorCenterYToSuperview()
 	}
+    
+    /// SwifterSwift: Blur the background of the view
+    ///
+    /// - Parameter effectStyle: The UIBlurEffectStyle to apply to the background
+    @available(iOS 9, *) public func blurBackground(using effectStyle: UIBlurEffectStyle) {
```
SD10: @drewpitchford Hey, thank you for your pull request 💪 🎉 Can this not be accomplished by the extensions we have on `UIView`?

Sources/Extensions/UIKit/UITextFieldExtensions.swift
```diff
@@ -123,6 +125,24 @@ public extension UITextField {
...
 	}
+    
+    /// SwifterSwift: Next text field to become first responder
+    ///
+    ///     textField.nextField = someOtherTextField
+    ///     textField.nextField?.becomeFirstResponder() -> 'someOtherTextField` is now first responder
+    ///
+    /// https://stackoverflow.com/a/27030181
+    @IBOutlet public var nextField: UITextField? {
+
```
SD10: I think this should probably be marked `weak`. However, it may not be a good fit for the library. I'm not fond of the global constant and I can't help but feel there is a Swifter way of doing this.

Sources/Extensions/UIKit/UIViewExtensions.swift
drewpitchford: I'm not sure if it can be, but the point of this method is to wrap it up into a single method rather than having the developer have to do all the logic on his end. If you do not think it is useful, that's fine. I just found myself using this method in a lot of my projects and figured it might be useful for others.

Sources/Extensions/UIKit/UITextFieldExtensions.swift
drewpitchford: Agreed on 'weak'. We can remove the global constant and just put '0' in that field. I'm not sure of any other way to make a property available to IB. I don't think @IBDesignable is applicable to this. So, perhaps I just remove this change.

Sources/Extensions/UIKit/UIViewExtensions.swift
SD10: I see. We already have this extension on `UIImageView`. I was thinking this was on `UIView`. It would be good to provide this extension on all `UIView` classes. @omaralbeik Is there any reason why this extension is on `UIImageView` opposed to `UIView`?

omaralbeik: @SD10 @drewpitchford Nope, I don't remember there was a reason for this, @drewpitchford it would be great to submit another PR with only `blurBackground` moved to UIView to separated it from other changes

PR #237, opened 2017-09-05
Add blur to background of a UIView

drewpitchford:
Adds ability to add and remove blurs to the background of UIViews. This is distinct from blurring an image itself that is found in UIImageView+Extensions.

## Checklist

### DIFF
--- a/Sources/Extensions/UIKit/UIViewExtensions.swift
+++ b/Sources/Extensions/UIKit/UIViewExtensions.swift
@@ -525,6 +525,42 @@ public extension UIView {
 		anchorCenterXToSuperview()
 		anchorCenterYToSuperview()
 	}
+    
+    /// SwifterSwift: Blur the background of the view
+    ///
+    /// - Parameter effectStyle: The UIBlurEffectStyle to apply to the background
+    @available(iOS 9, *) public func blurBackground(using effectStyle: UIBlurEffectStyle) {
+        
+        backgroundColor = .clear
+        removeBlurs()
+        
+        let blurEffect = UIBlurEffect(style: effectStyle)
+        let blurredEffectView = UIVisualEffectView(effect: blurEffect)
+        blurredEffectView.frame = bounds
+        
+        addSubview(blurredEffectView)
+        sendSubview(toBack: blurredEffectView)
+        blurredEffectView.fillToSuperview()
+        
+        let vibrancyEffect = UIVibrancyEffect(blurEffect: blurEffect)
+        let vibrancyEffectView = UIVisualEffectView(effect: vibrancyEffect)
+        vibrancyEffectView.frame = bounds
+        
+        blurredEffectView.contentView.addSubview(vibrancyEffectView)
+        vibrancyEffectView.fillToSuperview()
+    }
+    
+    // SwifterSwift: Remove previous blurs from the view
+    @available(iOS 9, *) public func removeBlurs() {
+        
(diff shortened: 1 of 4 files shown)


drewpitchford:
How do I start another build? It seems to have failed with some on the build server. The tests passed.

omaralbeik:
@drewpitchford I restarted the build again :)

Sources/Extensions/UIKit/UIViewExtensions.swift
```diff
@@ -525,6 +525,42 @@ public extension UIView {
...
+        backgroundColor = .clear
+        removeBlurs()
+        
+        let blurEffect = UIBlurEffect(style: effectStyle)
+        let blurredEffectView = UIVisualEffectView(effect: blurEffect)
+        blurredEffectView.frame = bounds
+        
+        addSubview(blurredEffectView)
+        sendSubview(toBack: blurredEffectView)
+        blurredEffectView.fillToSuperview()
```
SD10: @drewpitchford We're trying not to use SwifterSwift extensions inside of other SwifterSwift extensions. Maybe I should add this to the contributing guidelines 🤔

Sources/Extensions/UIKit/UIViewExtensions.swift
```diff
@@ -525,6 +525,42 @@ public extension UIView {
 		anchorCenterXToSuperview()
 		anchorCenterYToSuperview()
 	}
+    
+    /// SwifterSwift: Blur the background of the view
+    ///
+    /// - Parameter effectStyle: The UIBlurEffectStyle to apply to the background
+    @available(iOS 9, *) public func blurBackground(using effectStyle: UIBlurEffectStyle) {
+        
+        backgroundColor = .clear
```
SD10: Do we have to set the `backgroundColor` to `.clear`?

Sources/Extensions/UIKit/UIViewExtensions.swift
```diff
@@ -525,6 +525,42 @@ public extension UIView {
...
+        addSubview(blurredEffectView)
+        sendSubview(toBack: blurredEffectView)
+        blurredEffectView.fillToSuperview()
+        
+        let vibrancyEffect = UIVibrancyEffect(blurEffect: blurEffect)
+        let vibrancyEffectView = UIVisualEffectView(effect: vibrancyEffect)
+        vibrancyEffectView.frame = bounds
+        
+        blurredEffectView.contentView.addSubview(vibrancyEffectView)
+        vibrancyEffectView.fillToSuperview()
```
SD10: Same as above. Maybe we shouldn't be adding constraints either

Sources/Extensions/UIKit/UIViewExtensions.swift
drewpitchford: Yes, otherwise you do not get the blurred/frosted glass affect as content scrolls behind the view.

Sources/Extensions/UIKit/UIViewExtensions.swift
drewpitchford: If we don't add constraints, then the blur won't resize with the superview on rotation.

SD10:
Hey @drewpitchford,  I'm going to close this for now because I think this extension doesn't fit with some of SwifterSwift's technical considerations. However, we really appreciate you taking the time to put this together 👍

PR #241, opened 2017-09-07
Update files structure

omaralbeik:


Sources/Extensions/SwiftStdlib/DictionaryExtensions.swift
```diff
@@ -77,7 +77,8 @@ public extension Dictionary {
 		}
 		let options = (prettify == true) ? JSONSerialization.WritingOptions.prettyPrinted : JSONSerialization.WritingOptions()
 		let jsonData = try? JSONSerialization.data(withJSONObject: self, options: options)
-		return jsonData?.string(encoding: .utf8)
+		guard let data = jsonData else { return nil }
```
SD10: Was this an internal dependency on a SwifterSwift extension? 😃 We should remove any we can!

omaralbeik: @SD10 This is because when using a subspec, all code should be included within it. also as we discussed before we might have some duplicated code here and there, but this will make the code easier to be included manually as a small pieces in other projects

PR #248, opened 2017-09-16
Add isBetween method to date extension

BennX:
Added a `isBetween` method to the date extension which can be used to determine if a date is between two given dates including or excluding the given dates itself. Default is to include the dates itself.

## Checklist

### DIFF
--- a/Sources/Extensions/Foundation/DateExtensions.swift
+++ b/Sources/Extensions/Foundation/DateExtensions.swift
@@ -709,7 +709,21 @@ public extension Date {
 	public func daysSince(_ date: Date) -> Double {
 		return self.timeIntervalSince(date)/(3600*24)
 	}
-	
+    
+    /// SwifterSwift: check if a date is between two other dates
+    ///
+    /// - Parameter startDate: start date to compare self to.
+    /// - Parameter endDate: endDate date to compare self to.
+    /// - Parameter includeBounds: true if the start and end date should be included (default is true)
+    /// - Returns: true if the date is between the two given dates.
+    public func isBetween(_ startDate: Date, _ endDate: Date, includeBounds: Bool = true) -> Bool {
+        if includeBounds {
+            return startDate.compare(self).rawValue * self.compare(endDate).rawValue >= 0
+        } else {
+            return startDate.compare(self).rawValue * self.compare(endDate).rawValue > 0
+        }
+    }
+    
 }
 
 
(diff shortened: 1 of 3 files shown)


Sources/Extensions/Foundation/DateExtensions.swift
```diff
@@ -709,7 +709,21 @@ public extension Date {
...
 	}
-	
+    
+    /// SwifterSwift: check if a date is between two other dates
+    ///
+    /// - Parameter startDate: start date to compare self to.
+    /// - Parameter endDate: endDate date to compare self to.
+    /// - Parameter includeBounds: true if the start and end date should be included (default is true)
+    /// - Returns: true if the date is between the two given dates.
+    public func isBetween(_ startDate: Date, _ endDate: Date, includeBounds: Bool = true) -> Bool {
```
SD10: I'm fine with this. Clean & well documented 🎉 I'm just curious if we can use:
```Swift
let date1 = Date()
let date2 = Date()
let date3 = Date()
let bool = (date1...date2).contains(date3)
```
I agree having the bounds check makes this easier though. I would like to see it default to `false` though. Because `between` is technically not inclusive. What do you think @BennX?

BennX: @SD10 Of cause it does make more sense to have it set to false.

BennX: I also like the Range idea. Lets see if i can figure that out.

SD10: @BennX The implementation doesn't need to be changed. It's fine as is. This could be more performant actually.

BennX:
```Swift
extension ClosedRange where Bound == Date {
    public func contains(_ date: Date) -> Bool {
        return date.isBetween(self.lowerBound, self.upperBound)
    }
}
```
That would it be. Wonder if something like this should be in the same DateExtensions file?

SD10: @BennX No, we shouldn't add that extension. This is already possible and would conflict with the Swift standard library.

BennX:
@SD10 hope the format fits.

SD10:
@BennX This would go under the `Next Release` section. `v3.2.1` would be a bug fix release. We always add new CHANGELOG entries to the `Next Release` section

BennX:
@SD10 Got it. Cheers!

PR #252, opened 2017-09-27
[Added] Danger to continuous integration

SD10:
If I set this up properly we should see a message from @SwifterSwiftBot. I accidently left an ugly commit so we'll have to squash that

### DIFF
--- a/Dangerfile
+++ b/Dangerfile
@@ -1,0 +1,22 @@
+message(“Thank you for submitting a pull request to SwifterSwift. The team will review your submission as soon as possible.“)
+
+# Checks for modified source files
+source_changes_exist = !git.modified_files.grep(/Sources/).empty?
+
+# Checks for changelog entry
+no_changelog_entry = !git.modified_files.include?(CHANGELOG.md)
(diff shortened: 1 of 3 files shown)


Dangerfile
```diff
@@ -0,0 +1,22 @@
+message(“Thank you for submitting a pull request to SwifterSwift. The team will review your submission as soon as possible.“)
```
sunshinejr: YOU'RE BETRAYING DANGER-SWIFT WITH THIS PEACE OF RUBY? 😭

SD10: 🤣 You can see where my heart was in the previous commit. I can't even write Ruby. 
danger-swift is still blocked by Marathon but it's close!

SD10:
cc @omaralbeik feel free to merge this if you're good with it!

Dangerfile
orta: Hah, yeah, if anyone wants to try help on that it's a real fun one  -https://github.com/JohnSundell/Marathon/pull/127

PR #254, opened 2017-09-27
Add ColorExtensions

omaralbeik:


CHANGELOG.md
```diff
@@ -17,6 +17,20 @@ All notable changes to this project will be documented in this file.
...
 
 
+# v4.0.1
+
+### API Breaking
+N/A
+
+### Enhancements
+- **Color**
+  - Refactored duplicated code from `UIColorExtensions` and `NSColorExtensions` into `ColorExtensions`. thanks to [SD10](https://github.com/SD10).
```
SD10: Why did you list me lol?? I didn't do this PR!

omaralbeik: You inspired me to do it :)

omaralbeik: @SD10 btw, We'll have to do the same with UIView/NSView extensions, what do you think?

SD10: I'm not familiar enough with `NSView` to know for sure if it would be directly compatible.

omaralbeik: @SD10 things like cornerRadius, shadow, borderColor, ... are very similar

SD10: There may be somethings but a lot would be broken. I don't think `NSView` has an `animateWithDuration` method. Also, anything that `UIView` does with `UIImage` would also require `NSImage`. It could be tricky but you could save a couple extensions? I'm not sure if its worth it

PR #269, opened 2017-10-03
Add localized and localized(comment:) to String

Vyax:
## Checklist

I wasn't able to write tests for these extensions, because they require running with a different localization. As far as I know, it is currently not possible to set the 'Application Language' for a single test.

### DIFF
--- a/Sources/Extensions/SwiftStdlib/StringExtensions.swift
+++ b/Sources/Extensions/SwiftStdlib/StringExtensions.swift
@@ -371,6 +371,14 @@ public extension String {
 		}
 		return result
 	}
+    
+    /// SwifterSwift: Returns a localized string, with an optional comment for translators.
+    ///
+    ///        "Hello world".localized -> Hallo Welt
+    ///
+    public func localized(comment: String = "") -> String {
+        return NSLocalizedString(self, comment: comment)
+    }
 	
 	/// SwifterSwift: The most common character in string.
 	///


Sources/Extensions/SwiftStdlib/StringExtensions.swift
```diff
@@ -231,6 +231,13 @@ public extension String {
 	public var length: Int {
 		return characters.count
 	}
+    /// SwifterSwift: Returns a localized string.
+    ///
+    ///        "Hello!".localized -> Hallo!
+    ///
+    public var localized: String {
+        return NSLocalizedString(self, comment: "")
+    }
```
omaralbeik: I prefer having a single function here with a default value for comment like below:
```swift
func localized(comment: String = "") {
    ...
}
```

Sources/Extensions/SwiftStdlib/ArrayExtensions.swift
```diff
@@ -56,6 +56,17 @@ public extension Array {
...
+    /// SwifterSwift: Get the n last elements.
+    ///
+    ///        [1, 7, 1, 2, 4, 1, 6].last(n: 3) -> [4, 1, 6]
+    ///
+    /// - Parameter n: maximum number of elements.
+    /// - Returns: array slice of the n last elements, otherwise an empty array
+    public func last(n: Int) -> ArraySlice<Element> {
+        guard n > 0 && !isEmpty else { return [] }
+        return self[Swift.max(endIndex - n, 0)...]
+    }
```
omaralbeik: @Vyax I noticed you have another PR #270 with the same function, please make sure it is available in one of them only :)

Sources/Extensions/SwiftStdlib/StringExtensions.swift
Vyax: Thank you for the suggestion. However, I frequently use localized strings without comments and would prefer "HelloWorld".localized to "HelloWorld".localized().

Sources/Extensions/SwiftStdlib/ArrayExtensions.swift
Vyax: Thank you, I removed the extraneous commit.

Vyax:
Sorry for closing and reopening the pull request. Back to our conversation:
I frequently use localized strings without comments and would prefer "HelloWorld".localized to "HelloWorld".localized(). What do you think about it?

omaralbeik:
@Vyax I'd still prefer to have a single function :)
Lets ask other members here: @SD10, @LucianoPAlmeida @pvtmert what do you think guys?

LucianoPAlmeida:
Hey guys :) @omaralbeik, @Vyax , I like the localized to, but in my opinion I prefer a method with comment, because it's very useful when you work on teams to the other guys know what that localized key means. Maybe if you like use without comment you can define a default value like: 
``` swift 
      func localized(comment: String = "")  { 
            return NSLocalizedString(self, comment: comment)
      }
```
This way you can use both with or without comments :)

Vyax:
Okay, 2 vs 1. I changed it to a single function with an optional parameter.

PR #272, opened 2017-10-04
Add random to Bool, including a test

Vyax:
## Checklist

I'm not quite sure about the test, but the chance that it fails should be sufficiently small.

### DIFF
--- a/Tests/SwiftStdlibTests/BoolExtensionsTests.swift
+++ b/Tests/SwiftStdlibTests/BoolExtensionsTests.swift
@@ -25,6 +25,18 @@ class BoolExtensionsTests: XCTestCase {
 		XCTAssertFalse(true.toggled)
 		XCTAssert(false.toggled)
 	}
+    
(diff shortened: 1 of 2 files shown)


Tests/SwiftStdlibTests/BoolExtensionsTests.swift
```diff
@@ -25,6 +25,18 @@ class BoolExtensionsTests: XCTestCase {
 		XCTAssertFalse(true.toggled)
 		XCTAssert(false.toggled)
 	}
+    
+    func testRandom() {
+        var yes = 0, no = 0
+        for _ in 1...10000 {
```
SD10: It would be nice to not do this 10,000 times 😱 Just making note of it

omaralbeik: @SD10 I agree, 10 times is enough, will update it soon

PR #285, opened 2017-10-12
Improve Array.sum() and Array.average()

basvankuijck:
300% speed improvement.

```swift
import Foundation

extension Array where Element: Numeric {
    func sum() -> Element {
        var total: Element = 0
        forEach { total += $0 }
        return total
    }

    func sumReduce() -> Element {
        return self.reduce(0 as Element, +)
    }
}

let ar = Array(0..<100_000)

var startTime = CFAbsoluteTimeGetCurrent()
print("sum()")
print("  = \(ar.sum())")
print("  in \(CFAbsoluteTimeGetCurrent() - startTime)s")
print("")

startTime = CFAbsoluteTimeGetCurrent()
print("sumReduce()")
print("  = \(ar.sumReduce())")
print("  in \(CFAbsoluteTimeGetCurrent() - startTime)s")
```

Which would  result in:

```
sum()
  = 4999950000
  in 6.78056800365448s

sumReduce()
  = 4999950000
  in 0.0268669724464417s
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,11 +11,8 @@ All notable changes to this project will be documented in this file.
 > N/A
 >
 > ### Enhancements
-> - **Array**
->   - Improved performance of the `sum()` function.
-> - New **Data** extensions
->   - Added `chunks(of:)` to split a `Data` instance into preset sized chunks.
->
+> N/A
+> 
 > ### Bugfixes
 > N/A
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,11 +11,13 @@ All notable changes to this project will be documented in this file.
 > N/A
 >
 > ### Enhancements
-> N/A
+> - **Array**
+>   - Improved performance of the `sum()` function.
+> - New **Data** extensions
+>   - Added `chunks(of:)` to split a `Data` instance into preset sized chunks.
 >
 > ### Bugfixes
 > N/A
-
 
 # v4.0.1
 
--- a/Sources/Extensions/SwiftStdlib/ArrayExtensions.swift
+++ b/Sources/Extensions/SwiftStdlib/ArrayExtensions.swift
@@ -17,7 +17,11 @@ public extension Array where Element: Numeric {
 	///
 	/// - Returns: sum of the array's elements.
 	public func sum() -> Element {
-        return self.reduce(0 as Element, +)
+        var total: Element = 0
+        for i in 0..<count {
+            total += self[i]
+        }
+        return total
 	}
 	
 }
@@ -32,7 +36,10 @@ public extension Array where Element: FloatingPoint {
 	/// - Returns: average of the array's elements.
 	public func average() -> Element {
         guard !isEmpty else { return 0 }
-        let total: Element = sum()
+        var total: Element = 0
+        for i in 0..<count {
+            total += self[i]
+        }
         return total / Element(count)
 	}
 
--- a/Sources/Extensions/SwiftStdlib/ArrayExtensions.swift
+++ b/Sources/Extensions/SwiftStdlib/ArrayExtensions.swift
@@ -13,13 +13,11 @@ public extension Array where Element: Numeric {
 	
 	/// SwifterSwift: Sum of all elements in array.
 	///
-	///		[1, 2, 3, 4, 5].sum -> 15
+	///		[1, 2, 3, 4, 5].sum() -> 15
 	///
 	/// - Returns: sum of the array's elements.
 	public func sum() -> Element {
-        var total: Element = 0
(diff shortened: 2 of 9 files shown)


Tests/SwiftStdlibTests/SignedNumericExtensionsTests.swift
```diff
@@ -17,7 +17,6 @@ class SignedNumericExtensionsTests: XCTestCase {
 	
 	func testAsLocaleCurrency() {
 		let number: Double = 3.2
-		print(number.asLocaleCurrency)
```
SD10: Thanks 👍

Sources/Extensions/SwiftStdlib/ArrayExtensions.swift
```diff
@@ -29,13 +27,12 @@ public extension Array where Element: FloatingPoint {
...
 	///
-	///		[1.2, 2.3, 4.5, 3.4, 4.5].average = 3.18
+	///		[1.2, 2.3, 4.5, 3.4, 4.5].average() = 3.18
 	///
 	/// - Returns: average of the array's elements.
 	public func average() -> Element {
         guard !isEmpty else { return 0 }
-        var total: Element = 0
-        forEach { total += $0 }
+        let total: Element = sum()
```
SD10: We don't allow dependencies between SwifterSwift extensions 😕

Sources/Extensions/SwiftStdlib/ArrayExtensions.swift
```diff
@@ -13,13 +13,11 @@ public extension Array where Element: Numeric {
...
 	///
-	///		[1, 2, 3, 4, 5].sum -> 15
+	///		[1, 2, 3, 4, 5].sum() -> 15
 	///
 	/// - Returns: sum of the array's elements.
 	public func sum() -> Element {
-        var total: Element = 0
-        forEach { total += $0 }
-        return total
+        return self.reduce(0 as Element, +)
```
SD10: We probably want to use `reduce(into:)` because we're dealing with a copy-on-write type

basvankuijck:
You're right on the playground part, was jumping to conclusions a bit too fast.

I moved the benchmark tests to a unit-test and reran it:
https://gist.github.com/basvankuijck/35a399f082d175e9c49bd65ac15f27f0
In the end `total += self[i]` would give the best result as suggested by @LucianoPAlmeida

SD10:
@basvankuijck I’m just curious why you’re not using XCTest measure blocks

basvankuijck:
@SD10: No particular reason, apart from the fact that my Xcode doesn't store baseline values in some (yet unknown) cases. Which results into failed tests and me getting headaches. 
So I written a small command line .swift file in order to do some quick proof-of-concept

CHANGELOG.md
```diff
@@ -11,12 +11,14 @@ All notable changes to this project will be documented in this file.
 > N/A
 >
 > ### Enhancements
-> N/A
+> - **Array**
+>   - Improved performance of the `sum()` function.
+> - New **Data** extensions
+>   - Added `chunks(of:)` to split a `Data` instance into preset sized chunks.
```
SD10: @basvankuijck Can you remove this from the CHANGELOG?

SwifterSwift.xcodeproj/xcshareddata/xcschemes/SwifterSwift-iOS.xcscheme
```diff
@@ -26,7 +26,8 @@
       buildConfiguration = "Debug"
       selectedDebuggerIdentifier = "Xcode.DebuggerFoundation.Debugger.LLDB"
       selectedLauncherIdentifier = "Xcode.DebuggerFoundation.Launcher.LLDB"
-      language = ""
+      language = "en"
+      region = "US"
```
SD10: Any reason why we've modified language/region here? I don't think its a big problem but curious if we can throw it out

basvankuijck: I'm from the Netherlands, so for instance the `FloatExtensionsTests.testAsLocaleCurrency()` test would fail on my machine, because the result would not contain `10.23`:
```swift
XCTAssert(num.asLocaleCurrency.contains("\(num)"))
```
Actual result: `"€10,23"`

SD10: Thanks for clarifying @basvankuijck. I think we had this problem before and I'm wondering if we just ignored it. Looks fine to me

omaralbeik:
Thank you guys 🎉

PR #290, opened 2017-10-20
add semicolons to lastIndex(of:) and firstIndex(of:) doc

zinthemoney:
## Checklist

### DIFF
--- a/Sources/Extensions/SwiftStdlib/ArrayExtensions.swift
+++ b/Sources/Extensions/SwiftStdlib/ArrayExtensions.swift
@@ -523,9 +523,9 @@ public extension Array where Element: Equatable {
 	
 	/// SwifterSwift: First index of a given item in an array.
 	///
-	///		[1, 2, 2, 3, 4, 2, 5].firstIndex(of 2) -> 1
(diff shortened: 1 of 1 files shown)


wdcurry:
May i ask why firstIndex is deprecated? It is the simplest of checks, is logical, is intuitive. Flipping to a StringProtocal and range is just overkill, like numerous other Swift "improvements". Loving the language for the most part, but it is starting to become a bit "mySpace-y" in some areas.

guykogus:
I'm guessing it's because using `Int`s for the indices of characters in a `String` is a bad idea, that's not how `String`s are built anymore. You should be using `String.Index` values instead, which is what's provided by the functions mentioned in the deprecation message.
I'm open to being convinced otherwise. What's your use case where you need these functions?

wdcurry:
Fair enough, I still do strings old-school, learning to remap my brain a bit. Funny i rely on SwifterSwift so much that i forget to check and just pushed String.contains() into my code for a simple Bool return, which works well, apart from a few minor edits. I should have found that before asking ;)

SD10:
https://developer.apple.com/documentation/swift/string/2997120-firstindex?changes=latest_minor

It's part of the Swift Standard Library now

PR #296, opened 2017-10-22
add optional assignment operator

buddax2:
add optional assignment operator: if `rhs` is not `nil`, assign it to `lhs`.

## Checklist

### DIFF
--- a/Sources/Extensions/SwiftStdlib/OptionalExtensions.swift
+++ b/Sources/Extensions/SwiftStdlib/OptionalExtensions.swift
@@ -50,7 +50,7 @@ public extension Optional {
 // MARK: - Operators
 infix operator ??= : AssignmentPrecedence
 
-/// SwifterSwift: If `rhs` is not `nil`, assign it to `lhs`.
+/// SwifterSwift: Assign an optional value to a variable only if the value is not nil.
 ///
 ///     let someParameter: String? = nil
 ///     let parameters = [String:Any]() //Some parameters to be attached to a GET request
--- a/Sources/Extensions/SwiftStdlib/OptionalExtensions.swift
+++ b/Sources/Extensions/SwiftStdlib/OptionalExtensions.swift
@@ -46,3 +46,20 @@ public extension Optional {
 	}
 	
 }
+
+// MARK: - Operators
+infix operator ??= : AssignmentPrecedence
+
+/// SwifterSwift: If `rhs` is not `nil`, assign it to `lhs`.
+///
+///     let someParameter: String? = nil
+///     let parameters = [String:Any]() //Some parameters to be attached to a GET request
+///     parameters[someKey] ??= someParameter //It won't be added to the parameters dict
+///
+/// - Parameters:
+///   - lhs: Any?
+///   - rhs: Any?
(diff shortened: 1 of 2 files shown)


Sources/Extensions/SwiftStdlib/OptionalExtensions.swift
```diff
@@ -46,3 +46,20 @@ public extension Optional {
 	}
 	
 }
+
+// MARK: - Operators
+infix operator ??= : AssignmentPrecedence
+
+/// SwifterSwift: If `rhs` is not `nil`, assign it to `lhs`.
```
omaralbeik: I think **Assign an optional value to a variable only if the value is not nil** makes more sense here :)

buddax2:
@omaralbeik done. Thanks for the comment

Sources/Extensions/SwiftStdlib/OptionalExtensions.swift
```diff
@@ -46,3 +46,20 @@ public extension Optional {
 	}
```
LucianoPAlmeida: Hey @buddax2 , Just as a suggestion I think will be better put this extension as an Optional extension instead of global generic T. I mean, put this as an operator of Optional. Something like:
```swift 
// MARK: - Operators
infix operator ??= : AssignmentPrecedence

extension Optional {

    public static func ??= (lhs: inout Optional, rhs: Optional) {
        guard let rhs = rhs else { return }
        lhs = rhs
    }
}

```
I think it will provide the same behavior and makes more sense since the operator is applied to optional values only :)

SD10: Good thoughts 👍

buddax2: @LucianoPAlmeida great. Done

SD10: Another thing, have we thought about using `||=` for the operator?

buddax2: @SD10 for me personally `??=` looks more like optional rather than `||=`.

SD10: @buddax2 I was only suggesting this because I think other languages use the `||=` operator for this functionality. Thus, it would be more familiar. When I saw `??=` I had no idea what it did until I looked at the code

buddax2: @SD10 if I'm not mistaken `||=` works slightly in a different way for example in Ruby. It assigns new value only if there was no value in the var before. So `a ||= b` is more like `a || a = b`. 

But here in our case we want to assign `b` only in case if `b != nil`

SD10: @buddax2 I see what you mean, thanks for clarifying 👍

LucianoPAlmeida: @buddax2 Thank's for the changes :)

Issue #294, opened 2017-10-22
Check all extensions for cross-dependancies


omaralbeik:
The idea behind SwifterSwift is to have the widest collection of extensions grouped in one place, and at the same time give users the ability to copy any extension and use it as an independent unit in their code.

We want to double check one final time that all extensions follow the above pattern and the scope of each unit in inclosed within it.

One suggested approach to do this, is for each file in the framework:
- copy contents to a Swift playground
- see if Xcode complain about any error like `Type 'some_type' has no member 'some_extension'`
- finally check each unit to make sure its scope is inclosed within it

omaralbeik:
#286 can be addressed here as well

kaphacius:
@omaralbeik by each unit, do you mean that each extension has to be self-sufficient?

omaralbeik:
By unit I mean any variable or function inside an extension
example:

```
extension type_1 {
  var a {
    return ...
  }

  func b() {
    ...
    return ...
  }
}
```

```
extension type_2 {
  var c {
    return ...
  }

  func d() {
    ...
    return ...
  }
}
```

here `a`, `b`, `c` and `d` are units,
`c` for example should not depend on `d`, `a` any other unit

Issue #286, opened 2017-10-12
Some UIKit/Cocoa coupled dependecies on the Stdlib and Fondation modules 


LucianoPAlmeida:
[x] I've read, understood, and done my best to follow the [Contributing guidelines](https://github.com/SwifterSwift/SwifterSwift/tree/master/CONTRIBUTING.md) before opening this issue.
Hey guys @omaralbeik @SD10 :)
There is some Stdlib and Foundation files that are importing UIKit/Cocoa modules. I think this couple the Stdlib and foundantion when you have for exemple a String extension that takes an UIKit class as parameter. Especially now that SwifterSwift have subspecs, even if i only use the stdlib subspec, because of this coupled dependency import i end up the UIKit module also. 
Opening this to see what you guys think :)

PR #297, opened 2017-10-24
Remove cross references #294

stupergenius:
## Checklist

## Description
If I understood #294 correctly, then this should take care of cross-file references at least. I'm not sure of the issue with #286 since none of my findings include any UIKit extension references in the Foundation/Swift extension groups.

What this does _not_ do, currently, is intra-file references to other extensions since this is a difficult one to automate (see below). Specifically, I'm wondering if examples like in [NSAttributedString](https://github.com/SwifterSwift/SwifterSwift/blob/master/Sources/Extensions/Foundation/NSAttributedStringExtensions.swift#L51) would need to be refactored (since extensions like `.bolded` use the helper method `applying`)? Or if just examples like [NSURL](https://github.com/SwifterSwift/SwifterSwift/blob/master/Sources/Extensions/Foundation/URLExtensions.swift#L37) should be refactored since it's a public extension method that proxies to another? Some guidance here would be appreciated.

Side note: I created a small script that tries to compile each source file individually and emits any errors from the swift compiler to help automate this process. Perhaps this could be cleaned up and added to the Danger file with a nicer report (likely as a separate enhancement PR)?

```bash
#!/bin/sh

platformName="$1"

if [ "$platformName" == "" ]; then
	platformName="ios"
fi

if [ "$platformName" == "ios" ]; then
	for sourceFile in Sources/Extensions/**/*.swift; do
		swiftc -target arm64-apple-ios11.0 \
			-sdk $(xcrun --sdk iphoneos --show-sdk-path) \
			-emit-object $sourceFile \
			-o "${sourceFile%.*}.o"
		rm -f "${sourceFile%.*}.o"
		if [ ! $? -eq 0 ]; then
			echo "$sourceFile could not compile"
		fi
	done
fi

if [ "$platformName" == "macos" ]; then
	for sourceFile in Sources/Extensions/**/*.swift; do
		swiftc -target x86_64-apple-macos10.11 \
			-sdk $(xcrun --sdk macosx --show-sdk-path) \
			-emit-object $sourceFile \
			-o "${sourceFile%.*}.o"
		rm -f "${sourceFile%.*}.o"
		if [ ! $? -eq 0 ]; then
			echo "$sourceFile could not compile"
		fi
	done
fi
```

### DIFF
--- a/Sources/Extensions/Foundation/DataExtensions.swift
+++ b/Sources/Extensions/Foundation/DataExtensions.swift
@@ -6,11 +6,7 @@
 //  Copyright © 2016 Omar Albeik. All rights reserved.
 //
 
-#if os(macOS)
-	import Cocoa
-#else
-	import UIKit
-#endif
+import Foundation
 
 // MARK: - Properties
 public extension Data {
--- a/Sources/Extensions/Foundation/DateExtensions.swift
+++ b/Sources/Extensions/Foundation/DateExtensions.swift
@@ -788,7 +788,7 @@ public extension Date {
 	public init?(iso8601String: String) {
 		// https://github.com/justinmakaila/NSDate-ISO-8601/blob/master/NSDateISO8601.swift
 		let dateFormatter = DateFormatter()
-		dateFormatter.locale = .posix
+		dateFormatter.locale = Locale(identifier: "en_US_POSIX")
 		dateFormatter.timeZone = TimeZone.current
 		dateFormatter.dateFormat = "yyyy-MM-dd'T'HH:mm:ss.SSSZ"
 		if let date = dateFormatter.date(from: iso8601String) {
--- a/Sources/Extensions/Shared/ColorExtensions.swift
+++ b/Sources/Extensions/Shared/ColorExtensions.swift
@@ -85,10 +85,12 @@ public extension Color {
 	
 	/// SwifterSwift: Hexadecimal value string (read-only).
 	public var hexString: String {
-		let r = rgbComponents.red
-		let g = rgbComponents.green
-		let b = rgbComponents.blue
-		return String(format: "#%02X%02X%02X", r, g, b)
+        let components: [Int] = {
+            let c = cgColor.components!
+            let components = c.count == 4 ? c : [c[0], c[0], c[0], c[1]]
+            return components.map { Int($0 * 255.0) }
+        }()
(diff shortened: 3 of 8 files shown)


SD10:
Hey @stupergenius, thanks for this 👍 

I'm not going pretend that I understand the script. I'll have to look at it a bit more closely. It would be nice if you could integrate this with Danger like you mentioned.

You asked for a little more context:

Ultimately, we're trying to reduce the remove the dependency of any SwifterSwift extension on another SwifterSwift extension. It doesn't matter if both extensions are on the same type, in the same file, or in the same module. 

*Each SwifterSwift extension should be able to stand alone if you were to copy & paste it into a project.*

stupergenius:
Thanks @SD10. Am I on the right track with 73c1dc1? If so... this kind of use is quite prevalent throughout SwifterSwift so this effort might take some time... and it perhaps goes against the norm of DRY coding styles.

I'm happy to continue along the same lines if this is how the project is intended to be structured, but without a linter capable of preventing such cross use (or a diligent code review team), this style might be difficult to maintain going forward.

SD10:
@stupergenius Yeah this is the right track. We don't need to do everything all at once if there are many instances where this structure occurs. I'd be willing to accept this PR -- but we have to take care of the  build errors.

As for not following the DRY principle, this is intentional. A common use case of SwifterSwift is using this project as a reference/collection of code snippets. We want to have each extension stand alone because it gets confusing when you want to copy & paste a single extension into your project -- then have to hunt down all the other dependencies.

stupergenius:
So... I don't really know how to pass these coverage checks if the last commit doesn't push up the coverage enough. The lines that aren't covered are... very difficult to actually get covered (they're primarily graphics context methods that return nil in certain scenarios that don't happen in our SwifterSwift's use of them). They were also never actually covered, but by duplicating some methods it reduces coverage since there's now more overall code to cover.

Suggestions?

SD10:
@stupergenius Don't worry about trivial code coverage deltas 😆 We just try to write tests to the best of our ability. The only check that matters is Travis passing 👍 I should set this up so that Travis is required and codecov is optional, thus making it more explicit

stupergenius:
Heh ok whew. Was about to get creative there. 😂 Otherwise this is ready to merge from my perspective. It still does not do the intra-file de-referencing, but it at least knocks out some cross-file references and provides context for later.

PR #300, opened 2017-10-27
Adding padStart/padEnd and byPaddingStart/byPaddingEnd String extensions

LucianoPAlmeida:
## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,11 @@ All notable changes to this project will be documented in this file.
     - Add `padStart(lenght:with:)` and `padEnd(lenght:with:)` to pad the string to a lenght on the start or end.
     - Add `paddingStart(lenght:with:)` and `paddingEnd(lenght:with:)` to return a padding string to a lenght on the start or end. 
       [#300](https://github.com/SwifterSwift/SwifterSwift/pull/300) by [LucianoPAlmeida](https://github.com/LucianoPAlmeida)
+ - New **NSImage** extensions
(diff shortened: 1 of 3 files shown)


Alex293:
Hi

Nice addition ! Though I think naming could be rethink to match existing api. 
byPadding -> paddedstart(by:) for example. 

As there is only two possibilities with a enum for end/start we could have 

String.padded(by: 20, .start) or padded(by: 20, at: .start)

We use something like this in our project I’ll post our naming if it’s different.

LucianoPAlmeida:
Hey @Alex293 I agree that the naming can be improved :)
Thank you for the suggestion :)

LucianoPAlmeida:
@SD10 No problem man :) 
About more generic, I think now in swift 4 Collection conformance was restored on String. Maybe it can be something on a collection level, but I don't know if it's worth and also don't know what are the constraints on doing it with Collection. I think that something like this is best fit for String, but Collection it's also an option :)

CHANGELOG.md
```diff
@@ -10,9 +10,12 @@ All notable changes to this project will be documented in this file.
 > ### API Breaking
 > N/A
 >
-> ### Enhancements
-> N/A
-> 
+
+### Enhancements
+ - New **String** extensions: [#300](https://github.com/SwifterSwift/SwifterSwift/pull/300) by [LucianoPAlmeida](https://github.com/LucianoPAlmeida).
```
SD10: It would be better to move the `#300 by LucianoPalmeida` to the bottom of all the bullet points. (In case someone else adds more String extensions). I see some other CHANGELOG entries have this wrong format too. You can fix them up in this PR if you'd like 😅

LucianoPAlmeida: @SD10 Fixed :)

omaralbeik:
Thank you guys 💯

PR #305, opened 2017-11-05
Removing deprecated string.characters calls and xcode 9.1 recommend settings.

LucianoPAlmeida:
Xcode 9.1 project recommended settings
Removing deprecated string.characters calls.

## Checklist


Sources/Extensions/Shared/ColorExtensions.swift
```diff
@@ -92,7 +92,7 @@ public extension Color {
 	/// SwifterSwift: Short hexadecimal value string (read-only, if applicable).
 	public var shortHexString: String? {
 		let string = hexString.replacingOccurrences(of: "#", with: "")
-		let chrs = Array(string.characters)
+		let chrs = Array(string)
```
SD10: Won't this fail now?

Sources/Extensions/SwiftStdlib/StringExtensions.swift
```diff
@@ -39,7 +39,7 @@ public extension String {
 	/// SwifterSwift: Array of characters of a string.
 	///
 	public var charactersArray: [Character] {
-		return Array(characters)
+		return Array(self)
```
SD10: Do we get an array of characters here or an array of a Single string?

SD10: Guess I have to read about the new Strings 🤔 Implicit cast?

Sources/Extensions/Shared/ColorExtensions.swift
LucianoPAlmeida: I think not, it returns an array of characters the same way as before :)

Sources/Extensions/SwiftStdlib/StringExtensions.swift
LucianoPAlmeida: It returns an array of characters :)

LucianoPAlmeida: I think is because string now has collection conformance again, so it uses Array(s: Sequence) and returns a sequence of its characters. Did this test on playground
screen shot 2017-11-05 at 10 57 27

SD10: Can we do this? Forgive me I haven't had to deal with this yet 😃 But we may want to re-evaluate all the string extensions in general and see if some are obsolete
```Swift
var chars: [Characters] = "Swift"
```

LucianoPAlmeida: I think not, swift compiler complains about cannot convert String to [Character], 
screen shot 2017-11-05 at 11 31 13

SD10: @LucianoPAlmeida Appreciate you testing that! It's very hard for me to run a playground on my slow iMac 😭

Sources/Extensions/SwiftStdlib/StringExtensions.swift
```diff
@@ -229,7 +229,7 @@ public extension String {
 	///		"Hello world!".length -> 12
 	///
 	public var length: Int {
```
SD10: We should probably deprecate this one. It's not even a good name because it can easily be confused with `NSString.length`

Sources/Extensions/SwiftStdlib/StringExtensions.swift
```diff
@@ -401,7 +401,8 @@ public extension String {
 	///
 	/// - Returns: The reversed string.
 	public func reversed() -> String {
```
SD10: We should remove this one too. We can just use `String(str.reversed())` if we even need `String` typing

Sources/Extensions/SwiftStdlib/StringExtensions.swift
```diff
@@ -575,7 +576,7 @@ public extension String {
 	/// - Parameter string: substring to search for.
 	/// - Returns: first index of substring in string (if applicable).
 	public func firstIndex(of string: String) -> Int? {
```
SD10: This is probably obsolete now that `String` is a `Collection`?
We can use `func index(of element: Self.Element) -> Self.Index?` I think?

Sources/Extensions/SwiftStdlib/StringExtensions.swift
```diff
@@ -620,10 +621,10 @@ public extension String {
 	///   - length: amount of characters to be sliced after given index.
 	/// - Returns: sliced substring of length number of characters (if applicable) (example: "Hello World".slicing(from: 6, length: 5) -> "World")
 	public func slicing(from i: Int, length: Int) -> String? {
```
SD10: Now we can subscript because it's a collection so this can probably go too 😅

Sources/Extensions/SwiftStdlib/StringExtensions.swift
```diff
@@ -684,10 +685,10 @@ public extension String {
 	/// - Parameter i: string index the slicing should start from.
 	/// - Returns: sliced substring starting from start index (if applicable) (example: "Hello world".slicing(at: 6) -> "world")
 	public func slicing(at i: Int) -> String? {
```
SD10: ||

Sources/Extensions/SwiftStdlib/StringExtensions.swift
```diff
@@ -710,7 +711,7 @@ public extension String {
 	/// - Parameter separator: separator to split string by.
 	/// - Returns: array of strings separated by given string.
 	public func splitted(by separator: Character) -> [String] {
```
SD10: Not sure about this one but it looks like `split(separator: Element)` is the way to go

LucianoPAlmeida:
@SD10 sure, I'll take a look at them as soon as possible :)

Sources/Extensions/SwiftStdlib/StringExtensions.swift
omaralbeik: 💯, also `String.count` can be used in Swift 4

Sources/Extensions/SwiftStdlib/StringExtensions.swift
```diff
@@ -386,7 +386,7 @@ public extension String {
 	///
 	/// - Returns: The most common character.
 	public func mostCommonCharacter() -> String {
```
omaralbeik: how come `mostCommonCharacter` is returning String instead of Character here 🤔

Sources/Extensions/SwiftStdlib/StringExtensions.swift
```diff
@@ -442,7 +443,7 @@ public extension String {
 	///
 	/// - Parameter i: index.
 	public subscript(safe i: Int) -> String? {
```
omaralbeik: Shouldn't this return `Character?` as well?

LucianoPAlmeida:
@omaralbeik @SD10 Working on those that should be deprecated :) How deprecation should be done? Just annotate the method? Annotate and move to another file? In case of renaming, should I remove the previous method?

Sources/Extensions/SwiftStdlib/StringExtensions.swift
LucianoPAlmeida: I don't know about this one. I think handles some safe convenience cases that maybe aren't done with the subscript. @SD10  @omaralbeik  What you think?

Sources/Extensions/SwiftStdlib/StringExtensions.swift
```diff
@@ -86,7 +86,7 @@ public extension String {
 	///		"".firstCharacterAsString -> nil
 	///
 	public var firstCharacterAsString: String? {
-		guard let first = characters.first else {
+		guard let first = self.first else {
```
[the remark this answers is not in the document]
LucianoPAlmeida: I think we can't rename first because is a native String property now that its conformance with Collection was restored.

Sources/Extensions/SwiftStdlib/StringExtensions.swift
```diff
@@ -210,7 +210,7 @@ public extension String {
 	///		"".lastCharacterAsString -> nil
 	///
 	public var lastCharacterAsString: String? {
-		guard let last = characters.last else {
+		guard let last = self.last else {
```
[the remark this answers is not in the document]
LucianoPAlmeida: Same thing like first above :)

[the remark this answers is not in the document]
SD10: We should probably just remove these as well 😅 `"Some string".last` and casting when we need it should be prefered

Sources/Extensions/SwiftStdlib/StringExtensions.swift
SD10: I think we should turn it into a safe subscript that accepts a `Range<String.Index>` and returns a `Substring?` then. The name of this method doesn't make sense, I'd expect to get an `ArraySlice` out of it. Since it has `ing` it also makes me think it's mutating and it's not 😅

omaralbeik:
@LucianoPAlmeida Here is an [example from `v3.2`](https://github.com/SwifterSwift/SwifterSwift/blob/swift-3.2/Sources/Extensions/Foundation/Deprecated/FoundationDeprecated.swift)

basically:
1- create a new directory inside SwiftStdlib called Deprecated
2- create a new SwiftStdlibDeprecated.swift file
3- move methods to deprecate to that file
4- mark them as deprecated
5- done!

Sources/Extensions/SwiftStdlib/StringExtensions.swift
LucianoPAlmeida: I don't know if it can be turned in a subscript with range because the second parameter is a length, not an end index 😅 I think worth just rename and returning a substring instead. @SD10 @omaralbeik  What you think?

SD10: That's a good point @LucianoPAlmeida 🤔 But I'm having trouble thinking of a place where someone would be using a length and not an `endIndex` considering they have a `startIndex`. You can always get the `endIndex` by incrementing the `startIndex` by `n`

LucianoPAlmeida: I agree so it will be better deprecating this?

SD10: Yeah, we can always replace it with the `subscript` version in another PR. Sorry to make so many changes. I should've thought about this before we even released SwifterSwift 4.0. It completely slipped my mind

LucianoPAlmeida: What about the mutating versions? Should be maintained?

SD10: @LucianoPAlmeida lets not make this change right now... I just opened the file and there are a ton of these `slice` methods. We also already have `subscripts` from ranges but they return `String?` and not `Substring?` It's too much for me to look at right now...

LucianoPAlmeida: @SD10 No problem we can open a issue/PR later with only this methods to discuss :)

Sources/Extensions/SwiftStdlib/StringExtensions.swift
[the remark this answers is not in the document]
omaralbeik: I meant `guard let firstCharacter = self.first else { ...`

omaralbeik:
@SD10 @LucianoPAlmeida Please let me know if we can merge this soon, as other PRs are depending on it. We'll probably look again on String extensions later

LucianoPAlmeida:
Hey @omaralbeik That's it for this one, I was just waiting for both of you guys review it :)

CHANGELOG.md
```diff
@@ -7,9 +7,14 @@ All notable changes to this project will be documented in this file.
...
-> N/A
->
+### API Breaking
+ - **String** 
+    - `length` is deprecated, use native `count` instead.
+    - `slicing(i:)` is deprecated, use `string[safe: i]` instead.
+    - `slicing(from: to:)` is deprecated, use `string[safe: start..<end]`
+    - `firstIndex(of:)` is deprecated, use the natives `index(of: Character)` or `range(of: StringProtocol)` instead.
+    - `splitted(by:)` is deprecated,  use the native `split(separator: )` instead.
+    - `reversed() -> String` is deprecated, use the Swift 4 new `reversed() -> ReversedCollection<String>`
```
SD10: @LucianoPAlmeida Please add the PR number as a reference under this part of the CHANGELOG

LucianoPAlmeida: Added :)

PR #313, opened 2017-11-10
Rename weekday to workweek

kaphacius:
## Checklist

As per [this wikipedia page](https://en.wikipedia.org/wiki/Workweek_and_weekend)

Another variant is `isAWorkday` or `isAWeekday`, but I think `isInWorkweek` works better with `isInWeekend`

### DIFF
--- a/Tests/FoundationTests/DateExtensionsTests.swift
+++ b/Tests/FoundationTests/DateExtensionsTests.swift
@@ -275,9 +275,9 @@ final class DateExtensionsTests: XCTestCase {
 		XCTAssertEqual(date.isInWeekend, Calendar.current.isDateInWeekend(date))
 	}
 	
-	func testisInWorkweek() {
+	func testisWorkday() {
 		let date = Date()
-		XCTAssertEqual(date.isInWorkweek, !Calendar.current.isDateInWeekend(date))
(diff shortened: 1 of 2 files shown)


omaralbeik:
Thank you @kaphacius, as mentioned in [Apple's documentation](https://developer.apple.com/documentation/foundation/nsdatecomponents/1410442-weekday), weekday refers to any day in the week, including weekend, which might be confusing here.

Personally I've never heard the word workweek before, but English is not my main language 😃
We might also consider renaming it to `isWorkDay` or `isBusinessDay`, what do you think guys?

SD10:
I see, after looking at the documentation I wouldn't be opposed to `isWorkday`

kaphacius:
done, `isWorkday` it is :)

Tests/FoundationTests/DateExtensionsTests.swift
```diff
@@ -275,9 +275,9 @@ final class DateExtensionsTests: XCTestCase {
 		XCTAssertEqual(date.isInWeekend, Calendar.current.isDateInWeekend(date))
 	}
 	
-	func testIsInWeekday() {
+	func testisWorkday() {
```
omaralbeik: @kaphacius Could you please rename this to `testIsWorkday` to keep consistency with other tests :)

Issue #333, opened 2017-12-07
Add an extension to get all attributes on an NSAttributedString


SD10:
Add an extension to get all attributes on an NSAttributedString:

```Swift
extension NSAttributedString {
    var attributes: [NSAttributedStringKey: Any] {
        /// implementation ...
    }
}
```

nathanbacon:
Hey Steven, I'm running into trouble with this. There are two types of possible strings this could be applied to:

One where the attributed is applied across the whole string as in
`NSAttributedString("Test").bolded`

and another where there are multiple attributes of the same type across the string
`NSAttributedString("Red").colored(with: UIColor.red) + NSAttributedString("Blue").colored(with: UIColor.blue)`

The problem with the second type is that the returned dictionary could only contain one color because both portions have the same NSAtributedStringKey. Returning an Array<(NSAttributedStringKey, Any, NSRange)> would be better for this case.

If this extension is meant to work on homogenous strings, the solution is easy. Let me know what you think.

Thanks

SD10:
Hey @nathanbacon!
I was only thinking of homogenous Strings so it is indeed an easy extension. If you think you have a solution for strings with multiple attributes under a single key, I’d look at that too

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

SD10:
Closed by #340

PR #340, opened 2017-12-20
Attributes getter for NSAttributedString Issue #333

nathanbacon:
This property gets the attributes of a simple NSAttributedString. It works by checking the attributes from the first character and assuming that applies for the rest of the string. Let me know if you'd take another approach. 

## Checklist

### DIFF
--- a/Tests/FoundationTests/NSAttributedStringExtensionsTests.swift
+++ b/Tests/FoundationTests/NSAttributedStringExtensionsTests.swift
@@ -118,4 +118,31 @@ final class NSAttributedStringExtensionsTests: XCTestCase {
 		XCTAssertEqual(filteredAttributes.count, 1)
 	}
 	#endif
+    
+    #if !os(macOS)
+    func testAttributes() {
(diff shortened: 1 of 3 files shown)


Tests/FoundationTests/NSAttributedStringExtensionsTests.swift
```diff
@@ -30,7 +30,7 @@ final class NSAttributedStringExtensionsTests: XCTestCase {
 	func testUnderlined() {
 		let string = NSAttributedString(string: "Underlined")
 		let out = string.underlined
-		let attributes = out.attributes(at: 0, effectiveRange: nil)
```
SD10: @nathanbacon Mind if we change these back? I don't want the unit tests of 1 extension to depend on another extension. All SwifterSwift extensions should be self sufficient

nathanbacon: @SD10 I see, taken care of.

PR #342, opened 2017-12-25
Date extensions

n0an:
Added Date() extension - initiazlier using Ineger literal. Provides Date init from Int like 20171225, or 2017_12_25
    
    * New extension added to DateExtensions.swift
    * Test added to DateExtensionsTests.swift

## Checklist

- [] All extensions are declared as **public**.

### DIFF
--- a/Sources/Extensions/Foundation/DateExtensions.swift
+++ b/Sources/Extensions/Foundation/DateExtensions.swift
@@ -958,11 +958,7 @@ public extension Date {
     public init?(integerLiteral value: Int) {
         let formatter = DateFormatter()
         formatter.dateFormat = "yyyyMMdd"
-        
-        if let date = formatter.date(from: String(value)) {
-            self = date
-        } else {
-            return nil
-        }
+        guard let date = formatter.date(from: String(value)) else { return nil }
+        self = date
     }
 }
--- a/Sources/Extensions/Foundation/DateExtensions.swift
+++ b/Sources/Extensions/Foundation/DateExtensions.swift
@@ -952,3 +952,11 @@ public extension Date {
 	}
 	
 }
+
+extension Date: ExpressibleByIntegerLiteral {
+    public init(integerLiteral value: Int) {
+        let formatter = DateFormatter()
+        formatter.dateFormat = "yyyyMMdd"
+        self = formatter.date(from: String(value)) ?? Date()
+    }
+}
--- a/Sources/Extensions/Foundation/DateExtensions.swift
+++ b/Sources/Extensions/Foundation/DateExtensions.swift
@@ -950,17 +950,19 @@ public extension Date {
 	public init(unixTimestamp: Double) {
 		self.init(timeIntervalSince1970: unixTimestamp)
 	}
-	
-}
-
-extension Date: ExpressibleByIntegerLiteral {
+    
     /// SwifterSwift: Create date object from Int literal
     ///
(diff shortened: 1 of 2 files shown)


n0an:
Could you help, why codecov telling about tests coverage? I've added unit test for this new Date extension initializer.

Sources/Extensions/Foundation/DateExtensions.swift
```diff
@@ -952,3 +952,15 @@ public extension Date {
 	}
 	
 }
+
+extension Date: ExpressibleByIntegerLiteral {
```
omaralbeik: Please mark this as `public`

Sources/Extensions/Foundation/DateExtensions.swift
```diff
@@ -952,3 +952,15 @@ public extension Date {
...
+    /// SwifterSwift: Create date object from Int literal
+    ///
+    ///     let date = Date(inegerLiteral: 2017_12_25) // "2017-12-25 00:00:00 +0000"
+    /// - Parameter value: Int value, e.g. 20171225, or 2017_12_25 etc.
+    public init(integerLiteral value: Int) {
```
omaralbeik: I'd like this to be a failable initializer `init?(integerLiteral value: Int)`

Sources/Extensions/Foundation/DateExtensions.swift
```diff
@@ -952,3 +952,15 @@ public extension Date {
...
+        let formatter = DateFormatter()
+        formatter.dateFormat = "yyyyMMdd"
+        self = formatter.date(from: String(value)) ?? Date()
```
omaralbeik: Return `nil` if can't make a `Date` object here

Sources/Extensions/Foundation/DateExtensions.swift
n0an: @omaralbeik , we can't make it public when using protocol conformance. I've made it public, by removing protocol conformance

Sources/Extensions/Foundation/DateExtensions.swift
n0an: If we conform to ExpressibleByIntegerLiteral protocol, this init is predefined and it's not failable init. I've made it failable, removing protocol adoption.

Sources/Extensions/Foundation/DateExtensions.swift
n0an: Made it.

Sources/Extensions/Foundation/DateExtensions.swift
```diff
@@ -950,5 +950,19 @@ public extension Date {
...
+    
+    /// SwifterSwift: Create date object from Int literal
+    ///
+    ///     let date = Date(integerLiteral: 2017_12_25) // "2017-12-25 00:00:00 +0000"
+    /// - Parameter value: Int value, e.g. 20171225, or 2017_12_25 etc.
+    public init?(integerLiteral value: Int) {
+        let formatter = DateFormatter()
+        formatter.dateFormat = "yyyyMMdd"
+        
+        if let date = formatter.date(from: String(value)) {
```
omaralbeik: I think `guard` makes the code more readable here :)
```swift
guard let date = formatter.date(from: String(value)) else { return nil }
self = date
```

n0an: Agree :). Changed to guard-let

omaralbeik:
Thank you @n0an 🎉

PR #347, opened 2017-12-31
UIDevice countryCode extension added

n0an:
Extension provides countryCode for current device as String

## Checklist

### DIFF
--- a/Sources/Extensions/UIKit/UIDeviceExtensions.swift
+++ b/Sources/Extensions/UIKit/UIDeviceExtensions.swift
@@ -10,7 +10,7 @@
 import UIKit
 public extension UIDevice {
     /// Show country code of current locale for device
-    static var countryCode: String? {
+    static var countryISOCode: String? {
         return (Locale.current as NSLocale).object(forKey: NSLocale.Key.countryCode) as? String
     }
 }
--- a/Sources/Extensions/UIKit/UIDeviceExtensions.swift
+++ b/Sources/Extensions/UIKit/UIDeviceExtensions.swift
@@ -1,0 +1,17 @@
+//
+//  UIDeviceExtensions.swift
+//  SwifterSwift
+//
+//  Created by Anton Novoselov on 31/12/2017.
+//  Copyright © 2017 SwifterSwift
+//
+
+#if os(iOS) || os(tvOS)
+import UIKit
+public extension UIDevice {
+    /// Show country code of current locale for device
+    static var countryCode: String? {
+        return (Locale.current as NSLocale).object(forKey: NSLocale.Key.countryCode) as? String
+    }
+}
+#endif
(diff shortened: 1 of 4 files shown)


n0an:
Hi, @omaralbeik . You've got 'US' on the simulator, because US region was chosen on your simulator at the moment, so it's correct behavior. Why did you expect +1 or 1?
Yes, it's locale extensions, but I placed it in UIDevice, because it's kind of quick way to understand country code of region where device located. It's useful for situations for example, where you should select some localization for the UI, and asking user permission to get access to GPS coordinates can be a bit overkill. In this case you can just get country code from current locale. And I suppose it's pretty handy to get it as static property of device :)
If you want, I will refactor it to Locale extension.

omaralbeik:
I see, sorry I though we were talking about country phone code, as [this link](https://countrycode.org/) suggessts, it might be less confusing to rename the function to `countryISOCode`

n0an:
@omaralbeik , changed function name to `countryISOCode`

Sources/Extensions/UIKit/UIDeviceExtensions.swift
```diff
@@ -0,0 +1,17 @@
...
+//  SwifterSwift
+//
+//  Created by Anton Novoselov on 31/12/2017.
+//  Copyright © 2017 SwifterSwift
+//
+
+#if os(iOS) || os(tvOS)
+import UIKit
+public extension UIDevice {
+    /// Show country code of current locale for device
```
SD10: Need to prefix this with `SwifterSwift`. Please read [Adding new extensions, Adding documentation](https://github.com/SwifterSwift/SwifterSwift/blob/master/CONTRIBUTING.md#adding-new-extensions). Use existing extensions as a reference.

Sources/Extensions/UIKit/UIDeviceExtensions.swift
```diff
@@ -0,0 +1,17 @@
...
+    static var countryISOCode: String? {
```
LucianoPAlmeida: Hey @n0an, @omaralbeik, @SD10 would it make more sense being an instance property? To me it make more sense to someone using the API use `UIDevice.current.countryISOCode`  than `UIDevice.countryISOCode` since I want the current device code. It seems more verbose but I think it's more readable and make more sense. Just a suggestion, What you guys think? :)

SD10: I agree with @LucianoPAlmeida 👍 Welcome back!

omaralbeik: Welcome back @LucianoPAlmeida!
I agree with you about making it an instance property, but I still think that the right place for this extension is `Locale`

LucianoPAlmeida: @omaralbeik I agree, this makes more sense as a `Locale` extension :)

SD10:
@n0an Mind updating this to be an extension on `Locale`?

SD10:
Going to close this due to inactivity. Please feel free to reopen with the requested changes.

n0an:
Have been busy with repairing new flat and moving to flat. Will figure out where should I start to contribute soon :)

Sources/Extensions/UIKit/UIDeviceExtensions.swift
n0an: I don't mind. Moved to Locale extensions.

PR #352, opened 2018-01-03
Shorten sum and average methods for array

nathanbacon:
Updated sum and average methods to use fewer lines of code but maintain readability

### DIFF
--- a/Sources/Extensions/SwiftStdlib/ArrayExtensions.swift
+++ b/Sources/Extensions/SwiftStdlib/ArrayExtensions.swift
@@ -30,7 +30,7 @@ public extension Array where Element: FloatingPoint {
 	/// - Returns: average of the array's elements.
 	public func average() -> Element {
         guard !isEmpty else { return 0 }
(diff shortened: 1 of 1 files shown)


SD10:
Hey @nathanbacon, thank you for submitting this PR!

However, the reason we are using a `for loop` here is for performance reasons. In short, the `currentResult` argument for the `reduce` closure is immutable and therefore a copy is created during each iteration. This gets very expensive when working with arrays. 

One of our members @LucianoPAlmeida wrote more on the subject here:
https://medium.com/@lucianoalmeida1/a-little-bit-about-the-cool-reduce-in-swift-306edd9ceb57

Personally, I believe that a normal `for loop` is the fastest

nathanbacon:
Ah, I see. I was thinking there might have been a reason for it. Learned something new 😀

LucianoPAlmeida:
hey @nathanbacon, sorry a little late but just if you curiuos about this there is a discussion about that on #285 :)

PR #385, opened 2018-02-01
Extensions to UITableView, UICollectionView, UserDefaults, and Double

jason-ingenuity:
## Checklist

### DIFF
--- a/Sources/Extensions/Foundation/UserDefaultsExtensions.swift
+++ b/Sources/Extensions/Foundation/UserDefaultsExtensions.swift
@@ -37,5 +37,33 @@ public extension UserDefaults {
 	public func date(forKey key: String) -> Date? {
 		return object(forKey: key) as? Date
 	}
+    
+    /// SwifterSwift: Retrieves a Codable object from UserDefaults.
+    ///
+    /// - Parameters:
+    ///   - type: Class that conforms to the Codable protocol.
+    ///   - key: key to find the Codable object for.
+    /// - Returns: Codable object for key (if exists).
+    public func codable<T: Codable>(_ type: T.Type, with key: String) -> Codable? {
+        if let data = UserDefaults.standard.value(forKey: key) as? Data {
+            let object = try? JSONDecoder().decode(type.self, from: data)
+            
+            return object
+        }
+        
+        return nil
+    }
+    
+    /// SwifterSwift: Allows storing of Codable objects to UserDefaults
+    ///
+    /// - Parameters:
+    ///   - codable: Codable object to store.
+    ///   - key: Object identifier
+    public func set<T: Codable>(codable: T, forKey key: String) {
+        let data = try? JSONEncoder().encode(codable)
+        
+        set(data, forKey: key)
+        synchronize()
+    }
 	
 }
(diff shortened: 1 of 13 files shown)


Sources/Extensions/Foundation/UserDefaultsExtensions.swift
```diff
@@ -37,5 +37,33 @@ public extension UserDefaults {
...
 		return object(forKey: key) as? Date
 	}
+    
+    /// SwifterSwift: Retrieves a Codable object from UserDefaults.
+    ///
+    /// - Parameters:
+    ///   - type: Class that conforms to the Codable protocol.
+    ///   - key: key to find the Codable object for.
+    /// - Returns: Codable object for key (if exists).
+    public func codable<T: Codable>(_ type: T.Type, with key: String) -> Codable? {
```
SD10: I think using some name like `object` would be better? The function signature gives context to the fact that it is codable.

Sources/Extensions/SwiftStdlib/DoubleExtensions.swift
```diff
@@ -25,9 +26,80 @@ public extension Double {
 	public var cgFloat: CGFloat {
 		return CGFloat(self)
 	}
+    
+    /// SwifterSwift: Returns the current double (treated as meters) to kilometers.
+    /// Based on Apple's [The Swift Programming Language (Swift 4.0.3) - Extensions](https://developer.apple.com/library/content/documentation/Swift/Conceptual/Swift_Programming_Language/Extensions.html#//apple_ref/doc/uid/TP40014097-CH24-ID152)
```
SD10: I think this is fine to list Apple as a source, but let's not expose them in the documentation

jason-ingenuity: Noted. Will insert the link in the property body instead. Thanks

Sources/Extensions/Foundation/UserDefaultsExtensions.swift
jason-ingenuity: Realizing that, I think it makes sense. Will update it accordingly. Also, regarding your general comment (I don't know how to reply to that comment so I'll do it here instead), I've read the Contribution Guidelines but I just overlooked the checkbox for that 😄 . Would you like for me to break down this PR into smaller chunks?

SD10: @jason-ingenuity If it's not too much trouble can you make it 3 PRs? Defaults, Double, and UICollectionView/UITableView? This makes it easier to: a) Review it b) Accept the PR without rejecting the whole thing from a single flawed area

Sources/Extensions/Foundation/UserDefaultsExtensions.swift
```diff
@@ -37,5 +37,33 @@ public extension UserDefaults {
...
 	}
+    
+    /// SwifterSwift: Retrieves a Codable object from UserDefaults.
+    ///
+    /// - Parameters:
+    ///   - type: Class that conforms to the Codable protocol.
+    ///   - key: key to find the Codable object for.
+    /// - Returns: Codable object for key (if exists).
+    public func object<T: Codable>(_ type: T.Type, with key: String) -> Codable? {
+        if let data = UserDefaults.standard.value(forKey: key) as? Data {
```
LucianoPAlmeida: Hey @jason-ingenuity, since you are defining this as a UserDefaults exetension, you don't need to call the shared instance like `UserDefaults.standard.value(forKey: key)`, you can just call the instance method directly `self.value(forKey: key)` :)

jason-ingenuity: Yes, I overlooked it. Thanks.

PR #386, opened 2018-02-02
UITableView and UICollectionView extensions

jason-ingenuity:
## Checklist

### DIFF
--- a/Tests/UIKitTests/UICollectionViewExtensionsTests.swift
+++ b/Tests/UIKitTests/UICollectionViewExtensionsTests.swift
@@ -53,6 +53,15 @@ final class UICollectionViewExtensionsTests: XCTestCase {
 			XCTAssert(completionCalled)
 		}
 	}
+    
+    #if os(iOS)
+    func testRegisterCellWithNibUsingClass() {
+        let indexPath = IndexPath(row: 0, section: 0)
+        collectionView.register(nibWithCellClass: UICollectionViewCell.self, at: UICollectionViewExtensionsTests.self)
+        let cell = collectionView.dequeueReusableCell(withClass: UICollectionViewCell.self, for: indexPath)
+        XCTAssertNotNil(cell)
+    }
+    #endif
 	
 }
 
--- a/Tests/UIKitTests/UICollectionViewExtensionsTests.swift
+++ b/Tests/UIKitTests/UICollectionViewExtensionsTests.swift
@@ -54,14 +54,12 @@ final class UICollectionViewExtensionsTests: XCTestCase {
 		}
 	}
     
-    #if os(iOS)
     func testRegisterCellWithNibUsingClass() {
         let indexPath = IndexPath(row: 0, section: 0)
         collectionView.register(nibWithCellClass: UICollectionViewCell.self, at: UICollectionViewExtensionsTests.self)
         let cell = collectionView.dequeueReusableCell(withClass: UICollectionViewCell.self, for: indexPath)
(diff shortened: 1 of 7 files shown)


Tests/UIKitTests/UICollectionViewExtensionsTests.swift
```diff
@@ -53,6 +53,15 @@ final class UICollectionViewExtensionsTests: XCTestCase {
 			XCTAssert(completionCalled)
 		}
 	}
+    
+    #if os(iOS)
```
SD10: Is this necessary? I don't use`tvOS` so I'm not sure if it is available. This whole file is wrapped for `iOS` || `tvOS`

Tests/UIKitTests/UITableViewExtensionsTests.swift
```diff
@@ -133,6 +133,16 @@ final class UITableViewExtensionsTests: XCTestCase {
 		XCTAssertNotNil(cell)
 	}
 	#endif
+    
+    #if os(iOS)
```
SD10: Ditto

jason-ingenuity: Will update

Tests/UIKitTests/UICollectionViewExtensionsTests.swift
```diff
@@ -53,6 +53,13 @@ final class UICollectionViewExtensionsTests: XCTestCase {
 			XCTAssert(completionCalled)
 		}
 	}
+    
+    func testRegisterCellWithNibUsingClass() {
```
SD10: Now it errors. Maybe the resources aren't available to the test target? Can't look into it further

jason-ingenuity: My guess is that due to the removal of the `#if os(iOS)`, it executes the test on tvOS. Correct me if I'm wrong. Also, I've seen the same pattern to some tests in the UITableViewExtensionsTest that's why I implemented the same on my tests

jason-ingenuity: Will revert back to previous implementation since my tests consists of using an .xib file which is not supported for tvOS

SD10: If the method is available on `tvOS`, as declared in the source file, we shouldn't be restricting the test to `iOS` only. I'm wondering if the nib resources are not part of the `tvOS` target, causing the test to fail

Tests/UIKitTests/UITableViewExtensionsTests.swift
jason-ingenuity: Reverted due to issues with test cases being run on tvOS. Based pattern on already existing test cased on UITableViewExtensionTests

Tests/UIKitTests/UICollectionViewExtensionsTests.swift
jason-ingenuity: I see. I managed to make the tests succeed now. Will update PR

SD10: So it looks like it still fails. Maybe we should only make the source available to iOS? We should not expose them to tvOS if we can't get it to work.

jason-ingenuity: I believe so. So I'll just update the UITableView and UICollectionView extensions that make use of nibs

SD10:

PR #388, opened 2018-02-02
UserDefaults extension to save codable object

jason-ingenuity:
## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,7 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 - **URL**
   - added `queryParmeters` property to get the query parameters from a URL as a dictionary. [#370](https://github.com/SwifterSwift/SwifterSwift/pull/370) by [nathanbacon](https://github.com/nathanbacon).
 - **UserDefaults**
-  - added `codable(type: with key:)` method to be able to retrieve Codable objects from UserDefaults. [#388](https://github.com/SwifterSwift/SwifterSwift/pull/388) by [jason-ingenuity](https://github.com/jason-ingenuity).
+  - added `object(type: with key:)` method to be able to retrieve Codable objects from UserDefaults. [#388](https://github.com/SwifterSwift/SwifterSwift/pull/388) by [jason-ingenuity](https://github.com/jason-ingenuity).
   - added `set(codable: forKey key:)` method to be able to store Codable objects from UserDefaults. [#388](https://github.com/SwifterSwift/SwifterSwift/pull/388) by [jason-ingenuity](https://github.com/jason-ingenuity).
 > ### Changed
 > ### Deprecated
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,8 +26,8 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 - **URL**
   - added `queryParmeters` property to get the query parameters from a URL as a dictionary. [#370](https://github.com/SwifterSwift/SwifterSwift/pull/370) by [nathanbacon](https://github.com/nathanbacon).
 - **UserDefaults**
-  - added `object(type: with key:)` method to be able to retrieve Codable objects from UserDefaults. [#388](https://github.com/SwifterSwift/SwifterSwift/pull/388) by [jason-ingenuity](https://github.com/jason-ingenuity).
-  - added `set(codable: forKey key:)` method to be able to store Codable objects from UserDefaults. [#388](https://github.com/SwifterSwift/SwifterSwift/pull/388) by [jason-ingenuity](https://github.com/jason-ingenuity).
+  - added `object(type: with key: usingDecoder decoder:)` method to be able to retrieve Codable objects from UserDefaults. [#388](https://github.com/SwifterSwift/SwifterSwift/pull/388) by [jason-ingenuity](https://github.com/jason-ingenuity).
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -16,6 +16,9 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
   - added `withRoundedCorners(radius:)` to round the corners of an image. The parameter radius is optional, the resulting image will be round if it is unspecified. [#380](https://github.com/SwifterSwift/SwifterSwift/pull/380) by [@vyax](https://github.com/vyax).
 - **URL**
   - added `queryParmeters` property to get the query parameters from a URL as a dictionary. [#370](https://github.com/SwifterSwift/SwifterSwift/pull/370) by [nathanbacon](https://github.com/nathanbacon).
+- **UserDefaults**
+  - added `codable(type: with key:)` method to be able to retrieve Codable objects from UserDefaults. [#388](https://github.com/SwifterSwift/SwifterSwift/pull/388) by [jason-ingenuity](https://github.com/jason-ingenuity).
```
SD10: We have to update the method name here

Sources/Extensions/Foundation/UserDefaultsExtensions.swift
```diff
@@ -38,4 +38,29 @@ public extension UserDefaults {
 		return object(forKey: key) as? Date
 	}
 	
+    /// SwifterSwift: Retrieves a Codable object from UserDefaults.
+    ///
+    /// - Parameters:
+    ///   - type: Class that conforms to the Codable protocol.
+    ///   - key: key to find the Codable object for.
+    /// - Returns: Codable object for key (if exists).
+    public func object<T: Codable>(_ type: T.Type, with key: String) -> Codable? {
```
SD10: Can the return type be `T?`

Sources/Extensions/Foundation/UserDefaultsExtensions.swift
```diff
@@ -38,4 +38,33 @@ public extension UserDefaults {
...
 	}
 	
+    /// SwifterSwift: Retrieves a Codable object from UserDefaults.
+    ///
+    /// - Parameters:
+    ///   - type: Class that conforms to the Codable protocol.
+    ///   - key: key to find the Codable object for.
+    /// - Returns: Codable object for key (if exists).
+    public func object<T: Codable>(_ type: T.Type, with key: String) -> T? {
+        if let data = self.value(forKey: key) as? Data {
```
SD10: We can just use the `try?` keyword to ignore the thrown error and return `nil` 😛 
```Swift
let data = self.value(forKey: key) as? Data else { return nil }
return try? JSONDecoder().decode(type.self, from: data)
```

jason-ingenuity: Yes yes! I agree. Will update 😄

Sources/Extensions/Foundation/UserDefaultsExtensions.swift
```diff
@@ -38,4 +38,27 @@ public extension UserDefaults {
 		return object(forKey: key) as? Date
 	}
 	
+    /// SwifterSwift: Retrieves a Codable object from UserDefaults.
+    ///
+    /// - Parameters:
+    ///   - type: Class that conforms to the Codable protocol.
+    ///   - key: key to find the Codable object for.
+    /// - Returns: Codable object for key (if exists).
+    public func object<T: Codable>(_ type: T.Type, with key: String) -> T? {
```
omaralbeik: Lets add a decoder argument to this function incase user wanted to use a custom decoder:
```swift
public func object<T: Codable>(_ type: T.Type, with key: String, usingDecoder decoder: JSONDecoder = JSONDecoder()) -> T? {
    guard let data = self.value(forKey: key) as? Data else { return nil }
    return try? decoder.decode(type.self, from: data)
}
```

Same here let's add an encoder argument, incase user wanted to use a custom encoder:

```swift
public func set<T: Codable>(object: T, forKey key: String, usingEncoder encoder: JSONEncoder = JSONEncoder()) {
    let data = try? encoder.encode(object)
    self.set(data, forKey: key)
}
```

jason-ingenuity: I see. Will update, thanks!

Tests/FoundationTests/UserDefaultsExtensionsTests.swift
```diff
@@ -37,5 +41,13 @@ final class UserDefaultsExtensionsTests: XCTestCase {
 		XCTAssertNotNil(UserDefaults.standard.date(forKey: key))
 		XCTAssertEqual(UserDefaults.standard.date(forKey: key)!, date)
 	}
+    
+    func testCodable() {
```
SD10: We have to split this into two tests: one for set, one for get

jason-ingenuity: Noted, will update

jason-ingenuity: Question, what would be my assertion for the set test since verifying if the object was stored correctly is thru the get test?

SD10: @jason-ingenuity I really think the tests are the same. It probably just needs to be duplicated for the sake of consistency

jason-ingenuity: I see. Will do

PR #393, opened 2018-02-05
Refactors Int digits method for more performance

FrankKair:
This way of getting a number's digits is more performant and memory efficient, since we're not creating a new string and performing type castings.

Current version:

```swift
let intsArray = description.flatMap({Int(String($0))})		
return intsArray.map({Self($0)})
```

## Checklist

### DIFF
--- a/Sources/Extensions/SwiftStdlib/IntExtensions.swift
+++ b/Sources/Extensions/SwiftStdlib/IntExtensions.swift
@@ -80,16 +80,16 @@ public extension Int {
 
     /// SwifterSwift: Number of digits of integer value.
     public var digitsCount: Int {
-        var digits = [Int]()
+        guard self != 0 else { return 1 }
+        var digits = 0
         var number = self
 
         while number != 0 {
-            let x = (number % 10).abs
-            digits.append(x)
             number = number / 10
+            digits += 1
         }
 
-        return digits.count
+        return digits
     }
 
 }
--- a/Sources/Extensions/SwiftStdlib/IntExtensions.swift
+++ b/Sources/Extensions/SwiftStdlib/IntExtensions.swift
@@ -61,7 +61,37 @@ public extension Int {
 		}
 		return String(format: "\(sign)%ikk", abs / 100000)
 	}
-	
+
+    /// SwifterSwift: Array of digits of integer value.
+    public var digits: [Int] {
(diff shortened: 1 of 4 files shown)


Sources/Extensions/SwiftStdlib/IntExtensions.swift
```diff
@@ -61,7 +61,37 @@ public extension Int {
...
+            number = number / 10
+        }
+
+        digits.reverse()
+        return digits
+    }
+
+    /// SwifterSwift: Number of digits of integer value.
+    public var digitsCount: Int {
+        var digits = [Int]()
```
akuzminskyi: I suppose we can replace the array to a simple Int counter

FrankKair: Good one! Just changed it 👍

Sources/Extensions/SwiftStdlib/IntExtensions.swift
```diff
@@ -61,7 +61,37 @@ public extension Int {
...
-	
+
+    /// SwifterSwift: Array of digits of integer value.
+    public var digits: [Int] {
+        guard self != 0 else { return [0] }
+        var digits = [Int]()
+        var number = self
+
+        while number != 0 {
+            let x = (number % 10).abs
```
LucianoPAlmeida: Instead of calculate the abs on each iteration, maybe is better to apply the abs on `var number = self.abs`. That way you just have to calculate it once :)

akuzminskyi:
FYI: https://brilliant.org/wiki/finding-digits-of-a-number/ the fasters way to find a number of digits it's used log10(value)+1.

Sources/Extensions/SwiftStdlib/IntExtensions.swift
FrankKair: Good one! Done 😊👍🏻

FrankKair:
@akuzminskyi That's awesome! I just committed this change.

FrankKair:
@LucianoPAlmeida That's weird, I don't seem to be able to properly lint this file 🤔
Is there any chance you can commit the correct file?

FrankKair:
@LucianoPAlmeida Done!

LucianoPAlmeida:
Hey @FrankKair 
Thank you :)

PR #397, opened 2018-02-09
String to slug

FrankKair:
Creates a method to return a slug version of a string.
Example:

```swift
let str = "  A nice & hög _ Str    "
str.toSlug() // Returns "a-nice-&-hog-str"
```

## Checklist

### DIFF
--- a/Sources/Extensions/SwiftStdlib/StringExtensions.swift
+++ b/Sources/Extensions/SwiftStdlib/StringExtensions.swift
@@ -425,6 +425,34 @@ public extension String {
 		// https://stackoverflow.com/questions/42822838
 		return words().count
 	}
+
+    /// SwifterSwift: Transforms the string into a slug string.
+    ///
+    ///        "Swift is amazing".toSlug() -> "swift-is-amazing"
+    ///
+    /// - Returns: The string in slug format.
+    public func toSlug() -> String {
+        let lowercased = self.lowercased()
+        let latinized = lowercased.latinized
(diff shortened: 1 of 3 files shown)


FrankKair:
Do you guys think I'm missing some edge case here?
Or... is there a more efficient way of doing this?

Sources/Extensions/SwiftStdlib/StringExtensions.swift
```diff
@@ -425,6 +425,34 @@ public extension String {
 		// https://stackoverflow.com/questions/42822838
 		return words().count
 	}
+
+    /// SwifterSwift: Transforms the string into a slug string.
+    ///
+    ///        "Swift is amazing".toSlug() -> "swift-is-amazing"
+    ///
+    /// - Returns: The string in slug format.
+    public func toSlug() -> String {
```
SD10: I've never heard of a "slug" before 😂 But if this is something people need then it looks 👌

wdcurry: Slugs are used (in my experience) as safe endpoints to urls for blog posts etc.. Wordpress is all about the slug.

SD10: Ahhh, I see. Thanks for the info @wdcurry 👍

WestFlow127:
func toSlug() -> String does not remove all emoji's for example "str.toSlug() : test-🙋🏻‍♀️" ... I am using this function to create safe endpoints for our app. Would be nice to see an update to this issue in the next release. Thanks

Issue #360, opened 2018-01-10
Add an extension to get thumbnail image from a video URL


omaralbeik:
Add an extension to `UIImage` to generate a thumbnail image from a video URL using `AVAssetImageGenerator` from `AVFoundation`

BennX:
@omaralbeik I'd like to propsose this as a URL extension since UIImage generating a UIImage seems weird. It would look like `let thumbnail = URL(string: "path/to/resource").thumbnail()`. Taking a optional seconds in parameter and is returning an `UIImage?`. Furthermore I'd like to propse `URL(string: "path/to/resource").thumbnail() { UIImage? in  }` since requesting a the resource may take some time.

I got a PR ready just thinking about the testing here. Any ideas since I'm not that confident if it makes sense to test for nil / not nil?

PR #406, opened 2018-02-17
Add thumbnail method to URLExtensions

BennX:
This pull request should solve #360. 
It provides a simple extension to URL to generate a thumbnail from a given URL. It returns an optional `UIImage?` of an given url. If the AVFoundation cant generate the thumbnail it'll simply return nil.

In addition the documentation informs the user that it is recommended to dispatch the function call in case it should generate a thumbnail from a remote source since it may take some time.

The test may need a different "test" url where we are sure that it wont change in the future.

## Checklist

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -57,18 +57,28 @@ public extension SomeType {
 }
  ```
 
+- A pull request should only add one extension at a time.
 - All extensions should follow [Swift API Design Guidelines](https://developer.apple.com/videos/play/wwdc2016/403/)
 - Always declare extensions as **public**.
 - All extensions names should be as clear as possible.
 - All extensions should be well documented. see [Adding documentation](#adding-documentation)
 - Avoid using custom classes and objects the goal for this library is to extend the standards types available natively in Swift, iOS, macOS, watchOS and tvOS.
-- extensions could be:
- 	- enums
-	- Properties & static properties
-	- Methods & static methods
+- Extensions could be:
+ 	- Enums
+	- Instance properties & type properties
+	- Instance methods & type methods
 	- Initializers
-- Files are named based on the type that the contained extensions extend (example: all String extensions are found in "**StringExtensions.swift**" file)
-- extensions are ordered inside files in the following order:
+- All extensions should be tested.
+- Files are named based on the type that the contained extensions extend. 
+   - (example: all String extensions are found in "**StringExtensions.swift**" file)
+- All extensions files and test files have a one to one relation.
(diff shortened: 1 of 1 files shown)


Sources/Extensions/Foundation/URLExtensions.swift
```diff
@@ -55,5 +56,34 @@ public extension URL {
...
+    ///        let image_at_0seconds = url.thumbnail()
+    ///        let image_at_5seconds = url.thumbnail(fromTime: 5)
+    ///
+    ///        DisptachQueue.main.async {
+    ///            someImageView.image = url.thumbnial()
+    ///        }
+    ///
+    /// - Parameter fromTime: Seconds into the video where the image should be generated.
+    public func thumbnail(fromTime time: Float64 = 0) -> UIImage? {
+        let asset: AVAsset = AVAsset(url: self)
```
SD10: Can we remove the explicit `AVAsset` typing and let it be inferred?

Sources/Extensions/Foundation/URLExtensions.swift
```diff
@@ -55,5 +56,34 @@ public extension URL {
...
+        let imageGenerator = AVAssetImageGenerator(asset: asset)
+        let time = CMTimeMakeWithSeconds(time, 1)
+        
+        var actualTime: CMTime = CMTimeMake(0, 0)
```
SD10: Same with `actualTime`

Sources/Extensions/Foundation/URLExtensions.swift
```diff
@@ -55,5 +56,34 @@ public extension URL {
...
+// MARK: - Methods
+public extension URL {
+        
+    /// SwifterSwift: Generate a thumbnail image from given url. Returns nil if no thumbnail could be created. This function may take some time to complete. It's recommended to dispatch the call if the thumbnail is not generated from a local resource.
+    ///
+    ///        var url = URL(string: "https://video.golem.de/files/1/1/20637/wrkw0718-sd.mp4")!
+    ///        let image_at_0seconds = url.thumbnail()
+    ///        let image_at_5seconds = url.thumbnail(fromTime: 5)
+    ///
+    ///        DisptachQueue.main.async {
```
SD10: Let's remove the example of assigning here. Should be self explanatory

Sources/Extensions/Foundation/URLExtensions.swift
BennX: Sure 👍 However that slipped in the code.

BennX:
I'm confused i didn't close this PR.

Issue #399, opened 2018-02-16
Create an extension to stack UIViews


omaralbeik:
Create a `stacked` method to an extension of arrays of UIViews to create a UIStackView and add its arrengedSubViews, allowing users to set stack view's properties like spacing, axis, and others optionally via method arguments with default values.

PR #407, opened 2018-02-17
Add stack function to array with element UIView

BennX:
Adding the function to create a UIStackView from an Array of UIViews. Taking the defaults and parameter docs from Apples Defaults.

Solves #399

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
   - added `centerTextAndImage(spacing:)` to Center align title text and image on UIButton. [#365](https://github.com/SwifterSwift/SwifterSwift/pull/365) by [@imjog](https://github.com/imjog).
 - **Array**
   - added `divided(by:)` to separate an array into 2 arrays based on a predicate. [#367](https://github.com/SwifterSwift/SwifterSwift/pull/367) by [@neoneye](https://github.com/neoneye).
+  - added `stacked(orientation:, spacing:, alignment:, distribution:)` to create a stack of UIViews with parameters. [#407](https://github.com/SwifterSwift/SwifterSwift/pull/407) by [BennX](https://github.com/BennX).
 - **Int**
   - added `roundToNearest(n:)` to round an integer to the closest multiple of a given n. [#381](https://github.com/SwifterSwift/SwifterSwift/pull/381) by [@vyax](https://github.com/vyax).
 - **FileManager**
--- a/Sources/Extensions/SwiftStdlib/ArrayExtensions.swift
+++ b/Sources/Extensions/SwiftStdlib/ArrayExtensions.swift
@@ -24,6 +24,32 @@ public extension Array where Element: Numeric {
 	
 }
 
+#if os(iOS) || os(tvOS)
+// MARK: - Methods (UIView)
+public extension Array where Element: UIView {
+    
+    /// SwifterSwift: Generates an UIStackView of an array of UIView.
+    ///
+    ///        let views = [UIView(), UIView()]
+    ///        let stackView = views.stacked(orientation: .vertical)
+    ///
+    /// - Parameter orientation: The axis along which the arranged views are laid out.(default: .horizontal)
+    /// - Parameter spacing: The distance in points between the adjacent edges of the stack view’s arranged views.(default: 0.0)
+    /// - Parameter alignment: The alignment of the arranged subviews perpendicular to the stack view’s axis. (default: .fill)
+    /// - Parameter distribution: The distribution of the arranged views along the stack view’s axis.(default: .fill)
+    /// - Returns: A UIStackView containing all views of the array with the passed configuration
+    @available(iOS 9.0, *)
+    public func stacked(orientation: UILayoutConstraintAxis = .horizontal, spacing: CGFloat = 0.0, alignment: UIStackViewAlignment = .fill, distribution: UIStackViewDistribution = .fill) -> UIStackView {
+        let stack = UIStackView(arrangedSubviews: self)
+        stack.axis = orientation
+        stack.spacing = spacing
+        stack.alignment = alignment
+        stack.distribution = distribution
+        return stack
+    }
+}
+#endif
+
 // MARK: - Methods (FloatingPoint)
 public extension Array where Element: FloatingPoint {
 	
--- a/Tests/SwiftStdlibTests/ArrayExtensionsTests.swift
+++ b/Tests/SwiftStdlibTests/ArrayExtensionsTests.swift
@@ -460,4 +460,33 @@ final class ArrayExtensionsTests: XCTestCase {
                                                         Person(name: "Wade", age: nil)])
     }
     
+#if os(iOS) || os(tvOS)
+    func testStacked() {
+        let view1 = UIView()
+        let view2 = UIView()
+        let array = [view1, view2]
+        
+        XCTAssertEqual(array.stacked().arrangedSubviews.count, 2)
+        XCTAssertTrue(array.stacked().arrangedSubviews[0] === view1)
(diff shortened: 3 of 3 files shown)


LucianoPAlmeida:
Hey @BennX Thank you for taking your time to make this PR
The implementation seems nice, btw great tests 👍 
But some points I think is worth to discuss: 
* Although this is an Array extension, to me it feels like the ArrayExtensions is not the right place for this because it is the stdlib and this kinda inserts the UIKit dependency on something that is from stdlib.
* Another option for this could be convenience init on UIStackview that contains these parameters like:
```swift 
convenience init(arrangedSubviews: [UIView], orientation: UILayoutConstraintAxis = .horizontal, spacing: CGFloat = 0.0, alignment: UIStackViewAlignment = .fill, distribution: UIStackViewDistribution = .fill)
```    
What do you guys think? @BennX @SD10 @omaralbeik

BennX:
@LucianoPAlmeida I actually like the idea of having it as extension to an Array.

Since it will only exists if the Array is typed with UIView it wont extend every array and therefore only adds UIKit dependencies for that kind of array. Swift generics are on compile time therefore we should be good.

Having a convenience init on UIStackView is a good option too but for me it wont feal that `Swifty` to use a regular constructor for stacking a list. It feels like having a `sort(array:)` instead of a `array.sort`.

LucianoPAlmeida:
For sure @BennX totally agree, just point the convenience as an option :)
My main concern here is the first point, because of coupling dependency at a module level. I mean there's no problem in using generic constraints, in fact it is great, but to me the problem is where (in what module) this extension is defined, since we use subspecs to define the Stdlid extensions, UIKit ... and other modules. So defining a extension on a file of stdlib module that uses a type from uikit, makes the Stdlib module have a dependency on uikit module. That's what I mean with "dependency", btw sorry for the bad explanation on the first comment 😊

SD10:
@BennX How often do you have an array of `UIView` laying around to initialize a `UIStackView`?
I usually just create them 1 by 1:
```Swift
let btn = UIButton()
let sgmt = UISegmentedControl()
let view = UIView()
let stackView = UIStackView(views: [btn, sgmt, view])
```
Having to put them in an `Array` first feels kind of weird 🤔

calebkleveter:
If you where to implement an initializer, you could probably write it like this:
```swift
extension UIStackView {
    public init(views: [UIView], orientation: UILayoutConstraintAxis = .horizontal, spacing: CGFloat = 0.0, alignment: UIStackViewAlignment = .fill, distribution: UIStackViewDistribution = .fill) {
        self = views.stacked(orientation:, orientation, spacing: spacing, alignment: alignment, distribution: distribution)
    }
}
```

I don't know if you can do `self =` in a `UIStackView` init though.

SD10:
@calebkleveter Yeah, I don't think you can reference `self` because nothing has been initialized yet. It may be possible when initializing an enum or inside a failable initializer 🤔 

The only problem with the extension you listed above is that it creates a dependency between two extensions. We don't allow SwifterSwift extensions to be used inside other SwifterSwift extensions because most people copy & paste the code they need. This means they would have to search through the library to find all the extensions they need 😓

BennX:
@SD10 @calebkleveter @LucianoPAlmeida all of the comments are right. 

People tend to write things like:
```swift
let btn = UIButton()
let sgmt = UISegmentedControl()
let view = UIView()
let stackView = UIStackView(views: [btn, sgmt, view])
```
And we could easily create a variadic convenience constructor. But whenever you have the scenario like in a TableView where you don't know the content of the stack you'll build up an array for it, fill it and convert it into a stack. 

We could also add a variadic constructor and fully skip the adding of `[]` of the users. But I like the `array.stack` too. 
For me the difference between: 
```swift
let stackView = [btn, sgmt, view].stacked()
```
and
```swift
let stackView = UIStackView(views: [btn, sgmt, view])
```
is not that big. I'm fine having just one or both. It's just syntactic sugger. 

A third approach could be a variadic one:

```swift
extension UIStackView {
    public convenience init(orientation: UILayoutConstraintAxis = .horizontal, spacing: CGFloat = 0.0, alignment: UIStackViewAlignment = .fill, distribution: UIStackViewDistribution = .fill, views: UIView ...) {
        self.init(arrangedSubviews: views)
        self.axis = orientation
        self.spacing = spacing
        self.alignment = alignment
        self.distribution = distribution
    }
}
```

We could use it like this:
```swift
let btn = UIButton()
let sgmt = UISegmentedControl()
let view = UIView()
let stackView = UIStackView(views: btn, sgmt, view)
```

However having stacked on a array of UIViews seem to be handy in case someone need's it. If there  is no conclusion in "just have that one extension", we could simply provide all of them.

LucianoPAlmeida:
@BennX For me, the convenience init is the best option for all the reasons that the guys pointed out above :)

omaralbeik:
Hey guys 👋 
`UIStackView` has an initializer to set arranged subviews already [`init(arrangedSubviews:)`](https://developer.apple.com/documentation/uikit/uistackview/1616240-init)

I don't use storyboards, instead, I add my UI elements in code and customize them in closures which gives me a better separation and control over each view properties, like the code below:

```swift
class View: UIView {

	lazy var button: UIButton = {
		let button = UIButton()
		// customize the button
		return button
	}()

	lazy var label1: UILabel = {
		let label1 = UILabel()
		// customize the label
		return label1
	}()

	lazy var label2: UILabel = {
		let label2 = UILabel()
		// customize the label
		return label2
	}()

	lazy var stackView: UIStackView = {
		return [button, label1, label2].stacked()
	}()

}

private extension Array where Element == UIView {

	func stacked(/* argments like axis and others */) -> UIStackView {
		let stackView = UIStackView(arrangedSubviews: self)
		// customize the stackView based on method given arguments
		return stackView
	}

}
```

The `convenience initializer` sounds more like _UIKit_ but when I follow the code example above from top to bottom I think it also makes sense to stack `UIViews` as if they are an array :)

If we decided on adding this as an extension to _Array_, it should NOT be added to the current ArrayExtensions.swift file since it is used in SwiftStd subspec. Maybe we can create another ArrayExtensions.swift file under the `UIKit` subspec with extensions for arrays that has UIKit elements only

Thank you!

SD10:
> but when I follow the code example above from top to bottom I think it also makes sense to stack UIViews as if they are an array :)

These are in an Array too 🤔
```Swift
lazy var stackView: UIStackView = {
	return UIStackView([button, label1, label2])
}()
```
i-was-told-there-would-be-bikeshed
I'm not really bothered by this too much to be honest, but here are my thoughts:

- Feels weird to have an extension on `Array` that is related to UIKit 😂 
- It requires a different project structure
- Increases compile times for your project to resolve the generic constraint

omaralbeik:
@SD10 You got me convinced on your third point, I don't want to make my app compile slower, that's the whole reason why I don't use Storyboards 😉

Anyway, the idea is to set the arranged subviews and properties of a stack view in one line, since `convenience init` is a better fit for SwifterSwift, I totally agree with you 👍

```swift
public extension UIStackView {

	public convenience init(
		arrangedSubviews: [UIView],
		axis: UILayoutConstraintAxis,
		distribution: UIStackViewDistribution = .fill,
		alignment: UIStackViewAlignment = .leading,
		spacing: CGFloat = 0) {

		self.init(arrangedSubviews: arrangedSubviews)

		// set properties here
	}
}
```

omaralbeik:
@BennX Sorry for the inconvenience, could you please refactor the extension into a `convenience init` and add it to a new **`UIStackViewExtensions.swift`** file under the **UIKit** directory :)

wdcurry:
Interesting thread. While i love the time-saving of IB, and the intuitive nature of it, and doubt i would engage the direction of this thread, it is nice to see the effort as an option.

BennX:
@omaralbeik As soon as i have time I'll create a new "clean" PR for that. 👍

PR #411, opened 2018-02-18
Add thumbnail method to URLExtensions

BennX:
This pull request should solve #360. 
It provides a simple extension to URL to generate a thumbnail from a given URL. It returns an optional `UIImage?` of an given url. If the AVFoundation cant generate the thumbnail it'll simply return nil.

In addition the documentation informs the user that it is recommended to dispatch the function call in case it should generate a thumbnail from a remote source since it may take some time.

The test may need a different "test" url where we are sure that it wont change in the future.

Bring that back again. Somehow the old PR was closed because I did reset my branch.

## Checklist


BennX:
I want to add the note: the tests take quiet a while because it does requests agains a url in the web. A local video resource would be faster and more stable since that url may changes. (I didn't have a small video resource which I could share)

omaralbeik:
Thank you @BennX, I agree, it's preferred to have an offline video to test this extension, we can add [this video](http://www.sample-videos.com/video/mp4/720/big_buck_bunny_720p_1mb.mp4) to our test bundle, it is free and just 1 MB. this will be useful in future as well when we add more AV extensions.

BennX:
I can't download that video. Is the link wrong/down?

omaralbeik:
@BennX maybe because the link is not https 🤔 
Here is the video anyway 
[big_buck_bunny_720p_1mb.mp4.zip](https://github.com/SwifterSwift/SwifterSwift/files/1736567/big_buck_bunny_720p_1mb.mp4.zip) :)

BennX:
@omaralbeik I'll see when i got time for adding it.

BennX:
@omaralbeik on its way

PR #416, opened 2018-03-08
MKPolyline.coordinates and MKPolyline(coordinates:)

freak4pc:
This PR adds the following two additions:

- A `.coordinate` property on an instance of `MKPolyline`, to return a list of coordinates from a provided `MKPolyline`.

- A `init(coordinates:)` initializer, to initialize a new `MKPolyline` from a provided list of coordinates.

### DIFF
--- a/Sources/Extensions/MapKit/MKPolylineExtensions.swift
+++ b/Sources/Extensions/MapKit/MKPolylineExtensions.swift
@@ -1,0 +1,27 @@
+//
+//  MKPolylineExtensions.swift
+//  SwifterSwift
+//
+//  Created by Shai Mishali on 3/8/18.
+//  Copyright © 2018 SwifterSwift
+//
+
+#if !os(watchOS)
+
+import MapKit
+
+@available(tvOS 9.2, *)
+public extension MKPolyline {
+
+    /// SwifterSwift: Return an Array of coordinates representing the provided polyline.
+    public var coordinates: [CLLocationCoordinate2D] {
+        var coords = [CLLocationCoordinate2D](repeating: kCLLocationCoordinate2DInvalid,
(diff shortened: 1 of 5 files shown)


Sources/Extensions/MapKit/MKPolylineExtensions.swift
```diff
@@ -0,0 +1,36 @@
+//
+//  MKPolylineExtensions.swift
+//  SwifterSwift
+//
+//  Created by Shai Mishali on 3/8/18.
+//  Copyright © 2018 SwifterSwift
+//
+
+import MapKit
```
SD10: Looks like we'll have to wrap these extensions in `#if os(iOS) || os(macOS)`. We support every platform and it seems to be unsupported on `watchOS` and `tvOS < 9.2`

SD10:
We should also add a new subspec to the `Podspec` since this is the first `MapKit` extension:
```Ruby
  # MapKit Extensions
  s.subspec 'MapKit' do |sp|
    sp.source_files  = 'Sources/Extensions/MapKit/*.swift'
  end
```

Sources/Extensions/MapKit/MKPolylineExtensions.swift
freak4pc: I'm not entirely sure it builds it in watchOS given the fact its not part of that target 🤔 
f296d0b7-85c9-4bb0-b7f6-8dbe4c9a37c4

freak4pc:
@SD10 Should be good to go now 👍 

It's strange to me I need to define a build time rule for watchOS given the fact it's not part of that target but oh well 🤷‍♂️

freak4pc:
💚

PR #419, opened 2018-03-18
Update UITableViewExtensions.swift

happiehappie:
Setting the tableFooterView to nil won't remove the footer view, you need to create an empty UIView() to actually remove it

## Checklist

- [ ] I have added a changelog entry describing my changes.

### DIFF
--- a/Sources/Extensions/UIKit/UITableViewExtensions.swift
+++ b/Sources/Extensions/UIKit/UITableViewExtensions.swift
@@ -65,7 +65,7 @@ public extension UITableView {
 	
 	/// SwifterSwift: Remove TableFooterView.
 	public func removeTableFooterView() {
-		tableFooterView = nil
+		tableFooterView = UIView()
 	}
 	
 	/// SwifterSwift: Remove TableHeaderView.
--- a/Tests/UIKitTests/UITableViewExtensionsTests.swift
+++ b/Tests/UIKitTests/UITableViewExtensionsTests.swift
@@ -53,10 +53,8 @@ final class UITableViewExtensionsTests: XCTestCase {
 	}
 	
 	func testRemoveTableFooterView() {
-		tableView.tableFooterView = UIView()
-		XCTAssertNotNil(tableView.tableFooterView)
 		tableView.removeTableFooterView()
-		XCTAssertNil(tableView.tableFooterView)
+		XCTAssertEqual((tableView.tableFooterView ?? UIView()).frame.height, 0)
 	}
 	
 	func testRemoveTableHeaderView() {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,9 +10,9 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 > ### Removed
 > ### Fixed
 - **UITableView**
-  - Fixed removeTableFooterView() to actually remove the view. [#419](https://github.com/SwifterSwift/SwifterSwift/pull/419) by [happiehappie](https://github.com/happiehappie).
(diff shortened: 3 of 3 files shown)


happiehappie:
I didn't add any unit test cause the unit test is actually done correctly.

LucianoPAlmeida:
Hey @happiehappie :)) Thank you for taking the time to submit this PR. But tests are failing the build, could you take a look at the tests and see what's happening :))

Tests/UIKitTests/UITableViewExtensionsTests.swift
```diff
@@ -53,10 +53,8 @@ final class UITableViewExtensionsTests: XCTestCase {
 	}
 	
 	func testRemoveTableFooterView() {
-		tableView.tableFooterView = UIView()
-		XCTAssertNotNil(tableView.tableFooterView)
 		tableView.removeTableFooterView()
-		XCTAssertNil(tableView.tableFooterView)
+		XCTAssertEqual((tableView.tableFooterView ?? UIView()).frame.height, 0)
```
SD10: I would modify this test because we don't want to default back to another view. Better to check for nil explicitly and then compare frames

Sources/Extensions/UIKit/UITableViewExtensions.swift
```diff
@@ -65,7 +65,7 @@ public extension UITableView {
 	
 	/// SwifterSwift: Remove TableFooterView.
 	public func removeTableFooterView() {
-		tableFooterView = nil
+		tableFooterView = UIView()
```
SD10: I'm curious why this doesn't work? 🤔 We had a passing unit test for it

happiehappie:
Fair enough. If it’s nil it should fail immediately than having a safe fallback view

Sources/Extensions/UIKit/UITableViewExtensions.swift
happiehappie: Unless I should create a separate extension for this. 
Mainly is that. Setting tablefooterview to nil you still see your separators etc. To see an actual removed tablefooterview you need to pass an empty UIView()

SD10: @happiehappie Thanks for explaining. I did not know that. So IMO this is working properly, we just need to decide if we want to change this behavior. cc @LucianoPAlmeida

happiehappie: Yea so depends on you guys, maybe you guys indeed prefer remove to be “just-back-to-default” instead of remove. I can write a different function if you guys prefer the former

LucianoPAlmeida: @happiehappie 
Sorry for the late response 😅
I agree with @SD10, this extension is doing what it should do by the name `removeTableFooterView`. Maybe to this separator thing is better to have another extension. But I'm not sure since it's a very simple one line of code trick like assign to a property.

happiehappie: I see, it's fine, can close/delete this then :3

SD10: The problem is this extension is also 1 line @LucianoPAlmeida. Kind of pointless 🤔

LucianoPAlmeida: @SD10 Yep, I feel the same but since this one is already there :) 
I think there's no problem

SD10:
@happiehappie I'm going to close this because we believe this is working as expected and that this extension is trivial to begin with. Thanks for taking the time to put this PR together 👍 Although it did not get merged, we all learned something new!

PR #428, opened 2018-03-31
[WIP] Xcode 9.3 Swift 4.1 suport.

LucianoPAlmeida:
## Checklist


.travis.yml
```diff
@@ -1,5 +1,5 @@
 language: objective-c
-osx_image: xcode9.1
+osx_image: xcode9.3beta
```
SD10: only beta?

LucianoPAlmeida: Yeap, I was having issues with Travis because I was trying to put xcode9.3, but since it still doesn't have this it was running to the default xcode8.3, breaking it all 😂Took some commits and tests to notice that this was the issue

Sources/Extensions/SwiftStdlib/StringExtensions.swift
```diff
@@ -362,7 +362,7 @@ public extension String {
 		let formatter = NumberFormatter()
 		formatter.locale = locale
 		formatter.allowsFloats = true
-		return formatter.number(from: self) as? Double
+		return formatter.number(from: self)?.doubleValue
```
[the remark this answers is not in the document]
LucianoPAlmeida: Sorry @SD10 I didn't understand the question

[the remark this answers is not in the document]
SD10: No worries, just wanted to make sure that `doubleValue` is from the stdlib and not Swift

https://developer.apple.com/documentation/foundation/nsstring/1412321-floatvalue

Sources/Extensions/SwiftStdlib/BoolExtensions.swift
```diff
@@ -61,5 +61,5 @@ public extension Bool {
 		self = !self
```
LucianoPAlmeida: Should we deprecate this one? Now I think is in the stdlib

SD10: Yes I think that's a good idea

omaralbeik: Should we deprecate both `toggle` and `toggled`?

SD10: Yes

VincentSit: The built-in `toggle()` in Swift 4.2 conflicts with this, neither can be called. 

I ran into this problem because I am currently using Swift 4.2, I temporarily removed this method locally to solve this problem.

Maybe we need to remove this method in the next release, or add a swift version check condition.

PR #432, opened 2018-04-06
get instance from stroyboard

inchan:
- get instance from stroyboard

Find your instance in the storyboard and create it, this is claas method

## Checklist

- [ x ] New extensions are written in Swift 4. 
- [ x ] New extensions support iOS 8.0+ / tvOS 9.0+ / macOS 10.10+ / watchOS 2.0+.
- [ x ] I have added tests for new extensions, and they passed.
- [ x ] All extensions are declared as **public**.
- [  ] I have added a changelog entry describing my changes.

### DIFF
--- a/Sources/Extensions/UIKit/UIViewControllerExtensions.swift
+++ b/Sources/Extensions/UIKit/UIViewControllerExtensions.swift
@@ -83,6 +83,9 @@ public extension UIViewController {
         
         /// instance from storyboard
         /// ViewControllerClassName.instanceFromStoryboard()
+        /// - Parameters:
+        ///   - storyboardName: Storyboard to which the ViewController belongs (if nil -> search storyboard)
+        /// - Returns: self instance (optional value)
         ///
         public class func instanceFromStoryboard(_ storyboardName: String? = nil) -> Self? {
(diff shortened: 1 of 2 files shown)


LucianoPAlmeida:
Hey @inchan Thank you for taking the time to make this PR and sorry for the late response, we will take a look into this as soon as possible 👍

omaralbeik:
I see, however as far as I understood from your code, you're also relying on identifiers to instantiate view controller instances 🤔 

Unfortunately, I don't think this extension is a good fit for SwifterSwift for the following reasons:
- No sufficient test code is provided, we're not sure how this extensions will work on varius use cases.
- Methods are not self-contained, they depend on a private function, we assume users should be able to copy a single extension and paste it into their project and it should work.
- SwiftLint was not used to ensure syntax consistency with other extensions.
- [Similar extension](https://github.com/SwifterSwift/SwifterSwift/blob/master/Sources/Extensions/UIKit/UIStoryboardExtensions.swift#L27) already exists in the current codebase.

inchan:
@omaralbeik  ok... i understand.. i will be back  thank you omaralbeik

PR #437, opened 2018-04-10
added a simple method that checks to see if CoreLocation is available on the device

Tawfikk:


Sources/Extensions/CoreLocation/CLLocationExtensions.swift
```diff
@@ -70,8 +70,7 @@ public extension CLLocation {
 		return (degrees+360).truncatingRemainder(dividingBy: 360)
 	}
 
-    /// SwifterSwift: Validates if the CoreLocation is available on the devices
-    
+    /// SwifterSwift: Validates if the CoreLocation is available on the devices
```
SD10: When we say "CoreLocation" it makes me think of the availability of the framework. Do you think it would be better to be more explicit in this comment?

Possibly:
`/// SwifterSwift: Validates if the device has location permissions enabled`

Let me know what you think.

Tawfikk:
😂 🤣

SD10:
@Tawfikk I'm going to close this for now. Please feel free to reopen with the requested changes 👍

PR #442, opened 2018-04-12
Optional extensions

omaralbeik:
🚀 

## Checklist

### DIFF
--- a/Sources/Extensions/SwiftStdlib/OptionalExtensions.swift
+++ b/Sources/Extensions/SwiftStdlib/OptionalExtensions.swift
@@ -9,22 +9,22 @@
 // MARK: - Methods
 public extension Optional {
 
-	/// SwifterSwift: Get self of default value (if self is nil).
-	///
-	///		let foo: String? = nil
-	///		print(foo.unwrapped(or: "bar")) -> "bar"
-	///
-	///		let bar: String? = "bar"
-	///		print(bar.unwrapped(or: "foo")) -> "bar"
-	///
-	/// - Parameter defaultValue: default value to return if self is nil.
-	/// - Returns: self if not nil or default value if nil.
-	public func unwrapped(or defaultValue: Wrapped) -> Wrapped {
-		// http://www.russbishop.net/improving-optionals
-		return self ?? defaultValue
-	}
+    /// SwifterSwift: Get self of default value (if self is nil).
+    ///
+    ///		let foo: String? = nil
+    ///		print(foo.unwrapped(or: "bar")) -> "bar"
+    ///
+    ///		let bar: String? = "bar"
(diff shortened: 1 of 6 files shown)


Tests/SwiftStdlibTests/StringExtensionsTests.swift
```diff
@@ -627,10 +633,12 @@ final class StringExtensionsTests: XCTestCase {
     #if os(iOS)
     func testItalic() {
         let italicString = "hello".italic
+        // swiftlint:disable legacy_constructor
```
LucianoPAlmeida: Instead of disable all and enable all, it wouldn't be better just use //swiftlint:disable:next legancy_constructor ?

Sources/Extensions/SwiftStdlib/OptionalExtensions.swift
```diff
@@ -40,39 +40,50 @@ public extension Optional {
...
+    ///   - lhs: Any?
+    ///   - rhs: Any?
+    public static func ??= (lhs: inout Optional, rhs: Optional) {
+        guard let rhs = rhs else { return }
+        lhs = rhs
+    }
+
+}
+
+// MARK: - Methods (RandomAccessCollection)
```
LucianoPAlmeida: RandomAccessCollection comment is right?

omaralbeik: No, my bad, fixing it now 😅

Tests/SwiftStdlibTests/StringExtensionsTests.swift
omaralbeik: The reason I'm doing this is due to the modular nature of extensions in SwifterSwift, so when users copy a single extension to their project where SwiftLint is used, it should just work

LucianoPAlmeida: Ahh I see, but what I mean was 
Instead of 
```swift
// swiftlint:disable legacy_constructor
let range = NSMakeRange((text as NSString).length - 1, 1)
scrollRangeToVisible(range)
// swiftlint:enable legacy_constructor
```
Just to use 
```swift
// swiftlint:disable:next legacy_constructor
let range = NSMakeRange((text as NSString).length - 1, 1)
scrollRangeToVisible(range)
```
To disable the rule only for the next time, so it doesn't have disable and  then enable, just one command :))  But it's ok this way 👍

omaralbeik: Hmmm, now I understand, definitely it makes more sense, I'll update it throughout the project

SD10: Didn't know this syntax exists 💯

PR #444, opened 2018-04-12
Random numerics

guykogus:
🚀

## Summary
These new protocols/extensions allow us to generate random numbers of any type. E.g.
`let diceRoll = Int.random(upperBound: 6)`
`let angle = Double.random(lowerBound: -.pi, upperBound: .pi)`

It turns out that this was much more complicated than initially anticipated. I thought that for integers we'd be able to create one function for `BinaryInteger`s, since Apple's documentation for `numericCast` looked promising:
```
func random(in range: Range<Int>) -> Int {
    return numericCast(arc4random_uniform(numericCast(range.count))) + range.lowerBound
}
```

However there are 2 main issues with their example:
1. It will crash for signed integers when `range.count` is larger number than `Int` can represent, e.g. `Int.min..<Int.max`.
2. `arc4random_uniform` returns a `UInt32`, so it will crash if we want a 64-bit value larger than `UInt32.max`.

After a lot of research and experimentation I found this [StackOverflow entry](https://stackoverflow.com/questions/24007129/how-does-one-generate-a-random-number-in-apples-swift-language/25129039#25129039), which helped me find a very nice solution.

I've also added some future-ready functionality for `Strideable` so that we can get a random `Index` in a `Collection`. But that will be my next task.

## Checklist

### DIFF
--- a/Sources/Extensions/Shared/RandomizableInteger.swift
+++ b/Sources/Extensions/Shared/RandomizableInteger.swift
@@ -1,0 +1,154 @@
+//
+//  RandomizableInteger.swift
+//  SwifterSwift
+//
+//  Created by Guy Kogus on 13/04/2018.
+//  Copyright © 2018 SwifterSwift
+//
+
+/// An integer from which we can generate random numbers.
+public protocol RandomizableInteger: FixedWidthInteger {
+    /// SwifterSwift: Generate a random integer within a given range.
+    ///
+    /// - Parameters:
+    ///   - lowerBound: The lower bound of the integer range.
+    ///   - upperBound: The upper bound of the integer range.
+    /// - Returns: An integer between `lowerBound` and `upperBound` (exclusive).
+    static func random(lowerBound: Self, upperBound: Self) -> Self
+}
+
+/// A signed integer that has an unsigned integer counterpart
+public protocol SignedRandomizableInteger: RandomizableInteger, SignedInteger {
+
+    /// The `UnsignedInteger` counterpart with equal bit width
+    associatedtype Unsigned: UnsignedRandomizableInteger
+
+}
+
+/// An unsigned integer that has a signed integer counterpart
+public protocol UnsignedRandomizableInteger: RandomizableInteger, UnsignedInteger {
+
+    /// The `SignedInteger` counterpart with equal bit width
+    associatedtype Signed: SignedRandomizableInteger
+
+}
+
+// MARK: - Random functions
+
+// The randomization functions are adapted from @jstn's solution:
+// https://stackoverflow.com/questions/24007129/how-does-one-generate-a-random-number-in-apples-swift-language/25129039#25129039
+
+public extension UnsignedRandomizableInteger {
+    /// SwifterSwift: Generate a random integer within a given range.
+    ///
+    ///   This version is able to handle `UInt64`,
+    ///   which would not be possible using `BinaryInteger.random(lowerBound:upperBound:)`
+    ///
+    /// - Parameters:
+    ///   - lowerBound: The lower bound of the integer range. Defaults to 0.
+    ///   - upperBound: The upper bound of the integer range.
+    /// - Returns: An integer between `lowerBound` and `upperBound` (exclusive).
+    /// - Complexity:
+    ///   This could theoretically loop forever but each retry has p > 0.5 (worst case, usually far better).
+    public static func random(lowerBound: Self = 0, upperBound: Self) -> Self {
+        // Static asserts are not yet available in Swift
+        assert(Self.bitWidth == Signed.bitWidth)
+        assert(lowerBound < upperBound)
+
+        var mod: Self
+        let diff = upperBound - lowerBound
+        var rand: Self = arc4random()
+
+        if diff > Self(Signed.max) {
+            mod = 1 + ~diff
+        } else {
+            mod = ((max - (diff * 2)) + 1) % diff
+        }
+
+        while rand < mod {
+            rand = arc4random()
+        }
+
+        return (rand % diff) + lowerBound
+    }
+}
+
+public extension SignedRandomizableInteger {
+    /// SwifterSwift: Generate a random integer within a given range.
+    ///
+    ///   This version is able to handle `Int64`, as well as finding a random number in the range `Self.min..<Self.max`,
+    ///   which would not be possible using `BinaryInteger.random(lowerBound:upperBound:)` due to overflow
+    ///
+    /// - Parameters:
+    ///   - lowerBound: The lower bound of the integer range. Defaults to 0.
+    ///   - upperBound: The upper bound of the integer range.
+    /// - Returns: An integer between `lowerBound` and `upperBound` (exclusive).
+    public static func random(lowerBound: Self = 0, upperBound: Self) -> Self {
+        // Static asserts are not yet available in Swift
+        assert(Self.bitWidth == Unsigned.bitWidth)
+        assert(lowerBound < upperBound)
+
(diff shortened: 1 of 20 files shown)


Sources/Extensions/Shared/RandomizableInteger.swift
```diff
@@ -0,0 +1,154 @@
...
+    /// SwifterSwift: Generate a random integer within a given range.
+    ///
+    ///   This version is able to handle `Int64`, as well as finding a random number in the range `Self.min..<Self.max`,
+    ///   which would not be possible using `BinaryInteger.random(lowerBound:upperBound:)` due to overflow
+    ///
+    /// - Parameters:
+    ///   - lowerBound: The lower bound of the integer range. Defaults to 0.
+    ///   - upperBound: The upper bound of the integer range.
+    /// - Returns: An integer between `lowerBound` and `upperBound` (exclusive).
+    public static func random(lowerBound: Self = 0, upperBound: Self) -> Self {
```
calebkleveter: Would it make sense to give the `upperBound` parameter a default value of `Self.max`? That static property is defined in by `FixedWidthInteger` which `RandomizableInteger ` and requires conformance to.

guykogus: For the general use case I doubt it would make sense. Often when you want a random number it would be in the context of something like:
`let diceRoll = Int.random(upperBound: 6)`
or
`let randomElement = myArray[Int.random(upperBound: myArray.count)]` (although my next step is to get better random elements from collections 😉)

That opens the question of whether or not the function signature should be:
`random(_ lowerBound: Self = 0, _ upperBound: Self) -> Self`
That's a matter of taste, so I'd appreciate anybody else's input?

SwifterSwift.xcodeproj/project.pbxproj
```diff
@@ -1125,6 +1169,7 @@
 			attributes = {
 				LastSwiftUpdateCheck = 0900;
 				LastUpgradeCheck = 0930;
+				ORGANIZATIONNAME = SwifterSwift;
```
guykogus: This will make sure that whenever a new file is created it won't show `//  Copyright © 2018 ___ORGANIZATIONNAME___`

guykogus:
Is someone here able to test this on Swift 4.0? I'm reading [this article](https://medium.com/@slavapestov/behind-the-scenes-improvements-in-swift-4-1-269dd56e30c2) and it looks like returning `Self` in a generic protocol extension wasn't allowed in Swift 4.0. It's working for me because I'm runing Swift 4.1.

SwifterSwift.podspec
```diff
@@ -27,17 +27,17 @@ Pod::Spec.new do |s|
...
   # SwiftStdlib Extensions
   s.subspec 'SwiftStdlib' do |sp|
-    sp.source_files  = 'Sources/Extensions/SwiftStdlib/*.swift'
+    sp.source_files  = 'Sources/Extensions/SwiftStdlib/*.swift', 'Sources/Extensions/Shared/RandomizableInteger.swift'
   end
 
   # Foundation Extensions
   s.subspec 'Foundation' do |sp|
-    sp.source_files  = 'Sources/Extensions/Foundation/*.swift'
+    sp.source_files  = 'Sources/Extensions/Foundation/*.swift', 'Sources/Extensions/SwiftStdlib/BinaryFloatingPointExtensions.swift'
```
guykogus: Is this alright, or should anything that's shared be in the `Shared` folder?

SD10:
I haven't reviewed this but about CI, we should already be using Xcode 9.3

SwifterSwift.podspec
SD10: I don't have any time to review this PR because it's pretty large. But we shouldn't be adding extensions on the Swift Standard Libary to things like Core Graphics or UIKit

omaralbeik:
Thank you @guykogus, a very interesting way to handle randoms 😎 , I'm a huge fan of swift's protocols!
My only concern is how this model will fit into the **modular structure** we have, we assume users should be able to copy a single extension and use it directly without any complications. I personally don't use SwifterSwift as a dependency in my projects, I prefer to copy only what I need.

guykogus:
It seems weird to me to have it modular in a way that there are _no_ connections between them. After all, `Foundation` depends on the `SwiftStdLib`, and `UIKit` depends on `Foundation` and `CoreGraphics`, so it would make sense that we follow a similar structure in our extensions.

It would be wrong if we had `CoreLocation` suddenly depending on `UIKit`, since they're completely unrelated, so I'd understand having those 2 being separated.

Otherwise there's no reason why improving a framework shouldn't be passed on to its dependent frameworks.

omaralbeik:
Thank you @guykogus, I added a comment clarifying this [here](https://github.com/SwifterSwift/SwifterSwift/pull/445#discussion_r181561892), please let me know if you have any question 🙂

guykogus:
Your comment doesn't really clarify. You've written:

> units might use other system wide units; example: I can use `UIColor` or `Int` inside an extension for `UILabel`, since we asssume these types are available togehter when I import `UIKit` into my project.

`Int` isn't part of `UIKit`, it's part of `SwiftStdLib`. So if you're allowed to use `Int` in an extension of `UILabel` you should be able to use an extension of `Int` as well.

The example of this PR, where we can now get random values from any standard Swift numeric, could allow us to extend `Array`s to get a random element, or `Collection`s in general to do the same, or `String`s to get a random `Character`, or to generate random `Character`s/`Date`s/`Color`s, etc. But you're saying that this shouldn't be allowed, instead we should copy-and-paste the code every time we want to do such a thing?! That's **awful** software design!

The wonderful thing about using Cocoapods is that we can define dependencies cleanly. So we can make the `Foundation` extension depend on the `SwiftStdLib` extension, since `Foundation` already depends on Swift. Then we can make the `UIKit` extension depend on the `Foundation` dependency, and so on. That way we can maintain framework dependencies in a safe manner and leverage our extensions to get the most out of them.

omaralbeik:
I see your point @guykogus 👍

Of course, `Int` is not a part of `UIKit`, but you get `SwiftStdLib` types for free when you import `UIKit` in a swift file, you can technically use a custom extension of `Int` in a `UILabel` extension, however, if a user installed `SwifterSwift/UIKit` pod, things will break, since the `Int` extensin is not contained within the `SwifterSwift/UIKit` subspeced pod.

Also keep in mind users might not use Cocoapods at all, and just copy that single `UILabel` extension

The problem here is you are viewing the project as a single unit, while we're focusing on keeping the modularity of all extensions!

omaralbeik:
Well, we don't see this as a **poor** or **aweful** design for this specific project!

Again we're doing this for mainly 2 reasons:
- we're doing this to give users the flexibility to use a single extension from the project by just copying and pasting it in literally a few seconds.
- Say I don't like SwifterSwift's `Date` extensions and I want to use something like [SwiftDate](https://github.com/malcommac/SwiftDate) or [Timepiece](https://github.com/naoty/Timepiece), but at the same time I want to use only UIKit extensions from SwifterSwift; if `SwifterSwift/UIKit` depends on `SwifterSwift/Foundation`, this will be impossible due to the conflicts between the two framworks.

guykogus:
So, for example, if a user includes dependency A which has an extension on `class Foo` with function `bar()`, but then they try to include dependency B which also has an extension on `class Foo` with function `bar()`, then the extensions will clash? Is that the dilemma?

omaralbeik:
screen shot 2018-04-15 at 15 44 28

guykogus:
Dang. That makes life very difficult then. OK, so I guess this'll need to be its own project altogether.

PR #445, opened 2018-04-12
Add safeScrollToRow(at:at:animated:) to safely scroll table view

setoelkahfi:
🚀 

This extension will safely scroll UITableView to given IndexPath.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,10 +4,11 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 > # Upcoming release
 >
 > ### Added
-- **UITableViewExtentions**:
-  - Added `isValidIndexPath(_ indexPath:)` method to check whether given IndexPath is valid within UITableView. [#441](https://github.com/SwifterSwift/SwifterSwift/pull/441) by [setoelkahfi](https://github.com/setoelkahfi).
+- **UITableViewExtensions**:
+  - Added `isValidIndexPath(_:)` method to check whether given IndexPath is valid within UITableView. [#441](https://github.com/SwifterSwift/SwifterSwift/pull/441) by [setoelkahfi](https://github.com/setoelkahfi).
+  - Added `safeScrollToRow(at:at:animated:)` method to check whether given IndexPath is valid within UITableView. [#445](https://github.com/SwifterSwift/SwifterSwift/pull/445) by [setoelkahfi](https://github.com/setoelkahfi).
 > ### Changed
-- **UITableViewExtentions**:
+- **UITableViewExtensions**:
   - `dequeueReusableCell(withClass:for)`, `dequeueReusableCell(withClass)` now return `UITableViewCell` object, `fatalError(...)` if not found. [#439](https://github.com/SwifterSwift/SwifterSwift/pull/439) by [jdisho](https://github.com/jdisho)
   - `dequeueReusableHeaderFooterView(withClass)`now returns `UITableViewHeaderFooterView` object, `fatalError(...)` if not found. [#439](https://github.com/SwifterSwift/SwifterSwift/pull/439) by [jdisho](https://github.com/jdisho)
 - **UICollectionView**:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,14 +4,14 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 # Upcoming release
 
 ### Added
-- **UITableViewExtensions**:
+- **UITableView**:
   - Added `isValidIndexPath(_:)` method to check whether given IndexPath is valid within UITableView. [#441](https://github.com/SwifterSwift/SwifterSwift/pull/441) by [setoelkahfi](https://github.com/setoelkahfi).
-  - Added `safeScrollToRow(at:at:animated:)` method to check whether given IndexPath is valid within UITableView. [#445](https://github.com/SwifterSwift/SwifterSwift/pull/445) by [setoelkahfi](https://github.com/setoelkahfi).
+  - Added `safeScrollToRow(at:at:animated:)` method to safely scroll UITableView to the given IndexPath. [#445](https://github.com/SwifterSwift/SwifterSwift/pull/445) by [setoelkahfi](https://github.com/setoelkahfi).
 - **Optional**:
   - Added `isNilOrEmpty` property to check whether an optional is nil or empty collection.
 
 ### Changed
-- **UITableViewExtensions**:
+- **UITableView**:
   - `dequeueReusableCell(withClass:for)`, `dequeueReusableCell(withClass)` now return `UITableViewCell` object, `fatalError(...)` if not found. [#439](https://github.com/SwifterSwift/SwifterSwift/pull/439) by [jdisho](https://github.com/jdisho)
   - `dequeueReusableHeaderFooterView(withClass)`now returns `UITableViewHeaderFooterView` object, `fatalError(...)` if not found. [#439](https://github.com/SwifterSwift/SwifterSwift/pull/439) by [jdisho](https://github.com/jdisho)
 - **UICollectionView**:
--- a/Sources/Extensions/UIKit/UITableViewExtensions.swift
+++ b/Sources/Extensions/UIKit/UITableViewExtensions.swift
@@ -181,6 +181,18 @@ public extension UITableView {
         return indexPath.section < self.numberOfSections && indexPath.row < self.numberOfRows(inSection: indexPath.section)
     }
 
+    /// SwifterSwift: Safely scroll to possibly invalid IndexPath
+    ///
+    /// - Parameters:
+    ///   - indexPath: Target IndexPath to scroll to
+    ///   - scrollPosition: Scroll position
+    ///   - animated: Whether to animate or not
+    public func safeScrollToRow(at indexPath: IndexPath, at scrollPosition: UITableViewScrollPosition, animated: Bool) {
+        if isValidIndexPath(indexPath) {
+            self.scrollToRow(at: indexPath, at: scrollPosition, animated: animated)
+        }
+    }
+
 }
 #endif
 
--- a/Sources/Extensions/UIKit/UITableViewExtensions.swift
+++ b/Sources/Extensions/UIKit/UITableViewExtensions.swift
@@ -188,9 +188,9 @@ public extension UITableView {
     ///   - scrollPosition: Scroll position
     ///   - animated: Whether to animate or not
     public func safeScrollToRow(at indexPath: IndexPath, at scrollPosition: UITableViewScrollPosition, animated: Bool) {
-        if indexPath.section < numberOfSections && indexPath.row < numberOfRows(inSection: indexPath.section) {
-            self.scrollToRow(at: indexPath, at: scrollPosition, animated: animated)
-        }
+        guard indexPath.section < numberOfSections else { return }
+        guard indexPath.row < numberOfRows(inSection: indexPath.section) else { return }
+        scrollToRow(at: indexPath, at: scrollPosition, animated: animated)
     }
 
 }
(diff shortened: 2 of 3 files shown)


Sources/Extensions/UIKit/UITableViewExtensions.swift
```diff
@@ -181,6 +181,18 @@ public extension UITableView {
...
     }
 
+    /// SwifterSwift: Safely scroll to possibly invalid IndexPath
+    ///
+    /// - Parameters:
+    ///   - indexPath: Target IndexPath to scroll to
+    ///   - scrollPosition: Scroll position
+    ///   - animated: Whether to animate or not
+    public func safeScrollToRow(at indexPath: IndexPath, at scrollPosition: UITableViewScrollPosition, animated: Bool) {
+        if isValidIndexPath(indexPath) {
```
SD10: Thanks for this submission @setoelkahfi 👍 Unfortunately, we don't allow dependencies between two SwifterSwift extensions

setoelkahfi: Oh, I'll change it

Sources/Extensions/UIKit/UITableViewExtensions.swift
```diff
@@ -181,6 +181,18 @@ public extension UITableView {
...
     }
 
+    /// SwifterSwift: Safely scroll to possibly invalid IndexPath
+    ///
+    /// - Parameters:
+    ///   - indexPath: Target IndexPath to scroll to
+    ///   - scrollPosition: Scroll position
+    ///   - animated: Whether to animate or not
+    public func safeScrollToRow(at indexPath: IndexPath, at scrollPosition: UITableViewScrollPosition, animated: Bool) {
+        if indexPath.section < numberOfSections && indexPath.row < numberOfRows(inSection: indexPath.section) {
```
guykogus: Why not use `if isValidIndexPath(indexPath)`?

Tests/UIKitTests/UITableViewExtensionsTests.swift
```diff
@@ -104,6 +104,22 @@ final class UITableViewExtensionsTests: XCTestCase {
         XCTAssertFalse(tableView.isValidIndexPath(invalidIndexPath))
     }
 
+    func testSafeScrollToIndexPath() {
+        let validIndexPathTop = IndexPath(row: 0, section: 0)
+        XCTAssertNoThrow(tableView.safeScrollToRow(at: validIndexPathTop, at: .top, animated: false))
```
guykogus: Your method doesn't `throw` anything, neither does `UITableView.scrollToRow(at:at:)`, so you wouldn't technically `catch` it with this `XCTAssertNoThrow`.

It will still work though because `XCTAssert...` will report certain exceptions and continue running, but this isn't a valid test. E.g. You couldn't reverse it by calling
```XCTAssertThrowsError(tableView.scrollToRow(at: invalidIndexPath, at: .top, animated: false))```

You can either just calls these methods without `XCTest...` and test their side-effects (like you're already doing) or you can tell change the function to `throw` an error if the `indexPath` is invalid.

File: Tests/UIKitTests/UITableViewExtensionsTests.swift [hunk 4]
```diff
@@ -104,6 +104,22 @@ final class UITableViewExtensionsTests: XCTestCase {
...
+        let validIndexPathBottom = IndexPath(row: 7, section: 1)
+        let bottomOffset = CGPoint(x: 0, y: tableView.contentSize.height - tableView.bounds.size.height)
+        tableView.safeScrollToRow(at: validIndexPathBottom, at: .bottom, animated: false)
+        #if os(tvOS)
+            XCTAssertEqual(bottomOffset.y, tableView.contentOffset.y, accuracy: 15.0)
+        #else
+            XCTAssertEqual(bottomOffset.y, tableView.contentOffset.y, accuracy: 2.0)
+        #endif
+        let invalidIndexPath = IndexPath(row: 10, section: 0)
+        XCTAssertNoThrow(tableView.safeScrollToRow(at: invalidIndexPath, at: .bottom, animated: false))
```
guykogus: Same comment, method doesn't throw.

Sources/Extensions/UIKit/UITableViewExtensions.swift
SD10: https://github.com/SwifterSwift/SwifterSwift/pull/445#discussion_r181182596

guykogus: I don't understand your comment. `isValidIndexPath` is 10 lines above...

SD10: Each extension should be able to be copied and pasted without a dependency on other extensions

guykogus: `isValidIndexPath` is part of the same extension. More so, it's part of the same file. It's literally the function above.

SD10: Sorry, sometimes I do a poor job of explaining things. We don't allow using a SwifterSwift extension inside another SwifterSwift extension

omaralbeik: No @SD10 I think you explained it very well 🙂, @guykogus here is a further explanation:

```swift
extension A {
   func b() { ... }
   func c() { ... }
}

extension B {
   func b() { ... }
   func c() { ... }
}
```

- units from class `A` (`b` and `c`) **CAN NOT** be used in any unit for extension `B` and viceversa.
- unit `b` in class `A` **CAN NOT** be used in unit `c` in the same class extension
- all units in any extension should be **SELF DEPENDENT**, that means they **CAN NOT** rely on any other SwifterSwift unit.
- units might use other **system wide** units; example: I can use `UIColor` or `Int` inside an extension for `UILabel`, since we asssume these types are available togehter when I import `UIKit` into my project.

omaralbeik: @guykogus, here is how we will add this extension in SwifterSwift

```swift
extension UITableView {

	public func isValidIndexPath(_ indexPath: IndexPath) -> Bool {
		return indexPath.section < self.numberOfSections && indexPath.row < self.numberOfRows(inSection: indexPath.section)
	}

	public func safeScrollToRow(at indexPath: IndexPath, at scrollPosition: UITableViewScrollPosition, animated: Bool) {
		guard indexPath.section < self.numberOfSections else { return }
		guard indexPath.row < numberOfRows(inSection: indexPath.section) else { return }
		scrollToRow(at: indexPath, at: scrollPosition, animated: animated)
	}

}
```

omaralbeik: @guykogus, absolutely, that is how we all do software, no one likes to write the same code in 2+ places, this is common sense!, I see your point 💯

However, we see this project as a unique case where each extension should be encapsulated, even if this lead to some duplicated code.

guykogus: In this situation it seems pretty unnecessary, but ok.

setoelkahfi: Hi guys, sorry kinda late to this interesting discussion.
I don't really mind if I have to rewrite it to meet the requirements of this project.

Tests/UIKitTests/UITableViewExtensionsTests.swift
setoelkahfi: Thanks for the input.
I'd rather go with the first choice, since I wan't this function to fail silently if `IndexPath` is invalid.

Sources/Extensions/UIKit/UITableViewExtensions.swift
omaralbeik: Thank you @setoelkahfi 🤗 
Could you please replace the `if` condition with 2 `guard` statements like above

Tests/UIKitTests/UITableViewExtensionsTests.swift
omaralbeik: Thank you @setoelkahfi, I agree with @guykogus here, statements do not throw, since we are scrolling only if the index path is valid, we can set the offset to some point and check if the scroll offset did or didn't change based on the index path, something like:

```swift
func testSafeScrollToIndexPath() {
    let validIndexPathTop = IndexPath(row: 0, section: 0)

    tableView.contentOffset = .init(x: 0, y: 100)
    XCTAssertNotEqual(tableView.contentOffset, .zero)

    tableView.scrollToRow(at: validIndexPathTop, at: .top, animated: false)
    XCTAssertEqual(tableView.contentOffset, .zero)

    let validIndexPathBottom = IndexPath(row: 7, section: 1)
    let bottomOffset = CGPoint(x: 0, y: tableView.contentSize.height - tableView.bounds.size.height)

    tableView.contentOffset = .init(x: 0, y: 200)
    XCTAssertNotEqual(tableView.contentOffset, bottomOffset)

    tableView.safeScrollToRow(at: validIndexPathBottom, at: .bottom, animated: false)
    #if os(tvOS)
        XCTAssertEqual(tableView.contentOffset.y, bottomOffset.y, accuracy: 15.0)
    #else
        XCTAssertEqual(tableView.contentOffset.y, bottomOffset.y, accuracy: 2.0)
    #endif

    let invalidIndexPath = IndexPath(row: 213, section: 21)
    tableView.contentOffset = .zero

    tableView.safeScrollToRow(at: invalidIndexPath, at: .bottom, animated: false)
    XCTAssertEqual(tableView.contentOffset, .zero)
}
```

Sources/Extensions/UIKit/UITableViewExtensions.swift
omaralbeik: @setoelkahfi added below, sorry for the delay, time zone differences 😅

omaralbeik:
@setoelkahfi since we're here, could you fix the conflicts in Changelog as well 🙂

Tests/UIKitTests/UITableViewExtensionsTests.swift
setoelkahfi: Updated, could you please check :) @omaralbeik

CHANGELOG.md
```diff
@@ -4,13 +4,14 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 # Upcoming release
 
 ### Added
-- **UITableViewExtentions**:
-  - Added `isValidIndexPath(_ indexPath:)` method to check whether given IndexPath is valid within UITableView. [#441](https://github.com/SwifterSwift/SwifterSwift/pull/441) by [setoelkahfi](https://github.com/setoelkahfi).
-**Optional**:
+- **UITableViewExtensions**:
```
omaralbeik: One final thing before we merge this, @setoelkahfi could you please update this with **UITableView**

CHANGELOG.md
```diff
@@ -4,13 +4,14 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
...
+  - Added `isValidIndexPath(_:)` method to check whether given IndexPath is valid within UITableView. [#441](https://github.com/SwifterSwift/SwifterSwift/pull/441) by [setoelkahfi](https://github.com/setoelkahfi).
+  - Added `safeScrollToRow(at:at:animated:)` method to check whether given IndexPath is valid within UITableView. [#445](https://github.com/SwifterSwift/SwifterSwift/pull/445) by [setoelkahfi](https://github.com/setoelkahfi).
+- **Optional**:
   - Added `isNilOrEmpty` property to check whether an optional is nil or empty collection.
 
 ### Changed
-- **UITableViewExtentions**:
+- **UITableViewExtensions**:
```
omaralbeik: same here 🙂

CHANGELOG.md
setoelkahfi: done @omaralbeik

omaralbeik:
Thank you @setoelkahfi 👍

PR #446, opened 2018-04-12
Use UIImage's scale when resizing image

MaxHaertwig:
This is a bugfix.

### DIFF
--- a/Sources/Extensions/UIKit/UIImageExtensions.swift
+++ b/Sources/Extensions/UIKit/UIImageExtensions.swift
@@ -68,12 +68,13 @@ public extension UIImage {
 	///
 	/// - Parameters:
 	///   - toHeight: new height.
+    ///   - opaque: flag indicating whether the bitmap is opaque.
 	///   - orientation: optional UIImage orientation (default is nil).
 	/// - Returns: optional scaled UIImage (if applicable).
-	public func scaled(toHeight: CGFloat, with orientation: UIImageOrientation? = nil) -> UIImage? {
+    public func scaled(toHeight: CGFloat, opaque: Bool = false, with orientation: UIImageOrientation? = nil) -> UIImage? {
 		let scale = toHeight / size.height
 		let newWidth = size.width * scale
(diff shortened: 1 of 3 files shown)


Vyax:
Done 👍

Sources/Extensions/UIKit/UIImageExtensions.swift
```diff
@@ -73,7 +73,7 @@ public extension UIImage {
 	public func scaled(toHeight: CGFloat, with orientation: UIImageOrientation? = nil) -> UIImage? {
 		let scale = toHeight / size.height
 		let newWidth = size.width * scale
-		UIGraphicsBeginImageContext(CGSize(width: newWidth, height: toHeight))
+        UIGraphicsBeginImageContextWithOptions(CGSize(width: newWidth, height: toHeight), false, scale)
```
guykogus: It would be great if we could check for setting the value of `opaque`. You've put it to `false` by default, but we can potentially improve it. [This SO question](https://stackoverflow.com/questions/38999305/issue-in-checking-uiimage-have-alpha-transparant-color-or-not) seems to suggest that we can get the answer. Do you mind having a look?

SD10: Could also just add a parameter to let the user control this flag

Vyax:
Done 👍

VincentSit:
Is this PR buggy? I can't get the correct scaled image when the `scale` is set to image's scale.

[SwifterSwiftImageDemo.zip](https://github.com/SwifterSwift/SwifterSwift/files/2193129/SwifterSwiftImageDemo.zip)
02e5b4b7-0b76-4b42-bd7f-30c8ee97b112

guykogus:
Looks like it. That `scale` value doesn't make sense. Please go ahead and create a new PR with your fix :)

VincentSit:
@guykogus Thanks, I have created a new PR for this, please review it when you are free and feel free to tell me where I need to change.

PR #453, opened 2018-04-18
IntExtensions factors() func added

n0an:
## Checklist

### DIFF
--- a/Sources/Extensions/SwiftStdlib/IntExtensions.swift
+++ b/Sources/Extensions/SwiftStdlib/IntExtensions.swift
@@ -174,9 +174,17 @@ public extension Int {
     
     /// SwifterSwift: Array of factors for given integer.
     /// - Returns: array of factors
-    public func factors() -> [Int] {
-        let factors = (1...self).filter { self % $0 == 0 }
-        return factors
+    func factors() -> [Int] {
+        guard self > 0 else { return [] }
+        
+        let rootSquared = Int(Double(self).squareRoot())
+        var factors = Set<Int>()
+        
+        for i in 1...(rootSquared + 1) where self % i == 0 {
+            factors.formUnion([i, self/i])
+        }
+        
+        return factors.sorted()
     }
 }
 
--- a/Sources/Extensions/SwiftStdlib/IntExtensions.swift
+++ b/Sources/Extensions/SwiftStdlib/IntExtensions.swift
@@ -172,6 +172,12 @@ public extension Int {
 	}
 	// swiftlint:enable identifier_name
 
+    public func factors() -> [Int] {
+        let factors = (1...self).filter { self % $0 == 0 }
+//        let stringFactors = factors.map { String($0) }
+//        let joinedFactors = stringFactors.joined(separator: ", ")
+        return factors
+    }
 }
 
 // MARK: - Initializers
--- a/Sources/Extensions/SwiftStdlib/IntExtensions.swift
+++ b/Sources/Extensions/SwiftStdlib/IntExtensions.swift
@@ -171,11 +171,11 @@ public extension Int {
 		return n == 0 ? self : Int(round(Double(self) / Double(n))) * n
 	}
 	// swiftlint:enable identifier_name
-
+    
+    /// SwifterSwift: Array of factors for given integer.
+    /// - Returns: array of factors
     public func factors() -> [Int] {
         let factors = (1...self).filter { self % $0 == 0 }
-//        let stringFactors = factors.map { String($0) }
-//        let joinedFactors = stringFactors.joined(separator: ", ")
         return factors
     }
 }
--- a/Tests/SwiftStdlibTests/IntExtensionsTests.swift
+++ b/Tests/SwiftStdlibTests/IntExtensionsTests.swift
@@ -119,6 +119,7 @@ final class IntExtensionsTests: XCTestCase {
     }
     
     func testFactors() {
+        XCTAssertEqual(0.factors(), [])
         XCTAssertEqual(83.factors(), [1, 83])
         XCTAssertEqual(98.factors(), [1, 2, 7, 14, 49, 98])
     }
--- a/Tests/SwiftStdlibTests/IntExtensionsTests.swift
+++ b/Tests/SwiftStdlibTests/IntExtensionsTests.swift
@@ -117,6 +117,11 @@ final class IntExtensionsTests: XCTestCase {
         XCTAssert(63.roundToNearest(25) == 75)
         XCTAssert(42.roundToNearest(0) == 42)
     }
+    
+    func testFactors() {
+        XCTAssertEqual(83.factors(), [1, 83])
+        XCTAssertEqual(98.factors(), [1, 2, 7, 14, 49, 98])
+    }
 
     func testOperators() {
         XCTAssertEqual(5 ** 2, 25)


Sources/Extensions/SwiftStdlib/IntExtensions.swift
```diff
@@ -172,6 +172,12 @@ public extension Int {
 	}
 	// swiftlint:enable identifier_name
 
+    public func factors() -> [Int] {
+        let factors = (1...self).filter { self % $0 == 0 }
+//        let stringFactors = factors.map { String($0) }
```
LucianoPAlmeida: Can you remove this comment lines here?

Sources/Extensions/SwiftStdlib/IntExtensions.swift
```diff
@@ -172,6 +172,12 @@ public extension Int {
 	}
 	// swiftlint:enable identifier_name
```
LucianoPAlmeida: Can you add documentation comments for that?

n0an: Added doc comment.

Sources/Extensions/SwiftStdlib/IntExtensions.swift
```diff
@@ -171,7 +171,13 @@ public extension Int {
 		return n == 0 ? self : Int(round(Double(self) / Double(n))) * n
 	}
 	// swiftlint:enable identifier_name
-
+    
+    /// SwifterSwift: Array of factors for given integer.
+    /// - Returns: array of factors
+    public func factors() -> [Int] {
+        let factors = (1...self).filter { self % $0 == 0 }
```
guykogus: There is actually a more efficient way to get this. By checking [this SO question](https://stackoverflow.com/questions/6800193/what-is-the-most-efficient-way-of-finding-all-the-factors-of-a-number-in-python) we can convert the answer to:

```
func factors() -> [Int] {
    guard self > 0 else { return [] }
    return [Int]((1...Int(ceil(sqrt(Double(self)))) + 1)
        .reduce(Set<Int>(), { (result, val) -> Set<Int> in
            guard self % val == 0 else { return result }
            return result.union([val, self / val])
        }))
        .sorted()
}
```

You don't have to sort it, but I would imagine that the user would expect it to be so.

guykogus: Hmm, if you want you can make it generic for all `BinaryIntegers` (`UInt`, `Int8`, etc.), but it's a bit ugly.

```
extension BinaryInteger {
    func factors() -> [Self] {
        guard self > 0 else { return [] }
        return [Self]((1...Int64(Double(Int64(self)).squareRoot()) + 1)
            .reduce(Set<Self>(), { (result, val) -> Set<Self> in
                let val: Self = numericCast(val)
                guard self % val == 0 else { return result }
                return result.union([val, self / val])
            }))
            .sorted()
    }
}
```

LucianoPAlmeida: Hey @n0an @guykogus 
@guykogus I like your proposal and in fact it is more efficient
But maybe we can have as an option, instead of using `reduce` we use `reduce(into:)` to avoid  copy and use in-place mutation :))
Or even use a for loop implementation like:
```swift
func factors() -> [Int] {
        guard
            self > 0 else { return [] }
        
        let rootSquared = Int(Double(self).squareRoot())
        var factors = Set<Int>()
        for i in 1...(rootSquared + 1) where self % i == 0 {
            factors.insert(i)
            factors.insert(self/i)
        }
        return factors.sorted()
}
```
To me it seems more simpler :)) 
Let me know what you guys think?

SD10: Stop doing that stupid thing with the `guard` statement @LucianoPAlmeida 🤦‍♂️ 😭 😭

LucianoPAlmeida: @SD10 I can't help myself, it is stronger than me 😂😂

guykogus: @LucianoPAlmeida I think I prefer your for-loop method, although I'd still use `factors.formUnion([val, self / val])` instead of `insert`ing twice.
Also, what's the "stupid thing with the `guard` statement"? 🤔

n0an: @guykogus , @LucianoPAlmeida , @SD10 , I like suggested solution. As for `guard` I suppose this is about new line and indenting :)

LucianoPAlmeida: Yeap is about the new line indenting  😂😂
@guykogus About the `fromUnion` I agree :)) I just did with the `insert` because was just two calls and I wanted to avoid the array allocation to call the method, but I think is not a problem.

SD10: Anyone want to take a crack at updating this PR? 😬

Sources/Extensions/SwiftStdlib/IntExtensions.swift
```diff
@@ -171,7 +171,21 @@ public extension Int {
...
+    
+    /// SwifterSwift: Array of factors for given integer.
+    /// - Returns: array of factors
+    func factors() -> [Int] {
+        guard self > 0 else { return [] }
+        
+        let rootSquared = Int(Double(self).squareRoot())
+        var factors = Set<Int>()
+        
+        for i in 1...(rootSquared + 1) where self % i == 0 {
```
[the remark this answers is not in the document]
SD10: I wouldn't use an enum because it's not friendly in terms of SwifterSwift's "extensions should be easy to copy and paste" philosophy, however, I completely agree that we need to support negative ranges as well. Good catch 👍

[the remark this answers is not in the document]
n0an: @freak4pc , do you really think we need negative factors? As for me, it looks like unnecessary overcomplicated. Negative factors equal to positive factors negated. I think providing negative factors leads to less readability. I mean It's implied if we have positive factors, we have the same negative factors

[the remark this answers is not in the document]
guykogus: I agree with @n0an, there's no need for the negative factors.
This function could be adapted to handle a negative input, I suppose. Thoughts?

Sources/Extensions/SwiftStdlib/IntExtensions.swift
```diff
@@ -171,6 +171,22 @@ public extension Int {
...
+            factors.formUnion([i, self/i])
```
guykogus: Add a space: `self / i`

Sources/Extensions/SwiftStdlib/IntExtensions.swift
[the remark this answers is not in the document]
freak4pc: In that case I’d at least return a UInt. It’s all a question of what you deem as “do we need”? Wha is the use case for this extension? If it is to retrieve _all_ factors then it doesn’t do that correctly ATM 

If there is a “central” use case where these aren’t needed than I’d understand but this is a purely mathematical addition so I’m not sure it’s entirely correct that way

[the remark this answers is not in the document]
guykogus: Returning an unsigned int to show that we're using non-negative numbers [isn't recommended](https://google.github.io/styleguide/cppguide.html#Integer_Types). Also, this is an extension on `Int`, so returning an array of `UInt` may be confusing to a user of this function.

If you look up any example of finding factors of a number it will always show you the positive results only (e.g. WolframAlpha). As mentioned, the negative results are redundant. I wouldn't consider it to be an incomplete answer if we only have the positive results. Maybe the documentation for the class could be updated to mention this fact.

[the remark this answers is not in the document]
SD10: > Also, this is an extension on Int, so returning an array of UInt may be confusing to a user of this function.

Yeah, I wouldn't return a `UInt` here. We can return only positive factors but we should support negative input.

SD10:
Looks like this has gone stale so I'm going to close it for now 👍

Issue #459, opened 2018-04-26
error: package has unsupported layout


vkill:
Hi
I can't build on Xcode 9.3 Swift 4.1 .

```
$ cat Package.swift
import PackageDescription

let package = Package(
    name: "test",
    dependencies: [
        .package(url: "https://github.com/SwifterSwift/SwifterSwift", from: "4.3.0")
    ],
    targets: []
)

$ swift build
Fetching https://github.com/SwifterSwift/SwifterSwift
Cloning https://github.com/SwifterSwift/SwifterSwift
Resolving https://github.com/SwifterSwift/SwifterSwift at 4.3.0
'SwifterSwift' /path/.build/checkouts/SwifterSwift--6808214200715182586: error: package has unsupported layout; found loose source files: /path/.build/checkouts/SwifterSwift--6808214200715182586/Tests/SwifterSwiftTests.swift
```

```
$ cat Package.swift
import PackageDescription

let package = Package(
    name: "test",
    dependencies: [
        .package(url: "https://github.com/SwifterSwift/SwifterSwift", Package.Dependency.Requirement.branch("master"))
    ],
    targets: []
)

$ swift build  
Fetching https://github.com/SwifterSwift/SwifterSwift
Cloning https://github.com/SwifterSwift/SwifterSwift
Resolving https://github.com/SwifterSwift/SwifterSwift at master
'SwifterSwift' /path/.build/checkouts/SwifterSwift--6808214200715182586: error: package has unsupported layout; found loose source files: /path/.build/checkouts/SwifterSwift--6808214200715182586/Tests/SwifterSwiftTests.swift
```

SD10:
@vkill Can you try using
```swift
    dependencies: [
        .package(url: "https://github.com/SwifterSwift/SwifterSwift.git", .upToNextMajor(from: "4.3.0"))
    ],
```

vkill:
@SD10 not working

```
$ cat Package.swift
import PackageDescription

let package = Package(
    name: "test",
    products: [
        .library(
            name: "test",
            targets: ["test"]),
    ],
    dependencies: [
        .package(url: "https://github.com/SwifterSwift/SwifterSwift.git", .upToNextMajor(from: "4.3.0"))
    ],
    targets: [
        .target(
            name: "test",
            dependencies: []),
        .testTarget(
            name: "testTests",
            dependencies: ["test"]),
    ]
)

$  swift build       
[...]

PR #460, opened 2018-04-26
🚀 add `applyShadow(color:alpha:xOffset:yOffset:blur:spread)` method to ...

mmdock:
…apply a shadow to a CALayer with the same describing parameters as Sketch, Zeplin, etc. (#459)

## Checklist

- [X] New extensions are written in Swift 4.
- [X] New extensions support iOS 8.0+ / tvOS 9.0+ / macOS 10.10+ / watchOS 2.0+.

1)
I am not sure if shadows are supported in tvOS and watchOS - if not, proper checks need to be added
2)
I am not sure how to write a test for a UI based extension (If there is an example to reference, please link so I can learn)
3)
Please let me know thoughts on this.  I feel the parameter names and such may need a little adjusting, specifically xOffset and yOffset.  I wanted to just call them x and y like Sketch does, however, SwiftLint does not allow variable names of a length less than 3.

I happen to use this extension literally ALL the time.  (Well, one similar to it, but have found this one way easier to use, understand, and read... but the base functionality is the same.)

### DIFF
--- a/Sources/Extensions/CoreAnimation/CALayerExtensions.swift
+++ b/Sources/Extensions/CoreAnimation/CALayerExtensions.swift
@@ -1,0 +1,34 @@
+//
+//  CALayerExtension.swift
+//  SwifterSwift
+//
+//  Created by Morgan Dock on 4/26/18.
+//  Copyright © 2018 SwifterSwift
+//
+
+public extension CALayer {
+
+    /// SwifterSwift: Create a shadow affect, using the same parameters used in design software such as Sketch, Zeplin, and others.
+    /// This Extension originates from here: https://stackoverflow.com/questions/34269399/how-to-control-shadow-spread-and-blur/34270362
+    ///
+    /// - Parameters:
+    ///   - color: the tint of the shadow effect.  Default is UIColor.black
+    ///   - alpha: opacity of the color applied to the shaddow effect.  Default is 0.5
+    ///   - x: the horizontal offset applied to the shadow effect.  Default is 0.
+    ///   - y: the vertical offset applied to the shadow effect.  Default is 2.
+    ///   - blur: if set to 0 the shadow will be sharp, the higher the number, the more blurred it will be.  Default is 4.
+    ///   - spread: positive values increase the size of the shadow, negative values decrease the size. Default is 0 (the shadow is same size as blur).
+    public func applyShadow(color: UIColor = .black, alpha: Float = 0.5, xOffset: CGFloat = 0, yOffset: CGFloat = 2, blur: CGFloat = 4, spread: CGFloat = 0) {
+        shadowColor = color.cgColor
+        shadowOpacity = alpha
+        shadowOffset = CGSize(width: xOffset, height: yOffset)
+        shadowRadius = blur / 2.0
+        if spread == 0 {
+            shadowPath = nil
+        } else {
+            let dirX = -spread
+            let rect = bounds.insetBy(dx: dirX, dy: dirX)
+            shadowPath = UIBezierPath(rect: rect).cgPath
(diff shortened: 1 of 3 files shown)


Sources/Extensions/CoreAnimation/CALayerExtensions.swift
```diff
@@ -0,0 +1,34 @@
...
+    /// This Extension originates from here: https://stackoverflow.com/questions/34269399/how-to-control-shadow-spread-and-blur/34270362
+    ///
+    /// - Parameters:
+    ///   - color: the tint of the shadow effect.  Default is UIColor.black
+    ///   - alpha: opacity of the color applied to the shaddow effect.  Default is 0.5
+    ///   - x: the horizontal offset applied to the shadow effect.  Default is 0.
+    ///   - y: the vertical offset applied to the shadow effect.  Default is 2.
+    ///   - blur: if set to 0 the shadow will be sharp, the higher the number, the more blurred it will be.  Default is 4.
+    ///   - spread: positive values increase the size of the shadow, negative values decrease the size. Default is 0 (the shadow is same size as blur).
+    public func applyShadow(color: UIColor = .black, alpha: Float = 0.5, xOffset: CGFloat = 0, yOffset: CGFloat = 2, blur: CGFloat = 4, spread: CGFloat = 0) {
```
SD10: Would be better to use a `CGColor` here

mmdock: ah yeah, should stick to the CoreGraphics and CoreAnimation stuff

Sources/Extensions/CoreAnimation/CALayerExtensions.swift
```diff
@@ -0,0 +1,34 @@
...
+        shadowColor = color.cgColor
+        shadowOpacity = alpha
+        shadowOffset = CGSize(width: xOffset, height: yOffset)
+        shadowRadius = blur / 2.0
```
guykogus: Why is it specifically half of the "`blur`"? It's a seemingly random calculation. It also doesn't seem to fit with the description of `shadowRadius` in the documentation:
> The blur radius (in points) used to render the layer’s shadow.

Sources/Extensions/CoreAnimation/CALayerExtensions.swift
guykogus: These default values seem a bit random. Are they simply what you use for your project? I think it would make more sense to use the default values for their mapped values (e.g. `(0.0, -3.0)` for `xOffset` and `yOffset` respectively, `3.0` for `blur`).

Sources/Extensions/CoreAnimation/CALayerExtensions.swift
mmdock: @guykogus the blur applied by design tools such as Sketch, Zeplin, etc doesn't match what shadowRadius applies.  In order to match those design tools, you must half the blur value before applying it to the shadowRadius.

mmdock:
CALayer isn't supported by watchOS, so checked for that.

Next Step is to determine a good set of defaults.  In my most recent push I made the color not have a default value though.

Any suggestions @guykogus ?

SD10:
Omg I just merged the wrong PR 🙄

BennX:
@SD10 I was totally confused when i saw me pinged with this PR merged 😆

SD10:
Looks like this can't be reopened. Sorry for the inconvenience @mmdock 😐

mmdock:
@SD10  hahah I'll make the request again later tonight.  As for why it can't be reopened, see [this link](https://github.com/isaacs/github/issues/361)

PR #466, opened 2018-05-03
Add `isBetween(min:max:)` and `clamped(min:max:)` Comparable extensions

freak4pc:
🚀

This PR adds two extensions to `Comparable`

`isBetween(min:max:)` returns whether or not a `Comparable` is bounded in a range.

`clamp(min:max)` limits a `Comparable` within the provided bounds.

### DIFF
--- a/Sources/Extensions/SwiftStdlib/ComparableExtensions.swift
+++ b/Sources/Extensions/SwiftStdlib/ComparableExtensions.swift
@@ -1,0 +1,43 @@
+//
+//  ComparableExtensions.swift
+//  SwifterSwift
+//
+//  Created by Shai Mishali on 5/4/18.
+//  Copyright © 2018 SwifterSwift
+//
+
+// MARK: - Methods
+public extension Comparable {
+    /// SwifterSwift: Returns true if value is between provided `min`
+    /// and `max`.
+    ///
+    ///    1.isBetween(min: 5, max: 7) // false
+    ///    7.isBetween(min: 6, max: 12) // true
+    ///    date.isBetween(min: date1, max: date2)
+    ///    "c".isBetween(min: "a", max: "d") // true
+    ///    0.32.isBetween(min: 0.31, max: 0.33) // true
+    ///
+    /// - parameter min: Minimum comparable value.
+    /// - parameter max: Maximum comparable value.
+    ///
+    /// - returns: `true` if value is between `min` and `max`, `false` otherwise.
+    public func isBetween(min: Self, max: Self) -> Bool {
+        return self >= min && self <= max
+    }
+
+    /// SwifterSwift: Returns value limited within the provided range
+    /// between `min` and `max`.
+    ///
+    ///     1.clamped(min: 3, max: 8) // 3
+    ///     4.clamped(min: 3, max: 7) // 4
+    ///     "c".clamped(min: "e", max: "g") // "e"
+    ///     0.32.clamped(min: 0.1, max: 0.29) // 0.29
+    ///
+    /// - parameter min: Lower bound to limit the value to.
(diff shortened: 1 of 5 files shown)


Sources/Extensions/SwiftStdlib/ComparableExtensions.swift
```diff
@@ -0,0 +1,43 @@
...
+    ///    7.isBetween(min: 6, max: 12) // true
+    ///    date.isBetween(min: date1, max: date2)
+    ///    "c".isBetween(min: "a", max: "d") // true
+    ///    0.32.isBetween(min: 0.31, max: 0.33) // true
+    ///
+    /// - parameter min: Minimum comparable value.
+    /// - parameter max: Maximum comparable value.
+    ///
+    /// - returns: `true` if value is between `min` and `max`, `false` otherwise.
+    func isBetween(min: Self, max: Self) -> Bool {
```
SD10: @freak4pc Would you mind explicitly marking these two extensions as `public`? I know they're already public due to the scope of the extension, but for the sake of consistency in our codebase.

Tests/SwiftStdlibTests/ArrayExtensionsTests.swift
```diff
@@ -242,9 +242,11 @@ final class ArrayExtensionsTests: XCTestCase {
 
     func testGroupBy() {
         let array: [String] = ["james", "irving", "jordan", "jonshon", "iverson"]
-        let grouped = array.groupByKey { element -> String in
+
```
SD10: Good catch? What happened here 🤔

Sources/Extensions/SwiftStdlib/ComparableExtensions.swift
freak4pc: Done & rebased

Tests/SwiftStdlibTests/ArrayExtensionsTests.swift
freak4pc: Not entirely sure, also check out the edit in the CHANGELOG, seemed messed up as well

jdisho:
Why are you calling it `clamp` @freak4pc ? 😊

freak4pc:
@jdisho It's the common name for this kind of behavior:

https://developer.apple.com/documentation/swift/range/1786602-clamped

http://en.cppreference.com/w/cpp/algorithm/clamp

jdisho:
@freak4pc Ah good to know! Thanks! 👍

LucianoPAlmeida:
Hey guys :)) 
@freak4pc Nice extensions 💯 
This is merged but just point few things, that maybe we can have as future improvements if you guys think they are valid to the discuss :))
About `7.isBetween(min: 6, max: 12)` I don't know, I think this is a behavior that can be easily achieved with `6...12~=7` or `(6...12).contains(7)` But is still a valid extension :))
The `clamped()` extension, instead of using two parameters min and max, what do you guys think about using a signature with a range type like 
```swift 
 public func clamped(to range: ClosedRange<Self>) -> Self {
        return max(range.lowerBound, min(self, range.upperBound))
 }

// Used like "c".clamped(to: "e"..."g")
```
I think it can be a more clear API and that is a signature more like [this Apple one](https://developer.apple.com/documentation/swift/range/1786602-clamped)

@freak4pc Maybe you already thought about those points, but any way, let me know if you think they are valid points to discuss :)) 
cc @SD10

freak4pc:
Hey @LucianoPAlmeida . 

`ClosedCountableRange` only supports `Strideable` types. Notice that all your examples are using Numerics. Using that method you won't be able to confirm a String is in range or that a Date is in range, etc. `Comparable` is much more wide in its coverage.

Hope this clears up the reasoning on not using a range type.

PR #467, opened 2018-05-03
Value for query key

jdisho:
This PR 🚀 will add one extension to `URL`: value(for:) which returns the of a query key.

## Checklist

### DIFF
--- a/Sources/Extensions/Foundation/URLExtensions.swift
+++ b/Sources/Extensions/Foundation/URLExtensions.swift
@@ -63,6 +63,20 @@ public extension URL {
 		self = appendingQueryParameters(parameters)
 	}
 
+    /// SwifterSwift: Get value of a query key.
+    ///
+    ///    var url = URL(string: "https://google.com?code=12345")!
+    ///    url.value(for: code) -> "12345"
+    ///
+    /// - Parameter: Key of a query.
+    public func value(for queryKey: String) -> String? {
+        let stringURL = self.absoluteString
+        guard let items = URLComponents(string: stringURL)?.queryItems else { return nil }
+        for item in items where item.name == queryKey {
+            return item.value
+        }
+        return nil
+    }
 }
 
 // MARK: - Methods
--- a/Sources/Extensions/Foundation/URLExtensions.swift
+++ b/Sources/Extensions/Foundation/URLExtensions.swift
@@ -77,6 +77,31 @@ public extension URL {
         }
         return nil
     }
+    /// SwifterSwift: Returns a new URL by removing all the path components.
+    ///
+    ///     let url = URL(string: "https://domain.com/path/other")!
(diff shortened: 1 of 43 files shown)


Sources/Extensions/Foundation/URLExtensions.swift
```diff
@@ -63,6 +63,20 @@ public extension URL {
 		self = appendingQueryParameters(parameters)
 	}
 
+    /// SwifterSwift: Get value of a query key.
+    ///
+    ///    var url = URL(string: "https://google.com?code=12345")!
+    ///    url.value(for: code) -> "12345"
+    ///
+    /// - Parameter: Key of a query.
+    public func value(for queryKey: String) -> String? {
```
LucianoPAlmeida: Hey @jdisho I'm not sure about this naming here, to me `url.value(for:)` even with queryKey variable name,  doesn't make clear that is a value for a query parameter, seems like can be any value, I don't know, maybe something like `queryParameterValue(for key: String)`, but it has the con that is more verbose, but is more clear what it does.
What do you think? 
Note: I'm not good at naming this suggestion is probably not good naming at all 😂 
Let know your thoughts on that :))

Tests/FoundationTests/URLExtensionsTests.swift
```diff
@@ -37,6 +37,12 @@ final class URLExtensionsTests: XCTestCase {
 		XCTAssertEqual(parameters["empty"], nil)
 	}
 
+    func testValueForQueryKey() {
```
LucianoPAlmeida: Also, can you add a test case for a non-existing key?

Sources/Extensions/Foundation/URLExtensions.swift
SD10: I agree, and `value(for:)` is a pretty common API for stuff like NSUserDefaults and collections. Something like `queryValue` or `parameterValue` could be good names too. Good suggestion 👍

jdisho:
Thanks for the feedback guys. I applied it!

Sources/Extensions/Foundation/URLExtensions.swift
```diff
@@ -63,6 +63,20 @@ public extension URL {
 		self = appendingQueryParameters(parameters)
 	}
 
+    /// SwifterSwift: Get value of a query key.
+    ///
+    ///    var url = URL(string: "https://google.com?code=12345")!
+    ///    queryValue(for: "code") -> "12345"
+    ///
+    /// - Parameter: Key of a query.
```
SD10: This should be:
`/// - Parameter key: The key of a query value.`

jdisho: Good catch.

SD10:
Awesome! Thanks @jdisho @LucianoPAlmeida @oliviabrown9 👍

PR #468, opened 2018-05-04
Fix Comparable extensions to used `ClosedRange` vs `min/max`.

freak4pc:
Fixes concerns raised in #466. 

Didn’t add a CHANGELOG since its a recent addition but LMK if its needed.

### DIFF
--- a/Sources/Extensions/SwiftStdlib/ComparableExtensions.swift
+++ b/Sources/Extensions/SwiftStdlib/ComparableExtensions.swift
@@ -8,36 +8,34 @@
 
 // MARK: - Methods
 public extension Comparable {
-    /// SwifterSwift: Returns true if value is between provided `min`
-    /// and `max`.
+    /// SwifterSwift: Returns true if value is in the provided range.
     ///
-    ///    1.isBetween(min: 5, max: 7) // false
-    ///    7.isBetween(min: 6, max: 12) // true
-    ///    date.isBetween(min: date1, max: date2)
-    ///    "c".isBetween(min: "a", max: "d") // true
-    ///    0.32.isBetween(min: 0.31, max: 0.33) // true
+    ///    1.isBetween(5...7) // false
+    ///    7.isBetween(6...12) // true
+    ///    date.isBetween(date1...date2)
+    ///    "c".isBetween(a...d) // true
+    ///    0.32.isBetween(0.31...0.33) // true
     ///
     /// - parameter min: Minimum comparable value.
     /// - parameter max: Maximum comparable value.
     ///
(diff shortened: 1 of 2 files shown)


freak4pc:
IRT to the first method, I think it might be more explicit to have `isBetween` but if you think the standard `range` syntax is sufficient, I can get rid of it

SD10:
I thought about this when reviewing the initial PR but didn’t want to make the extension dependent on the range type. My use case was comparing two custom `Comparable` types that are not ranges. Am I missing something? cc @LucianoPAlmeida

LucianoPAlmeida:
Hey @SD10  :)) 
Range types are bounded to comparable. So you can create any range type (Range, ClosedRange, PartialRangeTo... and so on) with any custom type since this type conforms to comparable. I mean the only requirement to create a range with a type is this type to be comparable, so basically, any comparable can be made a range :)) 
I did a blog post a few months ago about range types in general, you can see [here](https://medium.com/@lucianoalmeida1/exploring-range-types-in-swift-e0e7ab27ab79) with more details.

freak4pc:
@LucianoPAlmeida except for `ClosedCountableRange`, which also needs `Strideable` :)

SD10:
@LucianoPAlmeida so if I have a `User` model that conforms to `Comparable`, I can just write this?
```swift
struct User: Comparable {}
let usr1 = User()
let usr2 = User()
usr1...usr2 // <-- this is valid?
```

freak4pc:
@SD10 Yes it is :) 

003a88e0-b233-4781-840b-ab7770d4fb0b

LucianoPAlmeida:
@SD10 Sure, this is valid :))
@freak4pc  That's right `CountableClosedRange`has Stridable requirements :))

SD10:
iiiiiinteresting 🙃 I feel like I just learned something great. Thanks guys!

LucianoPAlmeida:
@freak4pc  About `isBetween` I agree that is more explicity :)) 
I have a proposal maybe for a future extension, to have something like
```swift
extension Comparable {
    func isBounded<R: RangeExpression>(to range: R) -> Bool where R.Bound == Self {
        return range.contains(self)
    }
}
// Can support all ranges
5.isBounded(to: 5...)
6.isBounded(to: 5...10)
7.isBounded(to: 5..<10)
6.isBounded(to: ...5)
```
I don't  mean to replace `isBetween` because is between naming makes total sense to receive a ClosedRange :))

freak4pc:
@LucianoPAlmeida I like that idea ! Definitely worth having.

PR #479, opened 2018-05-15
create removeElement(_) in Array extension.

jdisho:
This PR introduces `removeElement(_ element:)`, which removes first instance of an element from array. 🎉

## Checklist

### DIFF
--- a/Sources/Extensions/SwiftStdlib/ArrayExtensions.swift
+++ b/Sources/Extensions/SwiftStdlib/ArrayExtensions.swift
@@ -231,6 +231,23 @@ public extension Array where Element: Equatable {
 		return indices
 	}
 
+    @discardableResult
+    /// SwifterSwift: Remove first instance of an element from array.
+    ///
+    ///        [1, 2, 2, 3, 4, 2, 5].removeElement(2) -> [1, 2, 3, 4, 2, 5]
+    ///        ["h", "e", "l", "l", "o"].removeElement("e") -> ["h", "l", "l", "o"]
+    ///
+    /// - Parameter element: element to remove.
+    /// - Returns: the removed element, otherwise nil.
+    public mutating func removeElement(_ element: Element) -> Element? {
+        if count > 0 {
+            for index in startIndex ..< endIndex where self[index] == element {
+                return self.remove(at: index)
+            }
+        }
+        return nil
+    }
+
 	@discardableResult
 	/// SwifterSwift: Remove all instances of an item from array.
 	///
--- a/Sources/Extensions/SwiftStdlib/ArrayExtensions.swift
+++ b/Sources/Extensions/SwiftStdlib/ArrayExtensions.swift
@@ -210,6 +210,19 @@ public extension Array {
 		return self
 	}
 
+    /// SwifterSwift: Removes the first element of the collection which satisfies the given predicate.
+    ///
+    ///        [1, 2, 2, 3, 4, 2, 5].removeFirst { $0 % 2 == 0 } -> [1, 2, 3, 4, 2, 5]
+    ///        ["h", "e", "l", "l", "o"].removeFirst { $0 == "e" } -> ["h", "l", "l", "o"]
+    ///
+    /// - Parameter predicate: A closure that takes an element as its argument and returns a Boolean value that indicates whether the passed element represents a match.
+    /// - Returns: The first element for which predicate returns true, after removing it. If no elements in the collection satisfy the given predicate, returns `nil`.
+    @discardableResult
+    public mutating func removeFirst(where predicate: (Element) throws -> Bool) rethrows -> Element? {
+        guard let index = try index(where: predicate) else { return nil }
+        return remove(at: index)
+    }
+
 }
 
 // MARK: - Methods (Equatable)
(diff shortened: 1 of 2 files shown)


Sources/Extensions/SwiftStdlib/ArrayExtensions.swift
```diff
@@ -231,6 +231,23 @@ public extension Array where Element: Equatable {
...
 
+    @discardableResult
+    /// SwifterSwift: Remove first instance of an element from array.
+    ///
+    ///        [1, 2, 2, 3, 4, 2, 5].removeElement(2) -> [1, 2, 3, 4, 2, 5]
+    ///        ["h", "e", "l", "l", "o"].removeElement("e") -> ["h", "l", "l", "o"]
+    ///
+    /// - Parameter element: element to remove.
+    /// - Returns: the removed element, otherwise nil.
+    public mutating func removeElement(_ element: Element) -> Element? {
```
SD10: This doesn't give any indication to the user that only the first element is removed.
Maybe `removeFirst(where condition: (Element) -> Bool) -> Element?`

Also, this can't be done by the Swift Standard Library? 🤔

Sources/Extensions/SwiftStdlib/ArrayExtensions.swift
```diff
@@ -231,6 +231,23 @@ public extension Array where Element: Equatable {
...
+        if count > 0 {
```
SD10: I would prefer to use a `guard` here

Sources/Extensions/SwiftStdlib/ArrayExtensions.swift
jdisho: `removeElement` removes only the specified element. Why should it be puzzling for the user?

jdisho: Btw this can’t be done by Swift lib. 
You can remove based on the `index` or by using `filter`

SD10: > removeElement removes only the specified element. Why should it be puzzling for the user?

Because the name doesn't convey to the user that **we're only removing the first element, not all elements**

guykogus: So I'm sure that I added this comment already, but it seems to be gone. I agree that the name isn't perfectly clear. How about `removeFirst(element: Element)`? You can't use `removeFirst(_ element: Element)` because then it would be ambiguous, if `Element == Int`, with `removeFirst(_ n: Int)`.

Sources/Extensions/SwiftStdlib/ArrayExtensions.swift
guykogus: You can make the method more concise:
```Swift
extension RangeReplaceableCollection where Element: Equatable {
    @discardableResult
    public mutating func removeFirst(element: Element) -> Element? {
        guard let index = index(where: { $0 == element }) else { return nil }
        return remove(at: index)
    }
}
```
Maybe making it an extension of `RangeReplaceableCollection` is a bit excessive... 😂

LucianoPAlmeida: @guykogus @jdisho @SD10  Actually, since the element is equatable we can use `index(of: Element)` instead of the where closure :))

SD10: I'm still wondering why not just make it:
`removeFirst(where condition: (Element) -> Bool) -> Element?`
You can do anything you want with this

SD10:
```swift
let first = [1, 2, 3].removeFirst(where: {$0 == 2})
```

LucianoPAlmeida: @SD10 I agree, It can have a lot more of use cases :))  And also, if you still want the Equatable one, you can build it in terms of this just simply calling a closure that does equal-equal, as you mention

guykogus: Why not have both?

```Swift
extension Array {
    @discardableResult
    public mutating func removeFirst(where condition: (Element) -> Bool) -> Element? {
        guard let index = index(where: condition) else { return nil }
        return remove(at: index)
    }
}

extension Array where Element: Equatable {
    @discardableResult
    public mutating func removeFirst(element: Element) -> Element? {
        guard let index = index(of: element) else { return nil }
        return remove(at: index)
    }
}
```

SD10: Seems like this PR has gone stale. I agree that `removeFirst(where:)` makes an existing task simpler, but TBH I don't see why we need both versions. It just adds boilerplate code to the library. 

Open to putting this up to a vote.
👍  for `removeFirst(where: (Element) -> Bool)`
👎  for  both `removeFirst(where: (Element) -> Bool)` && `removeFirst(element: Element)`

Sources/Extensions/SwiftStdlib/ArrayExtensions.swift
SD10: Needs to be throwing if we do `removeFirst(where:)`

Sources/Extensions/SwiftStdlib/ArrayExtensions.swift
SD10: cc @jdisho ☝️ If you're open to changing this to `removeFirst(where: throws (Element) -> Bool) -> rethrows Element?` we can get this one through

jdisho: Yes we can 😊 @SD10 
Btw, sorry for the delay, I will finish it asap.

guykogus:
You'd left it for a while so thought I'd help out. Sorry if that upset you. You're always welcome to revert my change ;)

jdisho:
>You'd left it for a while so thought I'd help out. Sorry if that upset you. You're always welcome to revert my change ;)

@guykogus

I always ask when I take over someone’s PR, that’s why I expressed my opinion. 

Regarding the changes, doesn’t make sense to revert it. It looks as expected 😊


SD10:
Thank you @jdisho @guykogus for this one 👍 We're sorry we didn't ask you before continuing work on this @jdisho

PR #482, opened 2018-05-19
Just making removeAllkeys more generic 

LucianoPAlmeida:
Just making removeAllkeys more generic so we can pass any sequence.
## Checklist

### DIFF
--- a/Sources/Extensions/SwiftStdlib/DictionaryExtensions.swift
+++ b/Sources/Extensions/SwiftStdlib/DictionaryExtensions.swift
@@ -30,8 +30,8 @@ public extension Dictionary {
     ///		dict.keys.contains("key2") -> false
     ///
     /// - Parameter keys: keys to be removed
-    public mutating func removeAll(keys: [Key]) {
-        keys.forEach({ removeValue(forKey: $0)})
+    mutating func removeAll<S: Sequence>(keys: S) where S.Element == Key {
+        Set<Key>(keys).forEach({ removeValue(forKey: $0) })
     }
 
     /// SwifterSwift: JSON Data from dictionary.
@@ -79,7 +79,7 @@ public extension Dictionary {
 }
 
 // MARK: - Methods (ExpressibleByStringLiteral)
-public extension Dictionary where Key: ExpressibleByStringLiteral {
+public extension Dictionary where Key: StringProtocol {
 
     /// SwifterSwift: Lowercase all keys in dictionary.
     ///
@@ -148,7 +148,7 @@ public extension Dictionary {
     ///   - lhs: dictionary
     ///   - rhs: array with the keys to be removed.
     /// - Returns: a new dictionary with keys removed.
-    public static func - (lhs: [Key: Value], keys: [Key]) -> [Key: Value] {
+    public static func - <S: Sequence>(lhs: [Key: Value], keys: S) -> [Key: Value] where S.Element == Key {
         var result = lhs
         result.removeAll(keys: keys)
         return result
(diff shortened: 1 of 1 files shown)


Sources/Extensions/SwiftStdlib/DictionaryExtensions.swift
```diff
@@ -30,8 +30,8 @@ public extension Dictionary {
     ///		dict.keys.contains("key2") -> false
     ///
     /// - Parameter keys: keys to be removed
-    public mutating func removeAll(keys: [Key]) {
-        keys.forEach({ removeValue(forKey: $0)})
+    public mutating func removeAll<S: Sequence>(keys: S) where S.Element == Key {
+        Set<Key>(keys).forEach({ removeValue(forKey: $0) })
```
guykogus: Is it necessary to create the `Set` at all? I think this qualifies as [premature optimisation](http://wiki.c2.com/?PrematureOptimization).

Let's calculate:
`n` = size of `self`
`k1` = original size of `keys`
`k2` = size of `Set(keys)`
Creating the set costs `O(k1)` (at best).
According to the documentation, `removeValue(forKey:)` costs `O(n)`
Therefore, removing values costs `O(k2 * n)` (assuming no keys were actually removed, for arguments' sake)
So the total cost = `O(k1 + (k2 * n))`
There's a decent chance that callers of this method won't use repeated keys (but who knows), so it would be more like `O(2 * k1 * n)`

If you don't create the set then it would simply be `O(k1 * n)` (i.e. twice as fast). And if somebody finds in writing their app that this function takes a long time because of repeated keys then they could optimise it themselves by creating the set.

guykogus: I think SwiftLint gives a warning when you can remove the parenthesis, e.g. in `forEach`. So it would just be
`forEach { removeValue(forKey: $0) }`

LucianoPAlmeida: I see the point of letting this decision to the user, I think it's o good option :))

LucianoPAlmeida: Btw nice article about premature optimisation, thank's for sharing 👍

guykogus: My old boss used to always call us up on it. He successfully drilled it into my head 🤯

Sources/Extensions/SwiftStdlib/DictionaryExtensions.swift
```diff
@@ -30,8 +30,8 @@ public extension Dictionary {
     ///		dict.keys.contains("key2") -> false
     ///
     /// - Parameter keys: keys to be removed
-    public mutating func removeAll(keys: [Key]) {
-        keys.forEach({ removeValue(forKey: $0)})
+    public mutating func removeAll<S: Sequence>(keys: S) where S.Element == Key {
+        keys.forEach({ removeValue(forKey: $0) })
```
guykogus: My old comment seems to have disappeared, but I think it's better to remove the parenthesis for `forEach`. I.e. `forEach { removeValue(forKey: $0) }`. It's what autocomplete does and I think SwiftLint complains about it too.

LucianoPAlmeida: Swiftlint doesn't complain, but no problem changing it

guykogus: Not all SwiftLint rules are applied by default. You're welcome to add it: `unneeded_parentheses_in_closure_argument` ;)
https://github.com/realm/SwiftLint/blob/master/Source/SwiftLintFramework/Rules/UnneededParenthesesInClosureArgumentRule.swift

Sources/Extensions/SwiftStdlib/DictionaryExtensions.swift
SD10: Good thoughts

SD10:
I approve this pull request

PR #486, opened 2018-05-28
update addsubviews

Yoloabdo:
I like this header from Johnsundell and it feels and looks way better 
source: https://github.com/JohnSundell/SwiftTips#55-dropping-suffixes-from-method-names-to-support-multiple-arguments

## Checklist

- [X] New extensions are written in Swift 4.
- [X] New extensions support iOS 8.0+ / tvOS 9.0+ / macOS 10.10+ / watchOS 2.0+.

### DIFF
--- a/Sources/Extensions/UIKit/UIViewExtensions.swift
+++ b/Sources/Extensions/UIKit/UIViewExtensions.swift
@@ -264,7 +264,7 @@ public extension UIView {
 	/// SwifterSwift: Add array of subviews to view.
 	///
 	/// - Parameter subviews: array of subviews to add to self.
-	public func addSubviews(_ subviews: [UIView]) {
+	public func add(_ subviews: UIView...) {
 		subviews.forEach({ self.addSubview($0) })
 	}
 
--- a/Tests/UIKitTests/UIViewExtensionsTests.swift
+++ b/Tests/UIKitTests/UIViewExtensionsTests.swift
@@ -201,13 +201,13 @@ final class UIViewExtensionsTests: XCTestCase {
 		let view = UIView(frame: frame)
 		XCTAssertEqual(view.subviews.count, 0)
 
-		view.addSubviews([UIView(), UIView()])
+		view.add(UIView(), UIView())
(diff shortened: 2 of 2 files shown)


Sources/Extensions/UIKit/UIViewExtensions.swift
```diff
@@ -264,7 +264,7 @@ public extension UIView {
 	/// SwifterSwift: Add array of subviews to view.
 	///
 	/// - Parameter subviews: array of subviews to add to self.
-	public func addSubviews(_ subviews: [UIView]) {
+	public func add(_ subviews: UIView...) {
```
SD10: @Yoloabdo Thanks for taking the time to submit this PR 💯 I'm going to disagree that this is a better API. The name `addSubviews` builds on the `addSubview` API of UIKit. In my opinion, the name `add` is not descriptive enough.

 I'd also prefer to have the argument be typed as an array rather than having a variadic parameter.

Yoloabdo: alright then, it's a matter of taste by the end I guess 👍

omaralbeik: Hey @Yoloabdo, thank you for your PR :)
It is actually not a matter of taste, our goal here at SwifterSwift is to keep naming conventions very close to Apple's.
- `add` is a very generic name for this case, you can add a subview, layout constraints, animations, gesture recognizers, and more to a view, that's why the extra `views` keyword adds a value here as @SD10 mentioned above.
- a variadic parameter accepts **zero or more** values of a UIView, users will be able to call the function without passing any arguments, like:
```swift
view.add()
```
which I think is not the desired usecase for this extension

SD10: @omaralbeik Good addition on the behavior of a variadic 👍

SD10:
@Yoloabdo Thanks for taking the time to try to improve SwifterSwift 💯 👍 We appreciate your suggestion but I'm going to close this for the reasons described above

PR #490, opened 2018-05-28
Add SKNode Descendants

oliviabrown9:
## Checklist

Note: I'm a bit nervous that I updated the podspec correctly, so I would especially appreciate if someone could review that (in addition to reviewing in general!)


oliviabrown9:
Any advice on the Travis CI failures?

Tests/SpriteKitTests/SpriteKitTests.swift
```diff
@@ -0,0 +1,22 @@
+//
+//  SpriteKitTests.swift
+//  SwifterSwift
+//
+//  Created by Olivia Brown on 5/28/18.
+//  Copyright © 2018 SwifterSwift
+//
+
+import XCTest
+import SpriteKit
```
SD10: Instead of importing `SpriteKit`, let's do `@testable import SwifterSwift`. It looks like CI can't find the method `descendants` from `SwifterSwift`.

Sources/Extensions/SpriteKit/SKNodeExtensions.swift
```diff
@@ -0,0 +1,22 @@
...
+//  SpriteKitExtensions.swift
+//  SwifterSwift
+//
+//  Created by Olivia Brown on 5/28/18.
+//  Copyright © 2018 SwifterSwift
+//
+
+import SpriteKit
+
+// MARK: - Properties
```
SD10: trivial, but this should be `Methods`

oliviabrown9:
@SD10 Fixed! Thanks for debugging with me 😁

Sources/Extensions/SpriteKit/SKNodeExtensions.swift
```diff
@@ -0,0 +1,22 @@
...
+import SpriteKit
+
+// MARK: - Methods
+public extension SKNode {
+    /// SwifterSwift: Return an array of all SKNode descendants
+    ///
+    ///         mySKNode.descendants() -> [childNodeOne, childNodeTwo]
+    ///
+    public func descendants() -> [SKNode] {
+        let firstLevelDescendants = children.compactMap { $0 }
```
guykogus: What's the need for `compactMap`? I don't think it's useful here since the values of children are non-nil.

guykogus: You can also place this whole function into a single line:
```Swift
return children + children.reduce(into: [SKNode]()) { $0 += $1.descendants() }
```

guykogus: It may not matter too much, but I just want to comment that it's not a standard search path (as far as I'm aware) that's been implemented here. It's neither depth-first nor breadth-first, but rather a kind of hybrid. I guess it can be forgiven if you make it a one-liner 😛

oliviabrown9: Good catch on compactMap. Definitely a lot cleaner!

SD10: Ok @guykogus I see what you're saying here. I guess order _technically_ has meaning here cuz Array is an ordered collection. But I'd be fine with just getting the ball rolling. If someone wants to come back later and implement a standard search path then I'm open to a PR 👍

Sources/Extensions/SpriteKit/SKNodeExtensions.swift
```diff
@@ -0,0 +1,20 @@
+//
+//  SpriteKitExtensions.swift
+//  SwifterSwift
+//
+//  Created by Olivia Brown on 5/28/18.
+//  Copyright © 2018 SwifterSwift
+//
+
+import SpriteKit
```
omaralbeik: Please wrap this by
```swift
#if canImport(SpriteKit)
...
#endif
```

omaralbeik: SpriteKit is also available for macOS, tvOS, and watchOS, could you please update `SpriteKitExtensions.swift` target membership to be included in other targets as well :)

omaralbeik:
@LucianoPAlmeida we should add `SpriteKit extensions` to
- List of extensions
- installation via Cocoapods instructions
on readme before releasing 🙂

Sources/Extensions/SpriteKit/SKNodeExtensions.swift
SD10: I have to learn how to figure this out from the Xcodeproj diff without actually cloning the project

oliviabrown9: @omaralbeik I thought that I added it to all targets with https://github.com/SwifterSwift/SwifterSwift/pull/490/commits/0487ecbec8e668223d21e651f936bb4d36be1936. What am I missing?

SD10: yeah, I'm curious too @omaralbeik. CI shows this extension being tested on 3 platforms (we don't run tests for watchOS) 🤔

omaralbeik: hmm, yeah I see you've added it to all targets, maybe I forgot to fetch origin, sorry 😅

PR #496, opened 2018-06-02
Add containsDuplicates to check whether a collection contains duplicates

Vyax:
## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,13 +24,17 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 
 - **URL**
   - Added `deletingAllPathComponents()` and `deleteAllPathComponents()` to delete all path components from a URL. [#441](https://github.com/SwifterSwift/SwifterSwift/pull/441) by [setoelkahfi](https://github.com/setoelkahfi).
-
   - Added `queryValue(for:)` to get the value of a query key from a URL. [#467](https://github.com/SwifterSwift/SwifterSwift/pull/467) by [jdisho](https://github.com/jdisho).
+
 - **UITableView**:
   - Added `isValidIndexPath(_:)` method to check whether given IndexPath is valid within UITableView. [#441](https://github.com/SwifterSwift/SwifterSwift/pull/441) by [setoelkahfi](https://github.com/setoelkahfi).
   - Added `safeScrollToRow(at:at:animated:)` method to safely scroll UITableView to the given IndexPath. [#445](https://github.com/SwifterSwift/SwifterSwift/pull/445) by [setoelkahfi](https://github.com/setoelkahfi).
+
 - **Optional**:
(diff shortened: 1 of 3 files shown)


Sources/Extensions/SwiftStdlib/CollectionExtensions.swift
```diff
@@ -188,3 +188,15 @@ public extension Collection where Element: FloatingPoint {
     }
 
 }
+
+// MARK: - Methods (Hashable)
```
LucianoPAlmeida: I think here it should be 'Properties' instead of 'Methods' :))

Tests/SwiftStdlibTests/CollectionExtensionsTests.swift
```diff
@@ -151,4 +151,9 @@ final class CollectionExtensionsTests: XCTestCase {
         XCTAssertEqual([Double]().average(), 0)
     }
 
+    func testContainsDuplicates() {
+        XCTAssert(["a", "b", "b", "c"].containsDuplicates)
+        XCTAssertFalse(["a", "b", "c", "d"].containsDuplicates)
```
BennX: I'm missing the edge condition with an empty array.

Vyax: Okay, will be added.

Sources/Extensions/SwiftStdlib/CollectionExtensions.swift
```diff
@@ -188,3 +188,15 @@ public extension Collection where Element: FloatingPoint {
     }
 
 }
+
+// MARK: - Properties (Hashable)
+public extension Collection where Element: Hashable {
```
guykogus: Can you put this a level up and make it an extension of `Sequence`?

guykogus: Actually, I just realised `Sequence` doesn't have a `count`. You could still put it there, but you'd need to manually add each element to the set, and check before if it exists. e.g.
extension Sequence where Element: Hashable {
```Swift
extension Sequence where Element: Hashable {
    func containsDuplicates() -> Bool {
        var set = Set<Element>()
        for element in self {
            if !set.insert(element).inserted {
                return true
            }
        }
        return false
    }
}
```

Sources/Extensions/SwiftStdlib/CollectionExtensions.swift
```diff
@@ -188,3 +188,15 @@ public extension Collection where Element: FloatingPoint {
...
+
+    /// SwifterSwift: Checks whether a collection contains duplicates.
+    ///
+    /// - Returns: true if the receiver contains duplicates.
+    public var containsDuplicates: Bool {
```
guykogus: This should be a `func`, since it won't have constant runtime.

SD10:
@guykogus @BennX This should be good to go if anyone else wants to review

PR #497, opened 2018-06-04
Add 'removeRandomElement()' to Dictionary

Vyax:
…to remove a random element from a dictionary

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 # Upcoming release
 
 ### Added
+- **Dictionary**:
+  - Added `removeValueForRandomKey()` to remove a value for a random key from a dictionary. [#497](https://github.com/SwifterSwift/SwifterSwift/pull/497) by [vyax](https://github.com/vyax).
+- **RangeReplaceableCollection**:
+  - Added `removeRandomElement()` to remove a random element from a collection. [#497](https://github.com/SwifterSwift/SwifterSwift/pull/497) by [vyax](https://github.com/vyax).
 - **UIView**
   - Added `addGestureRecognizers(_:)` which accepts an array of `UIGestureRecognizer` to add multiple gesture recognizers to a view with one call. [#523](https://github.com/SwifterSwift/SwifterSwift/pull/523) by [moyerr](https://github.com/moyerr)
   - Added `removeGestureRecognizers(_:)` which accepts an array of `UIGestureRecognizer` to remove multiple gesture recognizers from a view with one call. [#523](https://github.com/SwifterSwift/SwifterSwift/pull/523) by [moyerr](https://github.com/moyerr)
--- a/Sources/Extensions/SwiftStdlib/DictionaryExtensions.swift
+++ b/Sources/Extensions/SwiftStdlib/DictionaryExtensions.swift
@@ -34,6 +34,13 @@ public extension Dictionary {
         keys.forEach { removeValue(forKey: $0) }
     }
 
+    /// SwifterSwift: Remove a value for a random key from the dictionary.
+    @discardableResult public mutating func removeValueForRandomKey() -> Value? {
+        guard !isEmpty else { return nil }
+        let key = Array(keys)[Int(arc4random_uniform(UInt32(keys.count)))]
+        return removeValue(forKey: key)
+    }
+
     /// SwifterSwift: JSON Data from dictionary.
     ///
     /// - Parameter prettify: set true to prettify data (default is false).
--- a/Tests/SwiftStdlibTests/DictionaryExtensionsTests.swift
+++ b/Tests/SwiftStdlibTests/DictionaryExtensionsTests.swift
@@ -26,6 +26,17 @@ final class DictionaryExtensionsTests: XCTestCase {
         XCTAssertFalse(dict.keys.contains("key2"))
     }
 
+    func testRemoveElementForRandomKey() {
+        var emptyDict = [String: String]()
+        XCTAssertNil(emptyDict.removeValueForRandomKey())
+
+        var dict = ["key1": "value1", "key2": "value2", "key3": "value3"]
+        let elements = dict.count
+        let removedElement = dict.removeValueForRandomKey()
+        XCTAssertEqual(elements - 1, dict.count)
+        XCTAssertFalse(dict.contains(where: {$0.value == removedElement}))
+    }
+
     func testJsonData() {
         let dict = ["key": "value"]
 
--- a/Sources/Extensions/SwiftStdlib/RangeReplaceableCollectionExtensions.swift
+++ b/Sources/Extensions/SwiftStdlib/RangeReplaceableCollectionExtensions.swift
@@ -61,4 +61,10 @@ extension RangeReplaceableCollection {
         guard let index = try index(where: predicate) else { return nil }
         return remove(at: index)
     }
+    
+    /// SwifterSwift: Remove a random value from the collection.
+    @discardableResult public mutating func removeRandomElement() -> Element? {
+        guard !isEmpty else { return nil }
+        return remove(at: index(startIndex, offsetBy: numericCast(arc4random_uniform(numericCast(count)))))
+    }
 }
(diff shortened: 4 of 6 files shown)


Sources/Extensions/SwiftStdlib/DictionaryExtensions.swift
```diff
@@ -34,6 +34,14 @@ public extension Dictionary {
         keys.forEach { removeValue(forKey: $0) }
     }
 
+    /// SwifterSwift: Remove a random element from the dictionary.
+    @discardableResult public mutating func removeRandomElement() -> Value? {
+        if let key = Array(keys).randomItem {
```
guykogus: Unfortunately you can't use `SwifterSwift` functions internally. The project is designed so that each function can be copy-pasted as-is into your app. The main reason is that the whole library itself can't safely be added as a dependency because of potential name clashes with other 3rd party libraries.

Vyax: Okay, I will adjust the code accordingly.

SD10:
@Vyax Thanks for your pull request! This definitely feels like an odd extension though. What's your use case?

BennX:
@SD10 I could imagin if you have something like a card game and you want to draw cards of a stack/array. But I agree, seems really odd/special.

oliviabrown9:
I’ve used a similar extension a few times in game development - I think it’s got a solid use case :)

SD10:
@oliviabrown9 @BennX Thanks for the feedback and use cases!

I wonder if this extension could be made on `Collection` to be more generic?

guykogus:
@SD10 this can't be made generic because `keys` are a specific property of Dictionary.
If we wanted, we could add an equivalent method to `RangeReplaceableCollection`:
```Swift
extension RangeReplaceableCollection {
    @discardableResult public mutating func removeRandomElement() -> Element {
        return remove(at: index(startIndex,
                                offsetBy: numericCast(arc4random_uniform(numericCast(count)))))
    }
}

var arr = [0, 1, 2, 3, 4]
arr.removeRandomElement() // 2
arr // [0, 1, 3, 4]
```

LucianoPAlmeida:
Hy guys 
I think since this is a Dictionary extension a better naming will be `removeRandomKey`, to me removing a key that removing an element.
Also, that way we can implement  `removeRandomElement` as a `RandomAccessCollection` extension as suggested :))

SD10:
@LucianoPAlmeida I think `removeRandomKey` feels like a weird name, the key is not being removed but rather the key/value pair -- ultimately the value. For example, for an array extension we may call it `removeRandomElement` and not `removeRandomIndex`

LucianoPAlmeida:
@SD10 That makes sense 👍 
But I'm sure the `removeRandomElement` is not expressive to dictionary and more to Collection/Sequence that has Collection.Element
I don't know, maybe `removeRandomValue` or `removeRandomKeyValuePair` ?

guykogus:
The method of `Dictionary` being called in the method  is `removeValue(forKey:)`, so we could call it `removeValueForRandomKey()`, to be more in line with the built-in function and expressive.

CHANGELOG.md
```diff
@@ -19,16 +19,19 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
...
   - Added `isValidIndexPath(_:)` method to check whether given IndexPath is valid within UITableView. [#441](https://github.com/SwifterSwift/SwifterSwift/pull/441) by [setoelkahfi](https://github.com/setoelkahfi).
   - Added `safeScrollToRow(at:at:animated:)` method to safely scroll UITableView to the given IndexPath. [#445](https://github.com/SwifterSwift/SwifterSwift/pull/445) by [setoelkahfi](https://github.com/setoelkahfi).
-
 - **Optional**:
   - Added `isNilOrEmpty` property to check whether an optional is nil or empty collection.
 - **UIWindow**:
   - Added `switchRootViewController` method to switch root view controller with animation. [#494](https://github.com/SwifterSwift/SwifterSwift/pull/494) by [omaralbeik](https://github.com/omaralbeik).
-
+- **Optional**:
+  - Added `isNilOrEmpty` property to check whether an optional is nil or empty collection.
```
guykogus: You've duplicated this line

Vyax:
I think this pull request should be ready for merging now.

CHANGELOG.md
```diff
@@ -4,36 +4,40 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
...
+- **Sequence**
+  - added `containsDuplicates()` to check whether a sequence contains duplicates. [#496](https://github.com/SwifterSwift/SwifterSwift/pull/496) by [@vyax](https://github.com/vyax).
+  - Added `single(where:)` to get the only element of a sequence that matches a given condition. [#483](https://github.com/SwifterSwift/SwifterSwift/pull/483) by [andlang](https://github.com/andlang).
+- **SKNode**:
+  - Added `descendants` method to get an array of all descendants of an SKNode. [#490](https://github.com/SwifterSwift/SwifterSwift/pull/490) by [oliviabrown9](https://github.com/oliviabrown9).
+- **UIEdgeInsets**
+  - Added `horizontal` and `vertical` properties. Also `init(inset:)` and `init(horizontal: vertical:)` initializers for convenience. [#500](https://github.com/SwifterSwift/SwifterSwift/pull/500) by [LucianoPAlmeida](https://github.com/LucianoPAlmeida).
+- **UIGestureRecognizer**:
+  - Added `removeFromView()` method to remove recognizer from the view the recognizer is attached to. [#456](https://github.com/SwifterSwift/SwifterSwift/pull/456) by [mmdock](https://github.com/mmdock).
 - **UIScrollView**:
```
SD10: What's up with this diff on the CHANGELOG?

Vyax: I sorted the changes alphabetically.

Tests/SwiftStdlibTests/DictionaryExtensionsTests.swift
```diff
@@ -26,6 +26,17 @@ final class DictionaryExtensionsTests: XCTestCase {
         XCTAssertFalse(dict.keys.contains("key2"))
     }
 
+    func testRemoveElementForRandomKey() {
+        var emptyDict = [String: String]()
+        emptyDict.removeValueForRandomKey()
```
guykogus: Do you want to use `XCTAssertNil`, like you did for `RangeReplaceableCollectionExtTests`?

Tests/SwiftStdlibTests/RangeReplaceableCollectionExtensionsTests.swift
```diff
@@ -0,0 +1,25 @@
...
+//  SwifterSwift
+//
+//  Created by Max Härtwig on 12.06.18.
+//  Copyright © 2018 SwifterSwift
+//
+
+import XCTest
+@testable import SwifterSwift
+
+final class RangeReplaceableCollectionExtTests: XCTestCase {
```
guykogus: The class name should match the file name.

SD10:
I forgot why this didn't get merged but now there's a few review conflicts + comments

Tests/SwiftStdlibTests/RangeReplaceableCollectionExtensionsTests.swift
Vyax: I agree, but swiftlint doesn't allow class names longer than 40 characters causing the Travis build to fail.

guykogus: You can hint to swiftlint to ignore a specific rule for a specific line by adding a comment above it like this:
`swiftlint:disable next identifier_name`
I'm not sure if `identifier_name` is the exact rule that's being broken by the long class name, but it's something like that.

LucianoPAlmeida: I believe we already have this file and class created on master, it was created in #512 if I'm remembering correctly :)

guykogus: Ah true, and the class name ended up being `RangeReplaceableCollectionTests`, I'm guessing to avoid that issue.

PR #503, opened 2018-06-17
Add new tests

omaralbeik:
Add new tests and increase test coverage before the next release.

### DIFF
--- a/Sources/Extensions/UIKit/UISearchBarExtensions.swift
+++ b/Sources/Extensions/UIKit/UISearchBarExtensions.swift
@@ -9,7 +9,7 @@
 #if canImport(UIKit)
 import UIKit
 
-#if !os(watchOS)
+#if os(iOS)
 // MARK: - Properties
 public extension UISearchBar {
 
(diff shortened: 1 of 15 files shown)


Sources/Extensions/UIKit/UISearchBarExtensions.swift
```diff
@@ -9,7 +9,7 @@
 #if canImport(UIKit)
 import UIKit
 
-#if !os(watchOS)
+#if os(iOS)
```
LucianoPAlmeida: @omaralbeik  UISearchBar is also available on tvOS SDK, any reason to make it just iOS?

Tests/SwifterSwiftTests.swift
```diff
@@ -54,7 +54,7 @@ final class SwifterSwiftTests: XCTestCase {
 
 		XCTAssertEqual(value, 0, "Debounced function does not get executed right away")
 
-		waitForExpectations(timeout: 2, handler: { _ in
+		waitForExpectations(timeout: 2.5, handler: { _ in
```
LucianoPAlmeida: Awesome 💯

Sources/Extensions/UIKit/UISearchBarExtensions.swift
omaralbeik: It looks like it is supported, but you can not create an instance of it for some reason, I couldn't run tests on tvOS due to the following errors
```
- 'init()' is unavailable
- 'init(frame:)' is unavailable
```
https://forums.developer.apple.com/thread/21421

Issue #427, opened 2018-03-30
Refactor ArrayExtensions


omaralbeik:
Some methods like `sum`, `average` can be moved from ArrayExtensions to CollectionExtensions, this will make them accessible for other Collection types as well.

LucianoPAlmeida:
Hey @omaralbeik I think some extension also can be moved to Sequence instead of Collection. 
For example: `count(where:)` and `forEachReversed()` I think could be on Sequence.

PR #512, opened 2018-07-02
Changing rotate/rotated Array extensions to RangeReplacebleCollection extensions

LucianoPAlmeida:
This is still related to #427 
Notice that still has some array extensions that could be refactored. 
@omaralbeik Should we reopened that? Create a new one what do you think? 
## Checklist

### DIFF
--- a/Sources/Extensions/SwiftStdlib/RangeReplaceableCollectionExtensions.swift
+++ b/Sources/Extensions/SwiftStdlib/RangeReplaceableCollectionExtensions.swift
@@ -9,21 +9,21 @@
 // MARK: - Methods
 extension RangeReplaceableCollection {
 
-    /// SwifterSwift: Returns a new rotated array by the given places.
+    /// SwifterSwift: Returns a new rotated collection by the given places.
     ///
     ///     [1, 2, 3, 4].rotated(by: 1) -> [4,1,2,3]
     ///     [1, 2, 3, 4].rotated(by: 3) -> [2,3,4,1]
     ///     [1, 2, 3, 4].rotated(by: -1) -> [2,3,4,1]
     ///
-    /// - Parameter places: Number of places that the array be rotated. If the value is positive the end becomes the start, if it negative it's that start becom the end.
-    /// - Returns: The new rotated array
+    /// - Parameter places: Number of places that the array be rotated. If the value is positive the end becomes the start, if it negative it's that start becom the end.
+    /// - Returns: The new rotated collection.
     public func rotated(by places: Int) -> Self {
         //Inspired by: https://ruby-doc.org/core-2.2.0/Array.html#method-i-rotate
         var copy = self
         return copy.rotate(by: places)
     }
 
-    /// SwifterSwift: Rotate the array by the given places.
+    /// SwifterSwift: Rotate the collection by the given places.
     ///
     ///     [1, 2, 3, 4].rotate(by: 1) -> [4,1,2,3]
     ///     [1, 2, 3, 4].rotate(by: 3) -> [2,3,4,1]
--- a/Sources/Extensions/SwiftStdlib/RangeReplaceableCollectionExtensions.swift
+++ b/Sources/Extensions/SwiftStdlib/RangeReplaceableCollectionExtensions.swift
@@ -36,7 +36,7 @@ extension RangeReplaceableCollection {
         guard places != 0 else { return self }
         let placesToMove = abs(places) >= count ? places%count : places
         if placesToMove > 0 {
-            let range = index(endIndex, offsetBy: -placeToMove)...
+            let range = index(endIndex, offsetBy: -placesToMove)...
             let slice = self[range]
             removeSubrange(range)
             insert(contentsOf: slice, at: startIndex)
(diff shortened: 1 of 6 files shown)


guykogus:
Is it possible to move even more into `RangeReplaceable`?

Sources/Extensions/SwiftStdlib/RangeReplaceableCollectionExtensions.swift
```diff
@@ -0,0 +1,50 @@
...
+    ///
+    ///     [1, 2, 3, 4].rotate(by: 1) -> [4,1,2,3]
+    ///     [1, 2, 3, 4].rotate(by: 3) -> [2,3,4,1]
+    ///     [1, 2, 3, 4].rotated(by: -1) -> [2,3,4,1]
+    ///
+    /// - Parameter places: The number of places that the array should be rotated. If the value is positive the end becomes the start, if it negative it's that start become the end.
+    /// - Returns: self after rotating.
+    @discardableResult
+    public mutating func rotate(by places: Int) -> Self {
+        guard places != 0 && places < count else { return self }
```
guykogus: Isn't it possible to use `places % count` to work with situations of `places >= count`?

LucianoPAlmeida:
Hey @guykogus :))
 Yeap, not many, but some still can be moved. For example [removeFirst(where:)](https://github.com/SwifterSwift/SwifterSwift/blob/d3c3fdba03cd876baf3c7f675f03ad7c3f32c811/Sources/Extensions/SwiftStdlib/ArrayExtensions.swift#L226) can be moved. Also `indices(of item: Element)` with a little refactor can be moved `RandomAccessCollection` and so on...

Sources/Extensions/SwiftStdlib/RangeReplaceableCollectionExtensions.swift
LucianoPAlmeida: Yeap, but I'm not sure if this could be a behaviour to implement inside the method :))

guykogus: It just seems like unexpected behaviour to me to `return self` just because the rotation size was too big... It's also troublesome that it doesn't check `abs(places) < count` because this is designed to handle both positive and negative values, and if the negative value is too high this method will crash...
I know you're just moving this method, but while we're here might as well fix it, no? :)

LucianoPAlmeida: For sure it's a nice catch @guykogus the negative too big. I will fix it :)) 
But to me seems weird for thinking in an API usage, for example, have a collection of 4, send 5 but it rotates 1. Maybe if it's well documented, but still I think the expected behavior is still not rotate it :)

guykogus: It doesn't rotate 1, it rotates 5, but since the size is 4 then the final result is equivalent to a rotation of 1.
The idea is to think of the array as a circular container, and when you spin a circle 540 degrees it's the same as spinning it 180, the final result is the same. If I call a method to rotate something 540 degrees and the final result is that it doesn't spin at all because the maximum acceptable input is 360, I would find that _very_ surprising. I think that it's the same here.

LucianoPAlmeida: I see... It makes sense and It seems like this is the same behavior on [ruby api](https://apidock.com/ruby/Array/rotate). So let's change it then :))

Sources/Extensions/SwiftStdlib/RangeReplaceableCollectionExtensions.swift
```diff
@@ -0,0 +1,51 @@
...
+    ///     [1, 2, 3, 4].rotate(by: 1) -> [4,1,2,3]
+    ///     [1, 2, 3, 4].rotate(by: 3) -> [2,3,4,1]
+    ///     [1, 2, 3, 4].rotated(by: -1) -> [2,3,4,1]
+    ///
+    /// - Parameter places: The number of places that the array should be rotated. If the value is positive the end becomes the start, if it negative it's that start become the end.
+    /// - Returns: self after rotating.
+    @discardableResult
+    public mutating func rotate(by places: Int) -> Self {
+        guard places != 0 else { return self }
+        let placesToMove = abs(places) >= count ? places%count : places
```
guykogus: Why not just `let placesToMove = places % count`? Result will be the same.

LucianoPAlmeida: Sure :))

guykogus:
@LucianoPAlmeida you want to merge this in, or are you waiting for @omaralbeik to approve?

LucianoPAlmeida:
@guykogus Not problem, it was indeed it, just waiting for @omaralbeik aproval, since he aproves you were totally right to merge :)) 
Btw thank's for taking care of it 👍

omaralbeik:
Thank you guys 💯

SD10:
Good work everyone 😎

PR #517, opened 2018-07-15
Fixes #514

### LINKED ISSUE #514: Fix email validation in StringExtensions
`isEmail` in [`StringExtensions.swift`](https://github.com/SwifterSwift/SwifterSwift/blob/3cb82030a13a40a751b60fc5c8802e31c998baea/Sources/Extensions/SwiftStdlib/StringExtensions.swift#L163) is not checking for spaces inside the email address

Example:
```swift
let email = "john dow@ap ple.com"
print(email.isEmail) // prints true even the email is not valid!
```

omaralbeik:
Used [RFC 5322](http://emailregex.com/) in `isEmail`, an email address regex that 99.99% works.

## Checklist

### DIFF
--- a/Sources/Extensions/SwiftStdlib/Deprecated/SwiftStdlibDeprecated.swift
+++ b/Sources/Extensions/SwiftStdlib/Deprecated/SwiftStdlibDeprecated.swift
@@ -39,6 +39,20 @@ public extension Bool {
 
 extension String {
 
+	#if canImport(Foundation)
+	/// SwifterSwift: Check if string is valid email format.
+	/// **Note that this property does not validate the email address against an email server. It merely attempts to determine whether its format is suitable for an email address.**.
+	///
+	///		"john@doe.com".isEmail -> true
+	///
+	@available(*, deprecated: 4.5.0, message: "Use isValidEmail instead", renamed: "isValidEmail")
+	public var isEmail: Bool {
+		// http://emailregex.com/
+		let regex = "^(?:[\\p{L}0-9!#$%\\&'*+/=?\\^_`{|}~-]+(?:\\.[\\p{L}0-9!#$%\\&'*+/=?\\^_`{|}~-]+)*|\"(?:[\\x01-\\x08\\x0b\\x0c\\x0e-\\x1f\\x21\\x23-\\x5b\\x5d-\\x7f]|\\\\[\\x01-\\x09\\x0b\\x0c\\x0e-\\x7f])*\")@(?:(?:[\\p{L}0-9](?:[a-z0-9-]*[\\p{L}0-9])?\\.)+[\\p{L}0-9](?:[\\p{L}0-9-]*[\\p{L}0-9])?|\\[(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?|[\\p{L}0-9-]*[\\p{L}0-9]:(?:[\\x01-\\x08\\x0b\\x0c\\x0e-\\x1f\\x21-\\x5a\\x53-\\x7f]|\\\\[\\x01-\\x09\\x0b\\x0c\\x0e-\\x7f])+)\\])$"
+		return range(of: regex, options: .regularExpression, range: nil, locale: nil) != nil
+	}
+	#endif
+
 	/// SwifterSwift: Number of characters in string.
 	///
 	///		"Hello world!".length -> 12
--- a/Sources/Extensions/SwiftStdlib/StringExtensions.swift
+++ b/Sources/Extensions/SwiftStdlib/StringExtensions.swift
@@ -157,13 +157,13 @@ public extension String {
 
 	#if canImport(Foundation)
 	/// SwifterSwift: Check if string is valid email format.
+	/// **Note that this property does not validate the email address against an email server. It merely attempts to determine whether its format is suitable for an email address.**.
 	///
 	///		"john@doe.com".isEmail -> true
 	///
-	public var isEmail: Bool {
-		guard self == trimmingCharacters(in: .whitespacesAndNewlines) else { return false }
+	public var isValidEmail: Bool {
 		// http://emailregex.com/
-		let regex = "(?:[\\p{L}0-9!#$%\\&'*+/=?\\^_`{|}~-]+(?:\\.[\\p{L}0-9!#$%\\&'*+/=?\\^_`{|}~-]+)*|\"(?:[\\x01-\\x08\\x0b\\x0c\\x0e-\\x1f\\x21\\x23-\\x5b\\x5d-\\x7f]|\\\\[\\x01-\\x09\\x0b\\x0c\\x0e-\\x7f])*\")@(?:(?:[\\p{L}0-9](?:[a-z0-9-]*[\\p{L}0-9])?\\.)+[\\p{L}0-9](?:[\\p{L}0-9-]*[\\p{L}0-9])?|\\[(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?|[\\p{L}0-9-]*[\\p{L}0-9]:(?:[\\x01-\\x08\\x0b\\x0c\\x0e-\\x1f\\x21-\\x5a\\x53-\\x7f]|\\\\[\\x01-\\x09\\x0b\\x0c\\x0e-\\x7f])+)\\])"
+		let regex = "^(?:[\\p{L}0-9!#$%\\&'*+/=?\\^_`{|}~-]+(?:\\.[\\p{L}0-9!#$%\\&'*+/=?\\^_`{|}~-]+)*|\"(?:[\\x01-\\x08\\x0b\\x0c\\x0e-\\x1f\\x21\\x23-\\x5b\\x5d-\\x7f]|\\\\[\\x01-\\x09\\x0b\\x0c\\x0e-\\x7f])*\")@(?:(?:[\\p{L}0-9](?:[a-z0-9-]*[\\p{L}0-9])?\\.)+[\\p{L}0-9](?:[\\p{L}0-9-]*[\\p{L}0-9])?|\\[(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?|[\\p{L}0-9-]*[\\p{L}0-9]:(?:[\\x01-\\x08\\x0b\\x0c\\x0e-\\x1f\\x21-\\x5a\\x53-\\x7f]|\\\\[\\x01-\\x09\\x0b\\x0c\\x0e-\\x7f])+)\\])$"
 		return range(of: regex, options: .regularExpression, range: nil, locale: nil) != nil
 	}
 	#endif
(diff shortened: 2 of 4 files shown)


Tests/SwiftStdlibTests/StringExtensionsTests.swift
```diff
@@ -77,7 +77,12 @@ final class StringExtensionsTests: XCTestCase {
     func testIsEmail() {
         XCTAssert("john@appleseed.com".isEmail)
         XCTAssertFalse("john@appleseedcom".isEmail)
-        XCTAssertFalse("johnappleseed.com".isEmail)
+		XCTAssertFalse("johnappleseed.com".isEmail)
```
guykogus: You're indenting with tabs here, while it looks like the rest uses spaces. Should we have a defined standard?

Sources/Extensions/SwiftStdlib/StringExtensions.swift
```diff
@@ -161,9 +161,10 @@ public extension String {
 	///		"john@doe.com".isEmail -> true
 	///
 	public var isEmail: Bool {
-		// http://stackoverflow.com/questions/25471114/how-to-validate-an-e-mail-address-in-swift
-		return range(of: "[A-Z0-9a-z._%+-]+@[A-Za-z0-9.-]+\\.[A-Za-z]{2,}",
-					 options: .regularExpression, range: nil, locale: nil) != nil
+		guard self == trimmingCharacters(in: .whitespacesAndNewlines) else { return false }
```
guykogus: Instead of doing this you should just add `^` and `$` to the beginning and end of the regex, respectively. Doing that to the original version of this function would have solved #514 to begin with.

Sources/Extensions/SwiftStdlib/StringExtensions.swift
```diff
@@ -161,9 +161,10 @@ public extension String {
...
+		// http://emailregex.com/
+		let regex = "(?:[\\p{L}0-9!#$%\\&'*+/=?\\^_`{|}~-]+(?:\\.[\\p{L}0-9!#$%\\&'*+/=?\\^_`{|}~-]+)*|\"(?:[\\x01-\\x08\\x0b\\x0c\\x0e-\\x1f\\x21\\x23-\\x5b\\x5d-\\x7f]|\\\\[\\x01-\\x09\\x0b\\x0c\\x0e-\\x7f])*\")@(?:(?:[\\p{L}0-9](?:[a-z0-9-]*[\\p{L}0-9])?\\.)+[\\p{L}0-9](?:[\\p{L}0-9-]*[\\p{L}0-9])?|\\[(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?|[\\p{L}0-9-]*[\\p{L}0-9]:(?:[\\x01-\\x08\\x0b\\x0c\\x0e-\\x1f\\x21-\\x5a\\x53-\\x7f]|\\\\[\\x01-\\x09\\x0b\\x0c\\x0e-\\x7f])+)\\])"
```
guykogus: I've read many posts about people saying that email validation is not something that should be done by regex. If you really want to ensure an email is valid you need to send it an email and see that you get a success response from the mail server.
Having such an extensive regex for emails is probably useful but it can potentially be misleading. People might look at this and think that it covers all scenarios and that there will never be false positives or negatives.
My suggestion, therefore, is that we add a note to the comment of the property to make it clear that this should only be used as a _suggestion_ as to whether or not the string is an email.
Personally I'd rather this property be named something like `looksLikeAnEmail` but I think that that would be pretty useless 😝

Tests/SwiftStdlibTests/StringExtensionsTests.swift
omaralbeik: 💯true, will update this

Sources/Extensions/SwiftStdlib/StringExtensions.swift
omaralbeik: Adding `^` and `$` solves the problem, however the old regex fails in examples like
`email.@.email.com` 🤔

Tests/SwiftStdlibTests/StringExtensionsTests.swift
```diff
@@ -77,7 +77,12 @@ final class StringExtensionsTests: XCTestCase {
     func testIsEmail() {
```
LucianoPAlmeida: @omaralbeik There are some valid/invalid email test cases [here](https://blogs.msdn.microsoft.com/testing123/2009/02/06/email-address-test-cases/) 
Maybe it's worth to consider adding some of them on the tests :))

Sources/Extensions/SwiftStdlib/StringExtensions.swift
guykogus: Apparently that's invalid, according to the list in @LucianoPAlmeida 's comment.

Sources/Extensions/SwiftStdlib/StringExtensions.swift
omaralbeik: - the purpose of this function is to serve as a local validation step for login and signup screens, to validate the address format before sending a request to server.
- I agree with you, we should add a note to inform user it is just checking the format of the string, and not checking the address itself, how about this one: **Note that this property does not perform authentication to validate the email address. It merely determines whether its format is valid for an email address.**
- About naming, I found [Micosoft](https://docs.microsoft.com/en-us/dotnet/standard/base-types/how-to-verify-that-strings-are-in-valid-email-format) is using the name `IsValidEmail`, maybe we can do so as well 😅

omaralbeik:
cc @SD10 ☝️

Sources/Extensions/SwiftStdlib/StringExtensions.swift
LucianoPAlmeida: @omaralbeik @guykogus 
Looks like that list is based [RFC 2822](https://tools.ietf.org/html/rfc2822) that is obsolete by [RFC 5322](https://tools.ietf.org/html/rfc5322#section-3.4.1), but anyways I believe this is an invalid email because  it seems like [RFC 5322](https://tools.ietf.org/html/rfc5322) explicitly only allows dots inside the local part of the mail address, not at the beginning and end.

Sources/Extensions/SwiftStdlib/StringExtensions.swift
guykogus: - Agreed, I've used this kind of validation before for past projects.
- The note wouldn't be necessarily correct. Even in your new test for `email.@.email.com` you get a false positive, because according to @LucianoPAlmeida 's link of invalid email formats the trailing `.` before the `@` isn't acceptable. That was the point I was trying to make before, which I read in a number of articles, where it's seemingly impossible to have a regex that will tell you for sure if an email address is valid or invalid. So what we're doing here is trying to _verify_, rather than _validate_, the email address. So maybe the note should be something like: **Note that this property does not validate the email address against an email server. It merely attempts to determine whether its format is suitable for an email address.**
- Let's keep the name just to avoid a breaking change 😬

LucianoPAlmeida: I agree too :))
I believe rename is important if it can improve expressivity and explain better the funcionality. Also we don't need to remove `isEmail`, we can just deprecate it and remove in a future version. And add the note on documentation as well.  I like names like you guys suggested `IsValidEmail` or maybe `matchesEmailPattern`. What do you think?

SD10: The purpose of this property is to determine if some text is an email, locally, before passing it on to the server for further validation. I'm fine with the name `isEmail` or `isValidEmail`. It's an extension on `String` so I believe it's self explanatory that a Swift primitive would not be performing validation server side

omaralbeik: I see, then I'm deprecating `isEmail` and renaming it to `isValidEmail` to have better consistency with other properties like `isValidURL`

@guykogus I also updated the note 👍

Tests/SwiftStdlibTests/StringExtensionsTests.swift
omaralbeik: Great, I added tests from links as well 💯

Tests/SwiftStdlibTests/StringExtensionsTests.swift
```diff
@@ -74,10 +74,37 @@ final class StringExtensionsTests: XCTestCase {
         XCTAssertFalse("abc".isAlphaNumeric)
     }
 
-    func testIsEmail() {
-        XCTAssert("john@appleseed.com".isEmail)
-        XCTAssertFalse("john@appleseedcom".isEmail)
-        XCTAssertFalse("johnappleseed.com".isEmail)
+    func testisValidEmail() {
+        // https://blogs.msdn.microsoft.com/testing123/2009/02/06/email-address-test-cases/
```
LucianoPAlmeida: @omaralbeik Maybe we should add the untrimmed email like "     email@mail.com  " .isValidEmail test case as well. What you think?

Sources/Extensions/SwiftStdlib/Deprecated/SwiftStdlibDeprecated.swift
```diff
@@ -39,6 +39,20 @@ public extension Bool {
...
 extension String {
 
+	#if canImport(Foundation)
+	/// SwifterSwift: Check if string is valid email format.
+	/// **Note that this property does not validate the email address against an email server. It merely attempts to determine whether its format is suitable for an email address.**.
+	///
+	///		"john@doe.com".isEmail -> true
+	///
+	@available(*, deprecated: 4.5.0, message: "Use isValidEmail instead", renamed: "isValidEmail")
+	public var isEmail: Bool {
```
guykogus: Why did you move this to a different extension?

Sources/Extensions/SwiftStdlib/StringExtensions.swift
```diff
@@ -157,13 +157,14 @@ public extension String {
 
 	#if canImport(Foundation)
 	/// SwifterSwift: Check if string is valid email format.
+	/// **Note that this property does not validate the email address against an email server. It merely attempts to determine whether its format is suitable for an email address.**.
```
guykogus: You can use the `Note` annotation instead. I.e.
```
///
/// - Note:
///  Note that this property does not validate the email address against an email server. It merely attempts to determine whether its format is suitable for an email address.
///
```
That way it'll appear in bold in the callout.

Sources/Extensions/SwiftStdlib/Deprecated/SwiftStdlibDeprecated.swift
SD10: I'm guessing to deprecate the old API

LucianoPAlmeida: @guykogus We move the deprecate extensions to those files to exclude them for coverage reports :))

SD10: Yeah it is an API breaking change that's not backwards compatible. We're not following semver properly here

guykogus: I really need to read more carefully the name of the file! 😅

omaralbeik: btw, I think since we used `renamed`, Xcode will suggest replacing the deprecated function with one click 😃

Issue #303, opened 2017-11-04
SpriteKit and SceneKit extensions


omaralbeik:
Think about extensions to help game developers using SpriteKit and SceneKit; examples:
- Bring `UIColor` extensions to `SKColor`
- Add methods to simplify animating `SKNode` objects
- and more ..

Unfortunately my experience with developing games is very limited, but I think adding such extensions to SwifterSwift would be a nice addition, especially that we're now using spspecs!

BennX:
Actually:
```Swift
typealias SKColor = UIColor
```

PR #522, opened 2018-07-20
Added more SpriteKit extensions

moyerr:
Helping out with Issue #303.

## Checklist

### DIFF
--- a/Sources/Extensions/SpriteKit/SKPhysicsBodyExtensions.swift
+++ b/Sources/Extensions/SpriteKit/SKPhysicsBodyExtensions.swift
@@ -15,6 +15,10 @@ public extension SKPhysicsBody {
     /// SwifterSwift: Checks if a physics body is contacting another body
     /// with the given `categoryBitMask`
     ///
+    ///     if player.physicsBody?.isTouching(Category.fire) {
+    ///         // Reduce player health
+    ///     }
+    ///
     /// - Parameter category: The `categoryBitmMask` to check
     /// - Returns: `true` if the physics body is contacting another body
     ///             with the given bitmask, `false` otherwise.
--- a/Sources/Extensions/SpriteKit/SKPhysicsBodyExtensions.swift
+++ b/Sources/Extensions/SpriteKit/SKPhysicsBodyExtensions.swift
@@ -23,7 +23,7 @@ public extension SKPhysicsBody {
     /// - Returns: `true` if the physics body is contacting another body
     ///             with the given bitmask, `false` otherwise.
     public func isTouching(_ category: UInt32) -> Bool {
-        return allContactedBodies().contains(where: { $0.categoryBitMask == category })
+        return allContactedBodies().contains(where: { $0.categoryBitMask & category != 0 })
     }
     
     /// SwifterSwift: Checks if a physics body is contacting another body.
--- a/Sources/Extensions/SpriteKit/SKPhysicsBodyExtensions.swift
+++ b/Sources/Extensions/SpriteKit/SKPhysicsBodyExtensions.swift
@@ -10,7 +10,6 @@ import SpriteKit
 
 public extension SKPhysicsBody {
     
-    
     /// SwifterSwift: Checks if a physics body is contacting another body
     /// with the given `categoryBitMask`
     ///
--- a/Sources/Extensions/SpriteKit/SKPhysicsBodyExtensions.swift
+++ b/Sources/Extensions/SpriteKit/SKPhysicsBodyExtensions.swift
@@ -19,7 +19,7 @@ public extension SKPhysicsBody {
     ///         // Reduce player health
     ///     }
     ///
-    /// - Parameter category: The `categoryBitmMask` to check
+    /// - Parameter category: The `categoryBitMask` to check
     /// - Returns: `true` if the physics body is contacting another body
     ///             with the given bitmask, `false` otherwise.
     public func isTouching(_ category: UInt32) -> Bool {
--- a/Sources/Extensions/SpriteKit/SKPhysicsBodyExtensions.swift
+++ b/Sources/Extensions/SpriteKit/SKPhysicsBodyExtensions.swift
@@ -9,6 +9,7 @@
 #if canImport(SpriteKit)
 import SpriteKit
 
+// MARK: - Methods
 public extension SKPhysicsBody {
     
     /// SwifterSwift: Checks if a physics body is contacting another body
--- a/Sources/Extensions/SpriteKit/SKPhysicsBodyExtensions.swift
+++ b/Sources/Extensions/SpriteKit/SKPhysicsBodyExtensions.swift
@@ -1,0 +1,23 @@
+//
+//  SKPhysicsBodyExtensions.swift
+//  SwifterSwift
+//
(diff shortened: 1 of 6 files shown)


Sources/Extensions/SpriteKit/SKPhysicsBodyExtensions.swift
```diff
@@ -0,0 +1,37 @@
...
+    ///
+    ///     if player.physicsBody?.isTouching(Category.fire) {
+    ///         // Reduce player health
+    ///     }
+    ///
+    /// - Parameter category: The `categoryBitmMask` to check
+    /// - Returns: `true` if the physics body is contacting another body
+    ///             with the given bitmask, `false` otherwise.
+    public func isTouching(_ category: UInt32) -> Bool {
+        return allContactedBodies().contains(where: { $0.categoryBitMask == category })
```
guykogus: This is incorrect. To check that a bitmask intersects you need use a logical AND:
`($0.categoryBitMask & category) != 0`.

Sources/Extensions/SpriteKit/SKPhysicsContactExtensions.swift
```diff
@@ -0,0 +1,43 @@
...
+#if canImport(SpriteKit)
+import SpriteKit
+
+// MARK: - Properties
+public extension SKPhysicsContact {
+    
+    /// SwifterSwift: A convenience variable for accessing the
+    /// nodes involved in a physical contact.
+    public var nodes: [SKNode] {
+        return [bodyA.node, bodyB.node].compactMap({ $0 })
```
guykogus: You can remove the parentheses.
```Swift
return [bodyA.node, bodyB.node].compactMap { $0 }
```

Sources/Extensions/SpriteKit/SKPhysicsContactExtensions.swift
```diff
@@ -0,0 +1,43 @@
...
+    ///
+    /// - Parameters:
+    ///     - categoryA: The first category to check
+    ///     - categoryB: The second category to check
+    ///
+    /// - Returns: `true` if the collision involves the two specific
+    ///     categories, `false` otherwise.
+    ///
+    public func involves(_ categoryA: UInt32, and categoryB: UInt32) -> Bool {
+        return (bodyA.categoryBitMask == categoryA && bodyB.categoryBitMask == categoryB) ||
```
guykogus: Again, you need to use logical AND.

Sources/Extensions/SpriteKit/SKPhysicsContactExtensions.swift
```diff
@@ -0,0 +1,43 @@
...
+    ///     }
+    ///
+    /// - Parameters:
+    ///     - categoryA: The first category to check
+    ///     - categoryB: The second category to check
+    ///
+    /// - Returns: `true` if the collision involves the two specific
+    ///     categories, `false` otherwise.
+    ///
+    public func involves(_ categoryA: UInt32, and categoryB: UInt32) -> Bool {
```
guykogus: I feel like `contains` would be a better name, so it's used quite often in Swift.

Sources/Extensions/SpriteKit/SKSpriteNodeExtensions.swift
```diff
@@ -0,0 +1,59 @@
...
+        timePerFrame seconds: TimeInterval,
+        resize: Bool,
+        restore: Bool,
+        completion: (() -> Void)? = nil) {
+        
+        let textureAtlas = SKTextureAtlas(named: atlasName)
+        
+        let textures = textureAtlas.textureNames
+            .sorted()
+            .map({ SKTexture(imageNamed: $0) })
```
guykogus: Again you can remove parentheses. I feel like we should be using SwiftLint to check for these :)
```Swift
.map { SKTexture(imageNamed: $0) }
```

Sources/Extensions/SpriteKit/SKPhysicsContactExtensions.swift
moyerr: This one I disagree with. `contains` is indeed used often in Swift, however, in this instance I think saying physical contact *involves* two object is more accurate than saying it *contains* two objects.

moyerr:
@guykogus Any ideas for how to unit test this `SpriteKit` stuff?

guykogus:
I guess you'll need to create a few `SKPhysicsBody` objects and place them in such a way that some should be touching while others are not and test your functions to see that they work.

As far as the animation function goes, maybe add an atlas with just a few frames (keep it as small as possible so as to not bloat the repo), run a test that runs the animation and then check it's state.
E.g.
```Swift
var isFinished = false
let timePerFrame = TimeInterval(1.0 / 30.0)
let spriteNode = SKSpriteNode()
spriteNode.animate(withAtlasNamed: "atlasName",
                   timePerFrame: timePerFrame,
                   resize: false,
                   restore: false,
                   completion: {
                    isFinished = true
})

XCTAssertFalse(isFinished)
Thread.sleep(forTimeInterval: timePerFrame)
XCTAssertFalse(isFinished)
Thread.sleep(forTimeInterval: timePerFrame)
XCTAssertTrue(isFinished)
```

Sources/Extensions/SpriteKit/SKPhysicsContactExtensions.swift
moyerr: @guykogus Would you accept the name as-is based on my previous comment?

moyerr:
@guykogus Any idea why those tests fail on macOS but pass on iOS and tvOS?

Sources/Extensions/SpriteKit/SKPhysicsContactExtensions.swift
guykogus: Yeah, I think it's fine. I think I gave it a little 👍 before ;)

guykogus:
No idea. The only thing I can think of is the timeout of the expectation being short, but that would surprise me. Have you tried running the macOS tests locally?

Sources/Extensions/SpriteKit/SKPhysicsContactExtensions.swift
moyerr: @guykogus Thanks! It's still showing the ❌  by changes requested, which looks like it'll block merging.

moyerr:
@guykogus Yep, it's saying:

> Asynchronous wait failed: Exceeded timeout

However, the expectation should be fulfilled immediately when the scene is presented, as it is in iOS and tvOS.

SD10:
So I have no experience with `SpriteKit`, I'm kind of useless on this one

guykogus:
It's been a long time since I used SpriteKit, and never used it on macOS, so can't really help you. Maybe try increase the timeout, but there may be something that it's expecting to happen...

moyerr:
@guykogus So i just came to the realization that the `SKSpriteNode` animation method I added is probably useless, as the texture assets would be in a different Bundle from SwifterSwift. I think I should probably remove that one. 

Still not sure about the macOS unit tests failing.

LucianoPAlmeida:
Closing this due inactivity.
Feel free to reopen it anytime :))

PR #523, opened 2018-07-20
UIView: add and remove multiple gesture recognizers at once with array of UIGestureRecognizer

moyerr:
## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 # Upcoming release
 
 ### Added
+- **UIView**
+  - Added `addGestureRecognizers(_:)` with variadic parameter to add multiple gesture recognizers to a view with one call. [#523](https://github.com/SwifterSwift/SwifterSwift/pull/523) by [moyerr](https://github.com/moyerr)
+  - Added `removeGestureRecognizers(_:)` with variadic parameter to remove multiple gesture recognizers from a view with one call. [#523](https://github.com/SwifterSwift/SwifterSwift/pull/523) by [moyerr](https://github.com/moyerr)
 
 ### Changed
 - **RangeReplaceableCollection**:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,8 +5,8 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 
 ### Added
 - **UIView**
-  - Added `addGestureRecognizers(_:)` with variadic parameter to add multiple gesture recognizers to a view with one call. [#523](https://github.com/SwifterSwift/SwifterSwift/pull/523) by [moyerr](https://github.com/moyerr)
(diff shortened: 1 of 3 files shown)


Sources/Extensions/UIKit/UIViewExtensions.swift
```diff
@@ -328,6 +328,30 @@ public extension UIView {
...
 	}
+    
+    /// SwifterSwift: Attaches zero or more gesture recognizers to the view.
+    ///
+    /// Attaching gesture recognizers to a view defines the scope of the represented
+    /// gesture, causing it to receive touches hit-tested to that view and all of its
+    /// subviews. The view establishes a strong reference to the gesture recognizers.
+    ///
+    /// - Parameter gestureRecognizers: The gesture recognizers to be added to the view.
+    public func addGestureRecognizers(_ gestureRecognizers: UIGestureRecognizer...) {
```
omaralbeik: Variadic parameters are very helpful, however, I think they should not be used here as they require 0 or more objects, which means the function might be used without passing any object, wich is kinda weird 😅 
```swift
addGestureRecognizers()
```
how about using an array instead?

moyerr: @omaralbeik Arrays can be empty too 😂

moyerr: @omaralbeik In my opinion, this points to a flaw at the call site, not with the API. The API will (correctly) do nothing if you pass 0 parameters. Exactly as it would if you went the Array route and passed an empty array. So it becomes a matter of style. Since variadic parameters are essentially a more readable way to pass an array, I would certainly prefer:

```
addGestureRecognizers(pan, tap)
```
over
```
addGestureRecognizers([pan, tap])
```

guykogus: I disagree about variadic parameters being more readable. They feel more like a relic of C. When you use an array for the parameters it makes it very clear that they're grouped together for a single purpose, whereas when they're variadic it may be less clear.
It's not a great example, but this can be very confusing:
```Swift
func foo(_ a: String, _ b: String, _ c: String...)
```
I agree with your idea that for this case the issue lies with the API and not the call site, but it can still lead to confusion.

Another (more important) point is that having an array as the parameter allows you to, obviously, pass an array to the function. So if a user has an array of gesture recognisers that they want to add/remove in one go while storing them, they couldn't use your function for it. You're limiting usage by making it variadic instead of an array.

moyerr: @guykogus Very well, I'll go ahead and change the functions to accept an array of gesture recognizers. Although, I suppose *both* implementations (variadic and array-based) could be added in the extension.

CHANGELOG.md
```diff
@@ -4,6 +4,9 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 # Upcoming release
 
 ### Added
+- **UIView**
+  - Added `addGestureRecognizers(_:)` with variadic parameter to add multiple gesture recognizers to a view with one call. [#523](https://github.com/SwifterSwift/SwifterSwift/pull/523) by [moyerr](https://github.com/moyerr)
```
guykogus: Please update these comments to replace mention of `variadic parameter`.

PR #525, opened 2018-07-24
UIRefreshControl extension added to begin refresh programatically.

ratulSharker:
## Checklist

- [x] I have added a [changelog](https://github.com/SwifterSwift/SwifterSwift/blob/master/CHANGELOG_GUIDELINES.md)

When a UIRefreshControl is added as a subview inside of UITableView, then calling the `beginRefresh` method does not show up the UIRefreshControl inside the UITableView. Developer had to manually change the content offset of the UITableView instance.

Source: [Stack Overflow](https://stackoverflow.com/questions/14718850/uirefreshcontrol-beginrefreshing-not-working-when-uitableviewcontroller-is-ins/14719658#14719658)

### DIFF
--- a/Sources/Extensions/UIKit/UIRefreshControlExtensions.swift
+++ b/Sources/Extensions/UIKit/UIRefreshControlExtensions.swift
@@ -16,11 +16,19 @@ public extension UIRefreshControl {
     /// SwifterSwift: Programatically begin refresh control inside of UITableView.
     ///
     /// - Parameter tableView: UITableView instance, inside which the refresh control is contained.
-    public func beginRefreshProgramatically(in tableView: UITableView) {
+    /// - Parameter animated: Boolean, indicates that is the content offset changing should be animated or not.
+    /// - Parameter sendAction: Boolean, indicates that should it fire sendActions method for valueChanged UIControlEvents 
+    public func beginRefreshing(in tableView: UITableView, animated: Bool, sendAction: Bool) {
         //https://stackoverflow.com/questions/14718850/uirefreshcontrol-beginrefreshing-not-working-when-uitableviewcontroller-is-ins/14719658#14719658
+        assert(superview == tableView, "Refresh control does not belong to the receiving table view")
+
         beginRefreshing()
         let offsetPoint = CGPoint(x: 0, y: -frame.height)
-        tableView.setContentOffset(offsetPoint, animated: true)
+        tableView.setContentOffset(offsetPoint, animated: animated)
+
+        if sendAction {
+            sendActions(for: .valueChanged)
+        }
     }
 
 }
--- a/Sources/Extensions/UIKit/UIRefreshControlExtensions.swift
+++ b/Sources/Extensions/UIKit/UIRefreshControlExtensions.swift
@@ -6,19 +6,19 @@
 //  Copyright © 2018 SwifterSwift
 //
 
-#if canImport(UIKit)
+#if os(iOS)
 import UIKit
 
-#if !os(watchOS)
 // MARK: - Methods
 public extension UIRefreshControl {
 
     /// SwifterSwift: Programatically begin refresh control inside of UITableView.
     ///
-    /// - Parameter tableView: UITableView instance, inside which the refresh control is contained.
-    /// - Parameter animated: Boolean, indicates that is the content offset changing should be animated or not.
-    /// - Parameter sendAction: Boolean, indicates that should it fire sendActions method for valueChanged UIControlEvents 
-    public func beginRefreshing(in tableView: UITableView, animated: Bool, sendAction: Bool) {
+    /// - Parameters:
+    ///   - tableView: UITableView instance, inside which the refresh control is contained.
+    ///   - animated: Boolean, indicates that is the content offset changing should be animated or not.
+    ///   - sendAction: Boolean, indicates that should it fire sendActions method for valueChanged UIControlEvents 
+    public func beginRefreshing(in tableView: UITableView, animated: Bool, sendAction: Bool = false) {
         //https://stackoverflow.com/questions/14718850/uirefreshcontrol-beginrefreshing-not-working-when-uitableviewcontroller-is-ins/14719658#14719658
         assert(superview == tableView, "Refresh control does not belong to the receiving table view")
 
@@ -33,5 +33,4 @@ public extension UIRefreshControl {
 
 }
 
-#endif
 #endif
--- a/Sources/Extensions/UIKit/UIRefreshControlExtensions.swift
+++ b/Sources/Extensions/UIKit/UIRefreshControlExtensions.swift
@@ -1,0 +1,29 @@
+//
+//  UIRefreshControlExtensions.swift
+//  SwifterSwift
+//
+//  Created by ratul sharker on 7/24/18.
+//  Copyright © 2018 SwifterSwift
+//
+
+#if canImport(UIKit)
+import UIKit
+
+#if !os(watchOS)
+// MARK: - Methods
+public extension UIRefreshControl {
+
+    /// SwifterSwift: Programatically begin refresh control inside of UITableView.
+    ///
+    /// - Parameter tableView: UITableView instance, inside which the refresh control is contained.
+    public func beginRefreshProgramatically(in tableView: UITableView) {
+        //https://stackoverflow.com/questions/14718850/uirefreshcontrol-beginrefreshing-not-working-when-uitableviewcontroller-is-ins/14719658#14719658
+        beginRefreshing()
+        let offsetPoint = CGPoint.init(x: 0, y: -frame.size.height)
+        tableView.setContentOffset(offsetPoint, animated: true)
+    }
+
+}
+
+#endif
+#endif
--- a/Sources/Extensions/UIKit/UIRefreshControlExtensions.swift
+++ b/Sources/Extensions/UIKit/UIRefreshControlExtensions.swift
@@ -19,7 +19,7 @@ public extension UIRefreshControl {
     public func beginRefreshProgramatically(in tableView: UITableView) {
         //https://stackoverflow.com/questions/14718850/uirefreshcontrol-beginrefreshing-not-working-when-uitableviewcontroller-is-ins/14719658#14719658
         beginRefreshing()
-        let offsetPoint = CGPoint.init(x: 0, y: -frame.size.height)
+        let offsetPoint = CGPoint(x: 0, y: -frame.height)
         tableView.setContentOffset(offsetPoint, animated: true)
     }
 
--- a/Tests/UIKitTests/UIRefreshControlExntesionsTests.swift
+++ b/Tests/UIKitTests/UIRefreshControlExntesionsTests.swift
@@ -1,0 +1,28 @@
+//
+//  UIRefreshControlExntesionsTests.swift
+//  SwifterSwift
+//
+//  Created by ratul sharker on 7/24/18.
+//  Copyright © 2018 SwifterSwift
+//
+
+#if os(iOS)
+
+import XCTest
+@testable import SwifterSwift
+
+final class UIRefreshControlExtensionTests: XCTestCase {
+
+    func testBeginRefreshProgramatically() {
+        let tableFrame = CGRect.init(x: 0, y: 0, width: 400, height: 400)
+        let tableView = UITableView.init(frame: tableFrame)
+        XCTAssertEqual(tableView.contentOffset, CGPoint.zero)
+        let refreshControl = UIRefreshControl.init()
+        tableView.addSubview(refreshControl)
+        refreshControl.beginRefreshProgramatically(in: tableView)
+        XCTAssertEqual(tableView.contentOffset.y, -refreshControl.frame.size.height)
+    }
+
+}
+
+#endif
--- a/Tests/UIKitTests/UIRefreshControlExntesionsTests.swift
+++ b/Tests/UIKitTests/UIRefreshControlExntesionsTests.swift
@@ -29,7 +29,7 @@ final class UIRefreshControlExtensionTests: XCTestCase {
             let tableview = UITableView()
             XCTAssertEqual(tableview.contentOffset, .zero)
             tableview.refreshControl = UIRefreshControl()
-            tableview.refreshControl?.beginRefreshing(in: tableview, animated: true)
+            tableview.refreshControl?.beginRefreshing(in: tableview, animated: true, sendAction: true)
 
             XCTAssertTrue(tableview.refreshControl!.isRefreshing)
             XCTAssertEqual(tableview.contentOffset.y, -tableview.refreshControl!.frame.height)
--- a/Tests/UIKitTests/UIRefreshControlExntesionsTests.swift
+++ b/Tests/UIKitTests/UIRefreshControlExntesionsTests.swift
@@ -14,13 +14,14 @@ import XCTest
 final class UIRefreshControlExtensionTests: XCTestCase {
 
     func testBeginRefreshProgramatically() {
-        let tableFrame = CGRect.init(x: 0, y: 0, width: 400, height: 400)
+        let tableFrame = CGRect(x: 0, y: 0, width: 400, height: 400)
         let tableView = UITableView.init(frame: tableFrame)
-        XCTAssertEqual(tableView.contentOffset, CGPoint.zero)
+        XCTAssertEqual(tableView.contentOffset, .zero)
         let refreshControl = UIRefreshControl.init()
         tableView.addSubview(refreshControl)
         refreshControl.beginRefreshProgramatically(in: tableView)
-        XCTAssertEqual(tableView.contentOffset.y, -refreshControl.frame.size.height)
+        XCTAssertTrue(refreshControl.isRefreshing)
+        XCTAssertEqual(tableView.contentOffset.y, -refreshControl.frame.height)
     }
 
 }
--- a/Tests/UIKitTests/UIRefreshControlExntesionsTests.swift
+++ b/Tests/UIKitTests/UIRefreshControlExntesionsTests.swift
@@ -22,17 +22,15 @@ final class UIRefreshControlExtensionTests: XCTestCase {
 
         XCTAssertTrue(refreshControl.isRefreshing)
         XCTAssertEqual(tableView.contentOffset.y, -refreshControl.frame.height)
-    }
 
-    func testBeginRefreshAsRefreshControlProperty() {
         if #available(iOS 10.0, *) {
-            let tableview = UITableView()
-            XCTAssertEqual(tableview.contentOffset, .zero)
-            tableview.refreshControl = UIRefreshControl()
-            tableview.refreshControl?.beginRefreshing(in: tableview, animated: true, sendAction: true)
+            let anotherTableview = UITableView()
+            XCTAssertEqual(anotherTableview.contentOffset, .zero)
+            anotherTableview.refreshControl = UIRefreshControl()
+            anotherTableview.refreshControl?.beginRefreshing(in: anotherTableview, animated: true, sendAction: true)
 
-            XCTAssertTrue(tableview.refreshControl!.isRefreshing)
-            XCTAssertEqual(tableview.contentOffset.y, -tableview.refreshControl!.frame.height)
+            XCTAssertTrue(anotherTableview.refreshControl!.isRefreshing)
+            XCTAssertEqual(anotherTableview.contentOffset.y, -anotherTableview.refreshControl!.frame.height)
         }
     }
 
--- a/Tests/UIKitTests/UIRefreshControlExntesionsTests.swift
+++ b/Tests/UIKitTests/UIRefreshControlExntesionsTests.swift
@@ -14,11 +14,11 @@ import XCTest
 final class UIRefreshControlExtensionTests: XCTestCase {
 
     func testBeginRefreshAsRefreshControlSubview() {
-        let tableView = UITableView.init()
+        let tableView = UITableView()
         XCTAssertEqual(tableView.contentOffset, .zero)
-        let refreshControl = UIRefreshControl.init()
+        let refreshControl = UIRefreshControl()
         tableView.addSubview(refreshControl)
-        refreshControl.beginRefreshing(in: tableView, animated: true, sendAction: false)
+        refreshControl.beginRefreshing(in: tableView, animated: true)
 
         XCTAssertTrue(refreshControl.isRefreshing)
         XCTAssertEqual(tableView.contentOffset.y, -refreshControl.frame.height)
@@ -26,10 +26,10 @@ final class UIRefreshControlExtensionTests: XCTestCase {
 
     func testBeginRefreshAsRefreshControlProperty() {
         if #available(iOS 10.0, *) {
-            let tableview = UITableView.init()
+            let tableview = UITableView()
             XCTAssertEqual(tableview.contentOffset, .zero)
-            tableview.refreshControl = UIRefreshControl.init()
-            tableview.refreshControl?.beginRefreshing(in: tableview, animated: true, sendAction: false)
+            tableview.refreshControl = UIRefreshControl()
+            tableview.refreshControl?.beginRefreshing(in: tableview, animated: true)
 
             XCTAssertTrue(tableview.refreshControl!.isRefreshing)
             XCTAssertEqual(tableview.contentOffset.y, -tableview.refreshControl!.frame.height)
--- a/Tests/UIKitTests/UIRefreshControlExntesionsTests.swift
+++ b/Tests/UIKitTests/UIRefreshControlExntesionsTests.swift
@@ -13,15 +13,27 @@ import XCTest
 
 final class UIRefreshControlExtensionTests: XCTestCase {
 
-    func testBeginRefreshProgramatically() {
-        let tableFrame = CGRect(x: 0, y: 0, width: 400, height: 400)
-        let tableView = UITableView.init(frame: tableFrame)
+    func testBeginRefreshAsRefreshControlSubview() {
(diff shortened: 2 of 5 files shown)


Sources/Extensions/UIKit/UIRefreshControlExtensions.swift
```diff
@@ -0,0 +1,29 @@
...
+import UIKit
+
+#if !os(watchOS)
+// MARK: - Methods
+public extension UIRefreshControl {
+
+    /// SwifterSwift: Programatically begin refresh control inside of UITableView.
+    ///
+    /// - Parameter tableView: UITableView instance, inside which the refresh control is contained.
+    public func beginRefreshProgramatically(in tableView: UITableView) {
```
guykogus: I feel like this is backwards and that the function should be a `UITableView` extension. I.e.
```Swift
extension UITableView {
    @available(iOS 10.0, *)
    public func refresh(animated: Bool = true) {
        guard let refreshControl = refreshControl else { return }

        // https://stackoverflow.com/questions/14718850/uirefreshcontrol-beginrefreshing-not-working-when-uitableviewcontroller-is-ins/14719658#14719658
        refreshControl.beginRefreshing()
        let offsetPoint = CGPoint(x: 0, y: -refreshControl.frame.height)
        setContentOffset(offsetPoint, animated: animated)
    }
}
```
You could potentially add this as an extension of `UIScrollView` instead, which is a superclass of `UITableView` that defines the property, but it doesn't feel right because the refresh of a scroll view may not necessarily be along the top like it is for table views.

You could, as an addition, add it to `UITableViewController` as well:
```Swift
extension UITableViewController {
    public func refresh(animated: Bool = true) {
        guard let refreshControl = refreshControl else { return }

        // https://stackoverflow.com/questions/14718850/uirefreshcontrol-beginrefreshing-not-working-when-uitableviewcontroller-is-ins/14719658#14719658
        refreshControl.beginRefreshing()
        let offsetPoint = CGPoint(x: 0, y: -refreshControl.frame.height)
        tableView.setContentOffset(offsetPoint, animated: animated)
    }
}
```

Notice that for `UITableViewController` you don't need `@available` because it's been available since iOS 6.

Sources/Extensions/UIKit/UIRefreshControlExtensions.swift
```diff
@@ -0,0 +1,29 @@
...
+        //https://stackoverflow.com/questions/14718850/uirefreshcontrol-beginrefreshing-not-working-when-uitableviewcontroller-is-ins/14719658#14719658
+        beginRefreshing()
+        let offsetPoint = CGPoint.init(x: 0, y: -frame.size.height)
```
guykogus: If you noticed in my examples above, you can shorten this by removing `.init` and `.size`.

Tests/UIKitTests/UIRefreshControlExntesionsTests.swift
```diff
@@ -0,0 +1,28 @@
...
+
+#if os(iOS)
+
+import XCTest
+@testable import SwifterSwift
+
+final class UIRefreshControlExtensionTests: XCTestCase {
+
+    func testBeginRefreshProgramatically() {
+        let tableFrame = CGRect.init(x: 0, y: 0, width: 400, height: 400)
```
guykogus: Again in here you can shorten this and other similar calls by removing `.init`. It's superfluous.

Tests/UIKitTests/UIRefreshControlExntesionsTests.swift
```diff
@@ -0,0 +1,28 @@
...
+        let tableView = UITableView.init(frame: tableFrame)
+        XCTAssertEqual(tableView.contentOffset, CGPoint.zero)
```
guykogus: Another nice cleanup you can do here is just write `XCTAssertEqual(tableView.contentOffset, .zero)`. Swift _should_ be able to understand that in this context `.zero` is a reference to `CGPoint.zero`, since you're trying to call a generic function with a `CGPoint` for its first parameter.

Tests/UIKitTests/UIRefreshControlExntesionsTests.swift
```diff
@@ -0,0 +1,28 @@
...
+        let refreshControl = UIRefreshControl.init()
+        tableView.addSubview(refreshControl)
+        refreshControl.beginRefreshProgramatically(in: tableView)
+        XCTAssertEqual(tableView.contentOffset.y, -refreshControl.frame.size.height)
```
guykogus: Maybe also add after this `XCTAssert(refreshControl.isRefreshing)`

Sources/Extensions/UIKit/UIRefreshControlExtensions.swift
ratulSharker: I also had the similar feeling that the functionality should go with `UITableView` or `UIScrollView`. But in OOP way, which widget should get this functionality ?. This functionality belongs to `UIRefreshControl`. From developer (who are using SwifterSwift) perspective, it's more obvious to find this functionality in `UIRefreshControl`, which S/he attached to her/his `UITableView`/`UITableViewController`. 

Similarly this could also be a part of the `UITableViewController`, but in that way developer is expecting the functionality in the `UIRefreshControl`. If you check carefully i also kept the name similarity (original : `beginRefreshing`, extension: `beginRefreshProgramatically`). It happens to me every time typing the `tableView.reloadData()`, learning about the `tableView.reloadData { /*completion*/ }`, then using this functionality.

Another reason i didn't submit it as a `UITableView` extension is that, this would be only available for iOS 10+. To support iOS 8+ this extension should go with `UIRefreshControl`. Putting this functionality in `UITableViewController` won't completely solve the problem, to support `UITableView` scenario, the extension needed to be in `UITableView` or `UIRefreshControl`. If we choose `UITableView`, then the cases below iOS 10 require the same extension copied into `UIRefreshControl`. More over swifter swift is supporting iOS 8+.

Considering these cases, i ended up putting this extension into `UIRefreshControl` which could be used in all these scenarios.

Tests/UIKitTests/UIRefreshControlExntesionsTests.swift
ratulSharker: When i wrote the extension i sum up the test cases, but forget to put it up :)

Sources/Extensions/UIKit/UIRefreshControlExtensions.swift
ratulSharker: Before seeing this comment, i didn't know that the `CGRect` has a computed property name `height`. Today i learned it from your comment. Thanks a lot @guykogus 

I also minimized one test case as follows:
```
XCTAssertEqual(tableView.contentOffset.y, -refreshControl.frame.height)
```

Sources/Extensions/UIKit/UIRefreshControlExtensions.swift
guykogus: The problem is that you could be calling refresh for the wrong table view. E.g.
```Swift
let refreshControl1 = UIRefreshControl()
let refreshControl2 = UIRefreshControl()
let tableView = ...
tableView.refreshControl = refreshControl1
refreshControl2.beginRefreshProgramatically(in: tableView)
```
I know it's not likely to happen, you'd have to really mess it up, but it opens the door for a mistake like this.

I don't see why the functionality would belong to the refresh control instead of the table view. Since the function is only useful if the RC belongs to a TV, and since a RC can belong to anything while a TV can only have 1 RC, it seems to make a lot more sense for the function to be in the TV.

If you want to add pre-iOS 10 support it's not that hard.
```Swift
extension UITableView {
    public func refresh(animated: Bool = true) {
        let refreshControl: UIRefreshControl!
        if #available(iOS 10.0, *) {
            refreshControl = self.refreshControl
        } else {
            refreshControl = subviews.first(where: { $0 is UIRefreshControl }) as? UIRefreshControl
        }

        guard refreshControl != nil else { return }

        // https://stackoverflow.com/questions/14718850/uirefreshcontrol-beginrefreshing-not-working-when-uitableviewcontroller-is-ins/14719658#14719658
        refreshControl.beginRefreshing()
        let offsetPoint = CGPoint(x: 0, y: -refreshControl.frame.height)
        setContentOffset(offsetPoint, animated: animated)
    }
}
```

Sources/Extensions/UIKit/UIRefreshControlExtensions.swift
```diff
@@ -0,0 +1,29 @@
...
+public extension UIRefreshControl {
+
+    /// SwifterSwift: Programatically begin refresh control inside of UITableView.
+    ///
+    /// - Parameter tableView: UITableView instance, inside which the refresh control is contained.
+    public func beginRefreshProgramatically(in tableView: UITableView) {
+        //https://stackoverflow.com/questions/14718850/uirefreshcontrol-beginrefreshing-not-working-when-uitableviewcontroller-is-ins/14719658#14719658
+        beginRefreshing()
+        let offsetPoint = CGPoint(x: 0, y: -frame.height)
+        tableView.setContentOffset(offsetPoint, animated: true)
```
guykogus: This isn't enough to actually get it to work. To register for when a pull-to-refresh action has been called on a `UIRefreshControl` you need to add a target-action for `.valueChanged`. You can instantly call
`sendActions(for: .valueChanged)`
but then you won't see the animation for showing the refresh control. Taking into account this, and the comments mentioned above, I think that the final version should be:

```Swift
extension UITableView {
    public func refresh(animated: Bool = true) {
        let refreshControl: UIRefreshControl!
        if #available(iOS 10.0, *) {
            refreshControl = self.refreshControl
        } else {
            refreshControl = subviews.first(where: { $0 is UIRefreshControl }) as? UIRefreshControl
        }

        guard refreshControl != nil else { return }

        // https://stackoverflow.com/questions/14718850/uirefreshcontrol-beginrefreshing-not-working-when-uitableviewcontroller-is-ins/14719658#14719658
        refreshControl.beginRefreshing()
        let offsetPoint = CGPoint(x: 0, y: -refreshControl.frame.height)
        setContentOffset(offsetPoint, animated: animated)

        DispatchQueue.main.asyncAfter(deadline: .now() + (animated ? 0.3 : 0.0)) { [weak refreshControl] in
            refreshControl?.sendActions(for: .valueChanged)
        }
    }
}
```
I created a test project and it looks like `0.3` is the time used for `setContentOffset(_:, animated:)`. It may not be ideal to use it since it can change in future versions of iOS, but Apple doesn't provide us the constant (that I'm aware of) and we can't use the delegate callback `scrollViewDidEndScrollingAnimation(_:)` here.

Please also note the function name in my version. Maybe you can come up with a better one, but `beginRefreshProgramatically` is both a mouthful and it's not very "clean" to add `programmatically` (there are 2 m's, not 1) to the name of a programming function.

guykogus: If you want to avoid calling `sendActions` after the delay, the only alternative I can think of is to add a big note in the comments that the caller needs to ensure that the delegate of the table view calls the target-action callback manually once `scrollViewDidEndScrollingAnimation(_:)` was called, but that's very clunky.

ratulSharker: <h2>1. Calling with wrong `UITableView` as parameter</h2>
This could be easily checked is the developer called with wrong TV. RC is added as a subview of the TV. Below iOS 10 developer deliberately add RC as subview of TV, so it is obvious. For iOS 10+ i have tested as follows

```
func testExample() {
    let tableView = UITableView.init()
    tableView.refreshControl = UIRefreshControl.init()
    XCTAssertEqual(tableView, tableView.refreshControl?.superview)
}
```

The test passes. So if we put a checking in the extension like following

```
public func beginRefreshProgramatically(in tableView: UITableView) {
       if superview == tableView {
            beginRefreshing()
            let offsetPoint = CGPoint(x: 0, y: -frame.height)
            tableView.setContentOffset(offsetPoint, animated: true)
        } else {
            print("Refresh Control does not belong to this table view")
            // we can also put assertFailure here, but i'm unsure about it
        }
    }
```
The problem of setting content offset of the wrong TV can be avoided like this.

<h2>2. Supporting Pre iOS 10 Solution</h2>
In iOS 10+ RC can be added as subview in TV, and they just work fine just like pre iOS 10. The developer could've coded like this even after his app targeting iOS 10+. Adding the RC as subview won't set the TV's `refreshControl` property. i.e

```
func testTableViewAddSubView() {
        let tableView = UITableView.init()
        let refreshControl = UIRefreshControl.init()
        tableView.addSubview(refreshControl)
        XCTAssertNotEqual(tableView.refreshControl, refreshControl)
    }
```
In this case the solution won't work because the checking is put like following

```
if #available(iOS 10.0, *) {
    refreshControl = self.refreshControl
}
```
which won't get the RC object and the RC won't refresh anything.

More worse case could be like following. Think of an application which support pre iOS 10 using the solution you provided and the application is running on a device with iOS 10+. As the application supports pre iOS 10 then RC is added as subview, not set in the property. As your proposed solution the code should reach Line 2 of the following block.

```
1. if #available(iOS 10.0, *) {
2.    refreshControl = self.refreshControl
3. }
```
As the RC is added as subview in TV it won't get the `refreshControl` object, resulting in setting the nil into the `refreshControl` and ended up crashing while calling `beginRefreshing()` method or accessing the `frame` property.

<h2>3. Why the extension still in UIRefreshControl class</h2>
UIRefreshControl widget is responsible for showing refreshing animation, not the UITableView. It is more likely that, developer who using SwifterSwift will find refresh method in UIRefreshControl object. <b>Even after adding the `refreshControl` property in iOS 10+ you have to call `beginRefreshing` method on the RC not in the TV</b> . This is more developer friendly to put this extension into the RC instead of TV. You could not assume that the developer will read the SwifterSwift documentation and find this magic inside TV. Even i don't read the SwifterSwift documentation because i don't have to. All the extension i guessed, found them in the right place.

<h2>4. To see the animation the delay of 0.3</h2>
As you said apple does not provide any constant of the delays, i strongly disagree to add this portion in the extension. A try to move this extension from RC to TV is the obvious reason for this. This is another reason i don't want this extension in RC. I didn't think so far, but thanks for your finding. You may have ran your test project in few devices. Have you ran them on all devices with different iOS version running on them ? In past i had very difficult time for taking such decision, like introduce a constant delay to sync with some tasks. If you like you can lookup `DEFAULT_KEYBOARD_UP_ANIMATION_DURATION` 
<a href="https://github.com/ratulSharker/Keyboard-Animator/blob/master/Example%20Project/Testing/Testing/KeyboardAnimator/KeyboardAnimator.h">here</a> . For each and every iOS version i ended up tweaking this constant value. I could find a value which seems perfect for all iOS and devices. Later i came to know the key `UIKeyboardAnimationDurationUserInfoKey` which solves all the problem. The alternative to using constant duration very unfriendly api (letting the developer call `scrollViewDidEndScrollingAnimation` manually through a delegate ) and the delegate you are asking as a alternative is too much to use this functionality. It just an overkill.

<h2>5. Naming convention</h2>
Yes, in the name i missed two `m`. My mistake. The name may sound too mouthful, but i wanted to keep the `beingRefresh` as a prefix. The reason is while the developer will begin typing like

```
refreshControl.beg
```
I want xcode suggest the native `beginRefresh` and the extension side by side. Please let me know if you can suggest something less mouthful and cleaner which is prefixed with `beginRefresh` or i should go with `beginRefreshProgramatically` / `beginRefreshProgrammatically`.

Let me know your suggestion regarding these 5 statements.

guykogus: 1. I agree that we can easily create an assert for it.
2. You're right, the user might have added the RC as a subview so my check would have failed (it wouldn't have crashed because of the `guard`, though). That's easily fixed.
3. The user will be calling the function the same way. They won't need to call `beginRefreshing` themselves if they use the TV extension, but they do still need that target-action callback that calls `refreshControl.endRefreshing`, so I see your point that it makes sense to keep them together.
4. Yeah, the animation timing is definitely a bad idea. If the reloading of the table view is instantaneous, which in real world apps is unlikely, then they just won't see the refresh control, it'll scroll back up to the top and the user can keep using the table view instantly.
5. I think it's fine to just use `beginRefreshing` for the name. The user will see the difference between the original and the extension because of the parameters.

Based on your points, I suggest this extension:
```Swift
extension UIRefreshControl {
    public func beginRefreshing(in tableView: UITableView, animated: Bool = true) {
        let isSubView = {
            tableView.subviews.contains(where: { $0 == self })
        }
        if #available(iOS 10.0, *) {
            assert(isSubView() || tableView.refreshControl == self, "Refresh control does not belong to the receiving table view")
        } else {
            assert(isSubView(), "Refresh control does not belong to the receiving table view")
        }

        beginRefreshing()
        let offsetPoint = CGPoint(x: 0, y: -frame.height)
        tableView.setContentOffset(offsetPoint, animated: animated)

        sendActions(for: .valueChanged)
    }
}
```
`isSubview` is a closure because on non-debug builds (where `assert` isn't executed) we don't want to actually perform the calculation. Chances are that the number of subviews will be small, but why risk a potentially heavy calculation?

It works perfectly in my test VC:

```Swift
class ViewController: UIViewController, UITableViewDataSource {
    @IBOutlet weak var tableView: UITableView!

    var refreshControl: UIRefreshControl!

    override func viewDidLoad() {
        super.viewDidLoad()

        refreshControl = UIRefreshControl(frame: CGRect(origin: .zero, size: CGSize(width: view.frame.width, height: 44)))
        refreshControl.addTarget(self, action: #selector(refresh(_:)), for: [.valueChanged])
        if #available(iOS 10.0, *) {
            tableView.refreshControl = refreshControl
        } else {
            tableView.addSubview(refreshControl)
        }
    }

    override func viewDidAppear(_ animated: Bool) {
        super.viewDidAppear(animated)
        refreshControl.beginRefreshing(in: tableView)
    }

    @objc func refresh(_ sender: UIRefreshControl) {
        // Wait to show the RC
        DispatchQueue.main.asyncAfter(deadline: .now() + 0.5) { [weak self] in
            self?.refreshControl.endRefreshing()
        }
    }

    func tableView(_ tableView: UITableView, numberOfRowsInSection section: Int) -> Int {
        return 500
    }

    func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
        let cell = tableView.dequeueReusableCell(withIdentifier: "Cell", for: indexPath)
        cell.textLabel?.text = indexPath.description
        return cell
    }
}
```

Thoughts?

ratulSharker: In the suggested extension 

```
let isSubView = {
            tableView.subviews.contains(where: { $0 == self })
        }
        if #available(iOS 10.0, *) {
            assert(isSubView() || tableView.refreshControl == self, "Refresh control does not belong to the receiving table view")
        } else {
            assert(isSubView(), "Refresh control does not belong to the receiving table view")
        }
```

It finds the RC from the TV. Why not just check the `superview` of the RC ?

```
if superview == tableView {
        // valid
} else {
       // invalid -- assert here
}
```

It also survives the iOS 10+, as `refreshControl` property in TV is added as subview. i.e 

```
func testIsRefreshControlAddedAsSubview(){
        let tableView = UITableView.init()
        tableView.refreshControl = UIRefreshControl.init()
        XCTAssertEqual(tableView, tableView.refreshControl?.superview)
    }
```

which is passed.

This superview checking version is technically more efficient, manageable and most of all it supports both pre and post iOS 10.

I really appreciate the animation parameter. 👍 

But the `sendActions(for: .valueChanged)` i totally disagree with. As you can see, while changing state of the `UISwitch`, `UISegmentController`, `UIProgressView` and others, none of them call the added target while developer change the state of the widget. 

Let me clarify my point, say you added a target and selector to the `UISegmentSelector`. While the user taps the segment controller and select the different segment the selector of that target is getting called. But while developer programmatically set or change the selected index, the selector in the target not getting called. 

This is more likely that developer will expect the selector in the target won't get called. Most of the time developer want to change the state because the job required to do while the state changes are already done, thats why changing the state deliberately by the system. So changing the state does not mean to fire up the selector attached to it. But we can keep it with another parameter deciding to fire it up or not.

So summarizing everything, Here is a version that i suggest to keep.

```
public func beginRefreshing(in tableView: UITableView, animated: Bool, sendAction: Bool) {
        //https://stackoverflow.com/questions/14718850/uirefreshcontrol-beginrefreshing-not-working-when-uitableviewcontroller-is-ins/14719658#14719658
        assert(superview == tableView, "Refresh control does not belong to the receiving table view")
        
        beginRefreshing()
        let offsetPoint = CGPoint(x: 0, y: -frame.height)
        tableView.setContentOffset(offsetPoint, animated: animated)

        if sendAction {
            sendActions(for: .valueChanged)
        }
    }
```

Please share your thought about it.

Sources/Extensions/UIKit/UIRefreshControlExtensions.swift
SD10: 😱 So much going on here

Sources/Extensions/UIKit/UIRefreshControlExtensions.swift
guykogus: I totally agree with everything.
I was sure that I tested the `superview` value for iOS 10 and it came back as `false`, which is why I did both checks. I tested it again and you're right, it's `true`.
I don't even think we need the `sendAction` parameter. As you said, it would make it more in line with other `UIControl`s. Also, if the developer is already calling this method they know that they need to perform the reloading, they don't need to wait for the target-action callback. This makes a lot of sense.
Thanks for putting up with my debate :)
💯

ratulSharker:
While pushing the last changes, github was showing that If i tried to merge into master there is a merge conflict in `CHANGELOG.md` file. Then i resolved the conflict and committed the changes. After committing the changes Codecov reported that 

### **Merging #525 into master will decrease coverage by 0.04%.**

What was my fault, please mention how could i resolve it.

T.I.A

Tests/UIKitTests/UIRefreshControlExntesionsTests.swift
```diff
@@ -0,0 +1,41 @@
...
+//
+
+#if os(iOS)
+
+import XCTest
+@testable import SwifterSwift
+
+final class UIRefreshControlExtensionTests: XCTestCase {
+
+    func testBeginRefreshAsRefreshControlSubview() {
```
LucianoPAlmeida: I believe that in the [Contributing Guidelines](https://github.com/SwifterSwift/SwifterSwift/blob/master/CONTRIBUTING.md#adding-tests) we have a one test function per extension guide covering all the test cases. Also, can you add a test case where you pass send action as `true` to test the behavior on that and maintain the patch diff coverage?

ratulSharker: I tried to add a function which test the `sendAction` in the following way,

```swift
private var refreshControlValueChangedFired: Bool = false
private var refreshControlValueChangedExpectation: XCTestExpectation!

func testBeginRefreshSendAction() {
    let tableView = UITableView.init()
    let refreshControl = UIRefreshControl()
    refreshControl.addTarget(self, action: #selector(refreshControlValueChanged), for: .valueChanged)
    tableView.addSubview(refreshControl)
    refreshControl.beginRefreshing(in: tableView, animated: true, sendAction: true)
    refreshControlValueChangedExpectation = XCTestExpectation(description: "Refresh Control .valueChanged will fire")
    wait(for: [refreshControlValueChangedExpectation], timeout: 10)
    XCTAssertTrue(refreshControlValueChangedFired)
}

@objc func refreshControlValueChanged() {
    print("this function is called")
    refreshControlValueChangedFired = true
   refreshControlValueChangedExpectation.fulfill()
}
```

But my finding are `refreshControlValueChanged` are never fired. Then i came to know that <a href="https://stackoverflow.com/questions/39851205/how-to-test-if-a-uicontrolevents-has-been-fired">without host application `sendAction` won't work</a>. As it is a library only, no Host application within the SwifterSwift, I believe there is no straight forward way to test the `sendAction` is fired within a library. Please suggest me any other ways to do this.

LucianoPAlmeida: Hey @ratulSharker 
I see the issue here, but actually you don't need to assert if the the sendAction() method triggered the .valueChanged because, is an UIKit API and Apple for sure already tested this. What means you have the garante that if you call that, it will trigger the event, so there's no need to test it. What I mean by adding a test case was just in one of those cases that you are already testing, to put the parameter true just for it to pass there and count as coverage so it increase the path coverage, eventhough you are testing the other behaviors :))
Also can you move the to behaviors tests to only one test method, to follow the [Contributing Guidelines](https://github.com/SwifterSwift/SwifterSwift/blob/master/CONTRIBUTING.md#adding-tests) one test method to extension?

ratulSharker: Thanks @LucianoPAlmeida 
It increases the path coverage.

Tests/UIKitTests/UIRefreshControlExntesionsTests.swift
```diff
@@ -0,0 +1,41 @@
...
+        let refreshControl = UIRefreshControl()
+        tableView.addSubview(refreshControl)
+        refreshControl.beginRefreshing(in: tableView, animated: true)
+
+        XCTAssertTrue(refreshControl.isRefreshing)
+        XCTAssertEqual(tableView.contentOffset.y, -refreshControl.frame.height)
+    }
+
+    func testBeginRefreshAsRefreshControlProperty() {
+        if #available(iOS 10.0, *) {
```
LucianoPAlmeida: @ratulSharker Can you put this test case inside the same method than the other to follow the [Contributing Guidelines](https://github.com/SwifterSwift/SwifterSwift/blob/master/CONTRIBUTING.md#adding-tests) one to one relation test method per extension?

PR #526, opened 2018-07-25
set translatesAutoresizingMaskIntoConstraints when adding subview

overovermind:
Added a general helper for adding subviews when using constraints.

Before:
```
let subview = UIView()
subview.translatesAutoresizingMaskIntoConstraints = false
view.addSubview(subview)
```

Now:
```
let subview = UIView()
view.addSubview(subview, translatesAutoresizingMaskIntoConstraints: false)
```

Also possibile with array of views. Would love to see this added 😀

## Checklist

- [x] I checked the [**Contributing Guidelines**](https://github.com/SwifterSwift/SwifterSwift/blob/master/CONTRIBUTING.md) before creating this request.
- [x ] New extensions are written in Swift 4.
- [x] New extensions support iOS 8.0+ / tvOS 9.0+ / macOS 10.10+ / watchOS 2.0+.
- [x] I have added tests for new extensions, and they passed.
- [x] All extensions have a **clear** comments explaining their functionality, all parameters and return type in English.
- [x] All extensions are declared as **public**.
- [x] I have added a [changelog](https://github.com/SwifterSwift/SwifterSwift/blob/master/CHANGELOG_GUIDELINES.md) entry describing my changes.

PR #529, opened 2018-07-28
Add UIView.add method

sammy-SC:
## Checklist

I'm happy to make adjustments if required.


guykogus:
Thank you for taking the time to submit this PR. There is a very similar one here:
https://github.com/SwifterSwift/SwifterSwift/pull/526
which is trying to do the same thing. It is being rejected because it doesn't seem to be a good fit for SwifterSwift. Please refer to that PR for more information.

sammy-SC:
@guykogus thank you for the response.

On similar matter, how do you feel about making `addSubviews(_ subviews: [UIView])` into a method with `UIView` as a variadic parameter? It would provide cleaner interface.

guykogus:
Somebody recently added `addSubviews` and `removeSubviews` and wanted to do the same thing, but we convinced him to change it to use an array. The 2 main points are:
- With variadic parameters you can have 0 parameters, which means that it would technically be acceptable to write `addSubviews()`, which makes no sense, semantically. Of course you can pass an empty array, but you're still passing an object to the function.
- If you have a group of views that you want to add and remove you may very well save them in an array, in which case having an array, rather than a variadic parameter, allows you to reuse the array instead of having to write out all the views every time you want to use the function.

Personally, I feel like variadic parameters are a legacy of C code. I can't think of many examples where it would be preferable to arrays. And it's not like there's a lot of extra overhead between writing
`foo([a, b, c])` and `foo(a, b, c)`
especially since the variadic parameter version is more limiting.

PR #532, opened 2018-08-02
Add `insetBy(top:left:bottom:right:)` and `insetBy(horizontal:vertical)` to UIEdgeInsets.

VincentSit:
This is useful in some situations, for example:

```
  let insets = UIEdgeInsets(top: 10, left: 15, bottom: 20, right: 25)
  view.frame.inset(by: insets.insetBy(bottom: 5))
```
vs

```
  let insets = UIEdgeInsets(top: 10, left: 15, bottom: 20, right: 25)
  let insets2 = UIEdgeInsets(top: insets.top, left: insets.left, bottom: insets.bottom + 5, right: insets.bottom)
  view.frame.inset(by: insets2)
```

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 - **UIViewController**
   - Added `addChildViewController(_:toContainerView)` to easily add child view controllers. Accepts a `UIViewController` and a `UIView` to add the child's view to. 
   - Added `removeViewAndControllerFromParentViewController()` to remove a `UIViewController` from its parent.
+- **UIEdgeInsets**
+  - Added  `insetBy(top:)`, `insetBy(left:)`, `insetBy(bottom:)`, `insetBy(right:)`, `insetBy(horizontal:)` and `insetBy(vertical:)` to creates an `UIEdgeInsets` based on current value and adjusted by given offset. [#532](https://github.com/SwifterSwift/SwifterSwift/pull/532) by [VincentSit](https://github.com/VincentSit).
 
 ### Changed
 - **RangeReplaceableCollection**:
--- a/Sources/Extensions/UIKit/UIEdgeInsetsExtensions.swift
+++ b/Sources/Extensions/UIKit/UIEdgeInsetsExtensions.swift
@@ -44,6 +44,60 @@ extension UIEdgeInsets {
     public init(horizontal: CGFloat, vertical: CGFloat) {
         self.init(top: vertical/2, left: horizontal/2, bottom: vertical/2, right: horizontal/2)
     }
+
+    /// SwifterSwift: Creates an `UIEdgeInsets` based on current value and top offset.
+    ///
+    /// - Parameters:
+    ///   - top: Offset to be applied in to the top edge.
+    /// - Returns: UIEdgeInsets offset with given offset.
+    public func insetBy(top: CGFloat) -> UIEdgeInsets {
+        return UIEdgeInsets(top: self.top + top, left: left, bottom: bottom, right: right)
+    }
+
+    /// SwifterSwift: Creates an `UIEdgeInsets` based on current value and left offset.
+    ///
+    /// - Parameters:
+    ///   - left: Offset to be applied in to the left edge.
+    /// - Returns: UIEdgeInsets offset with given offset.
+    public func insetBy(left: CGFloat) -> UIEdgeInsets {
+        return UIEdgeInsets(top: top, left: self.left + left, bottom: bottom, right: right)
+    }
+
+    /// SwifterSwift: Creates an `UIEdgeInsets` based on current value and bottom offset.
+    ///
+    /// - Parameters:
+    ///   - bottom: Offset to be applied in to the bottom edge.
+    /// - Returns: UIEdgeInsets offset with given offset.
+    public func insetBy(bottom: CGFloat) -> UIEdgeInsets {
+        return UIEdgeInsets(top: top, left: left, bottom: self.bottom + bottom, right: right)
+    }
+
+    /// SwifterSwift: Creates an `UIEdgeInsets` based on current value and right offset.
+    ///
+    /// - Parameters:
+    ///   - right: Offset to be applied in to the right edge.
+    /// - Returns: UIEdgeInsets offset with given offset.
+    public func insetBy(right: CGFloat) -> UIEdgeInsets {
+        return UIEdgeInsets(top: top, left: left, bottom: bottom, right: self.right + right)
+    }
+
+    /// SwifterSwift: Creates an `UIEdgeInsets` based on current value and horizontal value equally divided and applied to right offset and left offset.
+    ///
+    /// - Parameters:
+    ///   - horizontal: Offset to be applied to right and left.
+    /// - Returns: UIEdgeInsets offset with given offset.
+    public func insetBy(horizontal: CGFloat) -> UIEdgeInsets {
+        return UIEdgeInsets(top: top, left: left + horizontal/2, bottom: bottom, right: right + horizontal/2)
+    }
+
+    /// SwifterSwift: Creates an `UIEdgeInsets` based on current value and vertical value equally divided and applied to top and bottom.
+    ///
+    /// - Parameters:
+    ///   - vertical: Offset to be applied to top and bottom.
+    /// - Returns: UIEdgeInsets offset with given offset.
+    public func insetBy(vertical: CGFloat) -> UIEdgeInsets {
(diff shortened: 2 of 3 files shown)


Sources/Extensions/UIKit/UIEdgeInsetsExtensions.swift
```diff
@@ -44,6 +44,29 @@ extension UIEdgeInsets {
...
+
+    /// SwifterSwift: Creates an `UIEdgeInsets` based on current value and offsets.
+    ///
+    /// - Parameters:
+    ///   - top: Offset to be applied in to the top edge.
+    ///   - left: Offset to be applied in to the left edge.
+    ///   - bottom: Offset to be applied in to the bottom edge.
+    ///   - right: Offset to be applied in to the right edge.
+    /// - Returns: UIEdgeInsets offset with given offsets.
+    public func insetBy(top: CGFloat = 0, left: CGFloat = 0, bottom: CGFloat = 0, right: CGFloat = 0) -> UIEdgeInsets {
```
LucianoPAlmeida: Giving default values for all parameter allow the user to call insets.insetBy() with no parameters. I'm not sure if it a viable option for an API standpoint. @VincentSit  What are your thoughts about it?

CHANGELOG.md
```diff
@@ -14,6 +14,8 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 - **UIViewController**
   - Added `addChildViewController(_:toContainerView)` to easily add child view controllers. Accepts a `UIViewController` and a `UIView` to add the child's view to. 
   - Added `removeViewAndControllerFromParentViewController()` to remove a `UIViewController` from its parent.
+- **UIEdgeInsets**
+  - Added  `insetBy(top:left:bottom:right:)` and `insetBy(horizontal:vertical)` to creates an `UIEdgeInsets` based on current value and  adjusted by given offsets. [#500](https://github.com/SwifterSwift/SwifterSwift/pull/532) by [VincentSit](https://github.com/VincentSit).
```
LucianoPAlmeida: The PR number is describe as 500? I believe it's 532 :))

VincentSit: Thank you, fixed.

Sources/Extensions/UIKit/UIEdgeInsetsExtensions.swift
VincentSit: I have also noticed the problem.

If do not add a default value, it is not necessary to invoke the method to pass all the values. This is probably the easiest way to do this, and my other idea is to add an edge option set and use it as an argument. ie:

```
public enum Edges: OptionSet {
    case top(CGFloat)
    case left(CGFloat)
    case bottom(CGFloat)
    case right(CGFloat)
    case horizontal(CGFloat)
    case vertical(CGFloat)
    case all(CGFloat)
}

public func insetBy(edge: Edges) -> UIEdgeInsets {}

let insets = UIEdgeInsets(top: 10, left: 15, bottom: 20, right: 25)
view.frame.inset(by: insets.insetBy(.bottom(5)))
view.frame.inset(by: insets.insetBy([.bottom(5), .right(10)]))
```

VincentSit: In theory, the parameter( `edges` in `insetBy(edges:)`) should use `OptionSet`, but I tried to find that `OptionSet` and the Enum case with parameters do not work well together. 

Currently I have two options:
1. Use an Edge array if you think that only one case is also needed to use an array is acceptable. E.g. `[.top(10)]`, `[.top(10), .left(15)]`.
2. Split into separate functions. E.g. `insetBy(top:)`, `insetBy(left:)`, `insetBy(bottom:)`...

If there is no better solution, it is okay to reject this PR.

SD10:
@VincentSit I think that makes sense now 👍  Sorry, reviewed this PR too early in the morning 😂

Sources/Extensions/UIKit/UIEdgeInsetsExtensions.swift
LucianoPAlmeida: Hey @VincentSit 
Seems a viable solution, but I believe that the fact it requires creating a new type makes it not fit well with this project :) 
I was thinking that if you only define only the four methods `insetBy(top:)`, `insetBy(left:)`, `insetBy(bottom:)` and `insetBy(right:)`. We can still have the abstraction, and from the API standpoint the user will be able to compose the call like
```swift
let insets = UIEdgeInsets(top: 10, left: 15, bottom: 20, right: 25)
view.frame.inset(by: insets.insetBy(bottom:5))
// or composing
view.frame.inset(by: insets.insetBy(bottom:5)
                                             .insetBy(right: 5))
view.frame.inset(by: insets.insetBy(bottom: 5)
                                             .insetBy(horizontal: 10))
```
I believe it's an option 👍 What you think?
No need to reject the PR, we can start a good discussion here and see if we can come up with more options :))

VincentSit:
@SD10 You don't need say sorry.  I am very grateful to you for taking the time to review the code.

Sources/Extensions/UIKit/UIEdgeInsetsExtensions.swift
VincentSit: Hi @LeLuckyVint 
I understand what you mean and agree with your suggestion. But if you only define the four methods you mentioned, `insetBy(horizontal:)` doesn't seem to be able to be called.

LucianoPAlmeida: Hey @VincentSit I mention the 4 just as example, my idea was all of them follow the same pattern thats why I put the horizontal in the code sample :))

VincentSit: Hi @LucianoPAlmeida I have updated the code. Please review and give your opinion when you are free.

Sources/Extensions/UIKit/UIEdgeInsetsExtensions.swift
```diff
@@ -44,6 +44,60 @@ extension UIEdgeInsets {
...
+    public func insetBy(bottom: CGFloat) -> UIEdgeInsets {
+        return UIEdgeInsets(top: top, left: left, bottom: self.bottom + bottom, right: right)
+    }
+
+    /// SwifterSwift: Creates an `UIEdgeInsets` based on current value and right offset.
+    ///
+    /// - Parameters:
+    ///   - top: Offset to be applied in to the right edge.
+    /// - Returns: UIEdgeInsets offset with given offset.
+    public func insetBy(right: CGFloat) -> UIEdgeInsets {
```
SD10: Think we can condense all of these into:

```swift
func insetBy(right: CGFloat? = nil, left: CGFloat? = nil, top: CGFloat? = nil, bottom: CGFloat? = nil) {
      // ...
}

func insetBy(vertical: CGFloat? = nil, horizontal: CGFloat? = nil) {
    // ...
}
```
Then users can add the parameters they choose?

VincentSit: Hi @SD10 this design will allow the user to call `insets.insetBy()` with no parameters. [Check out our discussion above](https://github.com/SwifterSwift/SwifterSwift/pull/532#discussion_r207216832).

SD10: Yup, great point again 😆 🥇

Sources/Extensions/UIKit/UIEdgeInsetsExtensions.swift
```diff
@@ -44,6 +44,60 @@ extension UIEdgeInsets {
...
+    ///   - top: Offset to be applied in to the bottom edge.
+    /// - Returns: UIEdgeInsets offset with given offset.
+    public func insetBy(bottom: CGFloat) -> UIEdgeInsets {
+        return UIEdgeInsets(top: top, left: left, bottom: self.bottom + bottom, right: right)
+    }
+
+    /// SwifterSwift: Creates an `UIEdgeInsets` based on current value and right offset.
+    ///
+    /// - Parameters:
+    ///   - top: Offset to be applied in to the right edge.
```
LucianoPAlmeida: @VincentSit  All the docs are with parameter `top` 😆

VincentSit: Whoops. Good catch, thanks. Fixed. :joy:

SD10:
I don't have time to review this and I'm not just going to slap my approval on it meaninglessly 😂 Feel free to make a decision without me. I trust you guys 💯

BennX:
@omaralbeik I wanted to merge it a second ago 😆

PR #536, opened 2018-08-06
Add operators for CGSize

kldaniel:
## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,15 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
   - Added `removeViewAndControllerFromParentViewController()` to remove a `UIViewController` from its parent.
 - **UIEdgeInsets**
   - Added  `insetBy(top:)`, `insetBy(left:)`, `insetBy(bottom:)`, `insetBy(right:)`, `insetBy(horizontal:)` and `insetBy(vertical:)` to creates an `UIEdgeInsets` based on current value and adjusted by given offset. [#532](https://github.com/SwifterSwift/SwifterSwift/pull/532) by [VincentSit](https://github.com/VincentSit).
+  - **CGSize**
+  - Added operator `+` to return the addition of two CGSize.
+  - Added operator `+=` to add a CGSize to another.
+  - Added operator `-` to return the subtraction of two CGSize.
+  - Added operator `-=` to subtract a CGSize from another.
+  - Added operator `CGSize * CGSize` to return the multiplication of two CGSize.
+  - Added operator `CGSize * CGFloat` and `CGFloat * CGSize` to return the multiplication of a CGSize and a CGFloat value.
+  - Added operator `CGSize *= CGSize` to multiply a CGSize with another one.
+  - Added operator `CGSize *= CGFloat` to multiply a CGSize with a CGFloat value.
 
 ### Changed
 - **RangeReplaceableCollection**:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
   - Added `removeViewAndControllerFromParentViewController()` to remove a `UIViewController` from its parent.
 - **UIEdgeInsets**
   - Added  `insetBy(top:)`, `insetBy(left:)`, `insetBy(bottom:)`, `insetBy(right:)`, `insetBy(horizontal:)` and `insetBy(vertical:)` to creates an `UIEdgeInsets` based on current value and adjusted by given offset. [#532](https://github.com/SwifterSwift/SwifterSwift/pull/532) by [VincentSit](https://github.com/VincentSit).
-  - **CGSize**
+- **CGSize**
   - Added operator `+` to return the addition of two CGSize.
   - Added operator `+=` to add a CGSize to another.
   - Added operator `-` to return the subtraction of two CGSize.
--- a/Sources/Extensions/CoreGraphics/CGSizeExtensions.swift
+++ b/Sources/Extensions/CoreGraphics/CGSizeExtensions.swift
@@ -52,9 +52,9 @@ public extension CGSize {
 
 }
 
-//MARK: - Operators
+// MARK: - Operators
 public extension CGSize {
-    
+
     /// SwifterSwift: Add two CGSize
     ///
     ///     let sizeA = CGSize(width: 5, height: 10)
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -16,6 +16,15 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
   - Added `removeViewAndControllerFromParentViewController()` to remove a `UIViewController` from its parent.
 - **UIEdgeInsets**
   - Added  `insetBy(top:)`, `insetBy(left:)`, `insetBy(bottom:)`, `insetBy(right:)`, `insetBy(horizontal:)` and `insetBy(vertical:)` to creates an `UIEdgeInsets` based on current value and adjusted by given offset. [#532](https://github.com/SwifterSwift/SwifterSwift/pull/532) by [VincentSit](https://github.com/VincentSit).
+  - **CGSize**
```
LucianoPAlmeida: Is this identation correct?

Sources/Extensions/CoreGraphics/CGSizeExtensions.swift
```diff
@@ -50,5 +50,138 @@ public extension CGSize {
...
+    ///     let sizeA = CGSize(width: 5, height: 10)
+    ///     let sizeB = CGSize(width: 3, height: 4)
+    ///     sizeA += sizeB
+    ///     //sizeA = CGPoint(width: 8, height: 14)
+    ///
+    /// - Parameters:
+    ///   - lhs: self
+    ///   - rhs: CGSize to add.
+    public static func += (lhs: inout CGSize, rhs: CGSize) {
+        lhs = CGSize(width: lhs.width + rhs.width, height: lhs.height + rhs.height)
```
LucianoPAlmeida: I believe you can do in-place mutation here like:
```swift
       lhs.height += rhs.height
       lhs.width += rhs.width
```
Instead of creating a new instance of cgsize and make the attribution.

Sources/Extensions/CoreGraphics/CGSizeExtensions.swift
```diff
@@ -50,5 +50,138 @@ public extension CGSize {
...
+    ///     let sizeA = CGSize(width: 5, height: 10)
+    ///     let sizeB = CGSize(width: 3, height: 4)
+    ///     sizeA -= sizeB
+    ///     //sizeA = CGPoint(width: 2, height: 6)
+    ///
+    /// - Parameters:
+    ///   - lhs: self
+    ///   - rhs: CGSize to subtract.
+    public static func -= (lhs: inout CGSize, rhs: CGSize) {
+        lhs = CGSize(width: lhs.width - rhs.width, height: lhs.height - rhs.height)
```
LucianoPAlmeida: The same comment about in-place mutation :))

Sources/Extensions/CoreGraphics/CGSizeExtensions.swift
```diff
@@ -50,5 +50,138 @@ public extension CGSize {
...
+    ///     let sizeA = CGSize(width: 5, height: 10)
+    ///     let sizeB = CGSize(width: 3, height: 4)
+    ///     sizeA *= sizeB
+    ///     //result = CGSize(width: 15, height: 40)
+    ///
+    /// - Parameters:
+    ///   - lhs: self.
+    ///   - rhs: CGSize to multiply.
+    public static func *= (lhs: inout CGSize, rhs: CGSize) {
+        lhs = CGSize(width: lhs.width * rhs.width, height: lhs.height * rhs.height)
```
LucianoPAlmeida: The same about in-place mutation :))

CHANGELOG.md
kldaniel: Absolutely not correct, i see. :) @LucianoPAlmeida thanks for your comments and suggetstions on my very first PR.

Sources/Extensions/CoreGraphics/CGSizeExtensions.swift
kldaniel: Modified all of them.

Sources/Extensions/CoreGraphics/CGSizeExtensions.swift
```diff
@@ -50,5 +50,142 @@ public extension CGSize {
...
+    ///     let sizeA = CGSize(width: 5, height: 10)
+    ///     let sizeB = CGSize(width: 3, height: 4)
+    ///     let result = sizeA * sizeB
+    ///     //result = CGSize(width: 15, height: 40)
+    ///
+    /// - Parameters:
+    ///   - lhs: CGSize to multiply.
+    ///   - rhs: CGSize to multiply with.
+    /// - Returns: The result comes from the multiplication of the two given CGSize structs.
+    public static func * (lhs: CGSize, rhs: CGSize) -> CGSize {
```
guykogus: I don't understand the idea of multiplying 2 sizes. Multiplying by a scalar makes perfect sense, but when will you ever multiply sizes? I'm just curious to know of any examples.

kldaniel: According to the docs (https://developer.apple.com/documentation/coregraphics/cgsize), CGSize struct may be used to represent a distance vector. The idea behind multiplying sizes comes from multiplying those vectors.

I have to admit i have never used CGSizes for this purpose before.

You may want to manipulate the size of a view's frame defined by a vector and not a scalar value (specially when you transform its size not proportionally).

I suppose a use-case like this below:
```swift
let someVector: CGSize = CGsize(width: 1.5, height: 4)
aView.frame.size *= someVector
```

In this case someVector is a type of CGSize but represents a vector rather than a size.

@guykogus what do you think? Am i wrong in this situation? Is there a better solution for this?

guykogus: I think that if you want a vector you'd be better off using `CGVector`, no?
But maybe I can see what you mean, like if you have an image that's scaled (losing aspect ratio) to fit a view, and you want to get the ratios you could maybe do `image.size / view.frame.size` and then use that size to multiply something connected.

kldaniel: I agree a vector is better to be represented as a CGVector rather than a CGSize.

I realize that in my last example it would be a better fit to multiply the size with a CGVector. Then we should add an operator for multiply a cgsize with a cgvector instead of cgsize with cgsize. Shall i? :)

guykogus: No, I think it's fine to leave it with `CGSize`.

LucianoPAlmeida:
Hey @kldaniel can you solve the conflicts with master :)

SD10:

PR #538, opened 2018-08-08
Add '?=' (ConditionalAssignment) operator to Optional

viktart:
🚀

## Checklist

### DIFF
--- a/Sources/Extensions/SwiftStdlib/OptionalExtensions.swift
+++ b/Sources/Extensions/SwiftStdlib/OptionalExtensions.swift
@@ -74,6 +74,22 @@ public extension Optional {
         lhs = rhs
     }
 
+    /// SwifterSwift: Assign an optional value to a variable only if the variable is nil.
+    ///
+    ///     var someText: String? = nil
+    ///     let newText = "Foo"
+    ///     let defaultText = "Bar"
+    ///     someText ?= newText //someText is now "Foo" because it was nil before
+    ///     someText ?= defaultText //someText doesn't change its value because it's not nil
+    ///
+    /// - Parameters:
+    ///   - lhs: Any?
+    ///   - rhs: Any?
+    public static func ?= (lhs: inout Optional, rhs: @autoclosure () -> Optional) {
+        if lhs == nil {
+            lhs = rhs()
+        }
+    }
 }
 
 // MARK: - Methods (Collection)
@@ -89,3 +105,4 @@ public extension Optional where Wrapped: Collection {
 
 // MARK: - Operators
 infix operator ??= : AssignmentPrecedence
+infix operator ?= : AssignmentPrecedence
--- a/Tests/SwiftStdlibTests/OptionalExtensionsTests.swift
+++ b/Tests/SwiftStdlibTests/OptionalExtensionsTests.swift
@@ -63,6 +63,24 @@ final class OptionalExtensionsTests: XCTestCase {
 		XCTAssert(parameters[key2] == parameter2)
 	}
 
+    func testConditionalAssignment() {
+        var text1: String?
+        text1 ?= "newText1"
+        XCTAssertEqual(text1, "newText1")
+
+        var text2: String? = "text2"
+        text2 ?= "newText2"
+        XCTAssertEqual(text2, "text2")
+
+        var text3: String?
+        text3 ?= nil
+        XCTAssertEqual(text3, nil)
+
+        var text4: String? = "text4"
+        text4 ?= nil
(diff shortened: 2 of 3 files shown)


Tests/SwiftStdlibTests/OptionalExtensionsTests.swift
```diff
@@ -63,6 +63,24 @@ final class OptionalExtensionsTests: XCTestCase {
 		XCTAssert(parameters[key2] == parameter2)
 	}
 
+    func testConditionalAssignment() {
+        var text1: String?
+        text1 ?= "newText1"
+        XCTAssert(text1 == "newText1")
```
guykogus: I think it's cleaner to use `XCTAssertEquals`

Tests/SwiftStdlibTests/OptionalExtensionsTests.swift
```diff
@@ -63,6 +63,24 @@ final class OptionalExtensionsTests: XCTestCase {
 		XCTAssert(parameters[key2] == parameter2)
 	}
 
+    func testConditionalAssignment() {
```
guykogus: What happens if you do multiple `?=` calls in 1 line, e.g.
```Swift
a ?= b ?= c
```
when `a` and `b` are `nil` but `c` has a value. Will `a` get the value of `c`? Will `b` also get its value? Can you test for that?
And if `b` is not `nil`, will `a` become `b` and `b` keep its value? Please test for that too.
And what if `a` is not `nil`, `b` is `nil` and `c` is not `nil`, will `b` become `c`?

Sources/Extensions/SwiftStdlib/OptionalExtensions.swift
```diff
@@ -89,3 +105,4 @@ public extension Optional where Wrapped: Collection {
 
 // MARK: - Operators
 infix operator ??= : AssignmentPrecedence
+infix operator ?= : AssignmentPrecedence
```
guykogus: How does precedence work? If I call
```Swift
a ?= b ?? c
```
will it know to use `??` before `?=` (which is what seems logical to me in this scenario)

Tests/SwiftStdlibTests/OptionalExtensionsTests.swift
viktart: It won't compile, because `?=` operator just like the assignment operator doesn't return any value ('a = b = c' won't compile too).

Sources/Extensions/SwiftStdlib/OptionalExtensions.swift
viktart: Sure, it will call `??` before `?=`, because `?=` is AssignmentPrecedence just like the assignment operator (https://developer.apple.com/documentation/swift/swift_standard_library/operator_declarations)

Tests/SwiftStdlibTests/OptionalExtensionsTests.swift
viktart: fixed

viktart:
Sorry, I amended the commit instead of adding a new one (my first pull request on GitHub :)) 
I changed `XCTAssert` to `XCTAssertEqual`.
I also made `@autoclosure rhs` instead of simple parameter so it won't evaluate unless needed.

Sources/Extensions/SwiftStdlib/OptionalExtensions.swift
BennX: Assignment precedence has a value of 90, which is the lowest precedence value of operators in swift. This should mean that it get evaluated last. (I think we can find those values in `import Swift`)

Tests/SwiftStdlibTests/OptionalExtensionsTests.swift
BennX:
```
a ?= (b ?= c)
``` 
should work  as @guykogus  explained. 

Could you add the tests for those scenarios?

viktart: My original idea was to create a conditional assignment operator with the same properties as the original assignment operator (and usual assignment operator doesn't return any value, so `a = (b = c)` won't compile.
On the other hand, the idea of the operator is borrowed from Ruby's conditional assignment operator `||=` and in Ruby assignment operators do return values.

I think that we should stick with Swift's paradigm of non-returning assignments for the sake of consistency.

SD10:
I haven't reviewed this PR but I like the idea 👍

SD10:

PR #539, opened 2018-08-11
Added method to anchor UIView to its superview

viktart:
## Checklist


Sources/Extensions/UIKit/UIViewExtensions.swift
```diff
@@ -464,6 +464,36 @@ public extension UIView {
...
 
+    /// SwifterSwift: Anchors all sides to superview's safe area.
+    ///
+    /// 1. On iOS 11 and latter anchors view's sides to superview's safe area
+    /// 2. On iOS 9-11 anchors leading and trailing to superview,
+    /// top and bottom to viewController's bottomLayoutGuide and topLayoutGuide.
+    ///
+    /// - Parameter viewController: view controller, of which bottomLahoutGuide
+    ///  and topLayoutGuide are used on iOS 9-11
+    @available(iOS 9, *) public func fillToSuperviewSafeArea(viewController: UIViewController) {
```
LucianoPAlmeida: I'm not sure if the parameter the view controller here is necessary. 
Can you explain a use case where passing the view controller just to use the topLayoutGuide necessary?
And not just use superview. topAnchor is enough.

Sources/Extensions/UIKit/UIViewExtensions.swift
```diff
@@ -464,6 +464,36 @@ public extension UIView {
...
 	}
 
+    /// SwifterSwift: Anchors all sides to superview's safe area.
+    ///
+    /// 1. On iOS 11 and latter anchors view's sides to superview's safe area
+    /// 2. On iOS 9-11 anchors leading and trailing to superview,
+    /// top and bottom to viewController's bottomLayoutGuide and topLayoutGuide.
+    ///
+    /// - Parameter viewController: view controller, of which bottomLahoutGuide
+    ///  and topLayoutGuide are used on iOS 9-11
```
LucianoPAlmeida: By naming convention I'm not sure if this is expressive enough, maybe something like `constraintEquallyToSuperview()`, because it express more the idea that the it rely on contraints. 
Just a suggestion, let me know you opinion on that :))

Sources/Extensions/UIKit/UIViewExtensions.swift
LucianoPAlmeida: Can you add unit tests for this?

Sources/Extensions/UIKit/UIViewExtensions.swift
SD10: Don't think we need to express `equally` specifically. Something like `constrainToSuperview`, or `fillSuperView` should be a good name. I think the best names are descriptive while not being overly explicit

SD10: Ok so we already have `fillToSuperview` in SwifterSwift. I think `fillToSuperviewSafeArea` is a good name for this

Sources/Extensions/UIKit/UIViewExtensions.swift
```diff
@@ -464,6 +464,36 @@ public extension UIView {
...
+        translatesAutoresizingMaskIntoConstraints = false
+
+        let leading, top, trailing, bottom: NSLayoutConstraint
+        if #available(iOS 11, *) {
+            leading = self.leadingAnchor.constraint(equalTo: superview.safeAreaLayoutGuide.leadingAnchor)
+            top = self.topAnchor.constraint(equalTo: superview.safeAreaLayoutGuide.topAnchor)
+            trailing = self.trailingAnchor.constraint(equalTo: superview.safeAreaLayoutGuide.trailingAnchor)
+            bottom = self.bottomAnchor.constraint(equalTo: superview.safeAreaLayoutGuide.bottomAnchor)
+        } else {
+            leading = self.leadingAnchor.constraint(equalTo: superview.leadingAnchor)
```
LucianoPAlmeida: Since we notice the there's already a extension `fillToSuperview()` that handles this case, I believe is a good idea not handle this case here, let this extension only handle the safeArea (since it's named `fillToSuperviewSafeArea`) case and also constraint the whole method to iOS 11, so the user can choose to use this the rely on safeArea or the other that we already have. 
@SD10 What you think?

SD10: @LucianoPAlmeida That makes sense, I'm not opposed to limiting this to iOS 11

Sources/Extensions/UIKit/UIViewExtensions.swift
viktart: At first I wanted to pass `bottomLayoutGuide` and `topLayoutGuide` as parameters, but if you use them in your code, you get deprecation warning. However, in the function `bottomLayoutGuide` and `topLayoutGuide` are used inside `else` closure to `#available(iOS 11, *)` so the compiler doesn't generate deprecation warning.

Sources/Extensions/UIKit/UIViewExtensions.swift
viktart: @LucianoPAlmeida I actually agree with you: `constraintViewToSuperviewSafeArea` is a better name than `fillToSuperviewSafeArea` since it uses the same amount of words, but gives us more information about the result of function's execution (I mean that it uses autolayout and not frames or autoresizing masks). 
I think it's worth considering changing this function's name together with already existing `fillToSuperview`.

Sources/Extensions/UIKit/UIViewExtensions.swift
viktart: Ok, I'll try to think of a way.

Sources/Extensions/UIKit/UIViewExtensions.swift
```diff
@@ -464,6 +464,36 @@ public extension UIView {
...
+    @available(iOS 9, *) public func fillToSuperviewSafeArea(viewController: UIViewController) {
+        guard let superview = self.superview else {
+            fatalError("\(#function) : the view isn't in view hierarchy")
+        }
+
+        translatesAutoresizingMaskIntoConstraints = false
+
+        let leading, top, trailing, bottom: NSLayoutConstraint
+        if #available(iOS 11, *) {
+            leading = self.leadingAnchor.constraint(equalTo: superview.safeAreaLayoutGuide.leadingAnchor)
```
guykogus: Please remove `self.` (wherever possible)

Sources/Extensions/UIKit/UIViewExtensions.swift
```diff
@@ -464,6 +464,36 @@ public extension UIView {
 		}
 	}
 
+    /// SwifterSwift: Anchors all sides to superview's safe area.
+    ///
+    /// 1. On iOS 11 and latter anchors view's sides to superview's safe area
```
guykogus: later

Sources/Extensions/UIKit/UIViewExtensions.swift
```diff
@@ -464,6 +464,36 @@ public extension UIView {
...
+    /// 2. On iOS 9-11 anchors leading and trailing to superview,
+    /// top and bottom to viewController's bottomLayoutGuide and topLayoutGuide.
```
guykogus: Please add backticks (\`) around `bottomLayoutGuide` and `topLayoutGuide`

Sources/Extensions/UIKit/UIViewExtensions.swift
```diff
@@ -464,6 +464,36 @@ public extension UIView {
...
+    ///
+    /// - Parameter viewController: view controller, of which bottomLahoutGuide
```
guykogus: backticks
type: `bottomLayoutGuide`

Sources/Extensions/UIKit/UIViewExtensions.swift
```diff
@@ -464,6 +464,36 @@ public extension UIView {
...
+    ///  and topLayoutGuide are used on iOS 9-11
+    @available(iOS 9, *) public func fillToSuperviewSafeArea(viewController: UIViewController) {
+        guard let superview = self.superview else {
```
guykogus: Maybe you should also add a check for `super == viewController.view`? Does that make sense to be a requirement?

guykogus: No need for `self.` here.

viktart: Not necessary, safe area works even for subviews.

Sources/Extensions/UIKit/UIViewExtensions.swift
LucianoPAlmeida: Hey @viktart, sorry for the delay on responding 😭 
But for sure, It's worth considering but since we found that we already have `fillToSuperview`, renaming it will be a breaking change to take into account too :))

Sources/Extensions/UIKit/UIViewExtensions.swift
viktart: done

Sources/Extensions/UIKit/UIViewExtensions.swift
viktart: done

Sources/Extensions/UIKit/UIViewExtensions.swift
viktart: done

Sources/Extensions/UIKit/UIViewExtensions.swift
viktart: done

Sources/Extensions/UIKit/UIViewExtensions.swift
viktart: done

viktart:
Fixed the notes. Added tests. Added iOS11+, tvOS11+ method without parameters. Fixed changelog.

guykogus:
Anyone know why codecov is suddenly complaining? It doesn't look like that much extra code is being missed by tests. Unfortunately using `#available(...)` causes XCTest to not be able to test every path, but it shouldn't be showing such a huge difference in coverage...

SD10:
@guykogus I've been noticing this for a while but haven't had time to dig. I think it's ok to ignore codecov in certain situations like this

guykogus:
@viktart please resolve the conflicts so that we can merge this in :)

LucianoPAlmeida:
Closing this due inactivity.
Feel free to reopen it anytime :))

PR #542, opened 2018-08-12
Data extension for converting to a JSON object

guykogus:
🚀

I've found this extension particularly useful when debugging JSON web responses. If you try and just print out the strings lldb prints them out as one big jumbled line, but if you print out these `jsonObject` responses it's _much_ cleaner.

What do you think about this being added to `String` as well?

## Checklist

### DIFF
--- a/Sources/Extensions/Foundation/DataExtensions.swift
+++ b/Sources/Extensions/Foundation/DataExtensions.swift
@@ -31,5 +31,16 @@ public extension Data {
 		return String(data: self, encoding: encoding)
 	}
 
+    /// SwifterSwift: Returns a Foundation object from given JSON data.
+    ///
+    /// - Parameter options: Options for reading the JSON data and creating the Foundation object.
+    ///
+    ///   For possible values, see `JSONSerialization.ReadingOptions`.
+    /// - Returns: A Foundation object from the JSON data in the receiver, or `nil` if an error occurs.
+    /// - Throws: An `NSError` if the receiver does not represent a valid JSON object.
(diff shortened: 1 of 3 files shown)


Sources/Extensions/Foundation/DataExtensions.swift
```diff
@@ -31,5 +31,16 @@ public extension Data {
...
 	}
 
+    /// SwifterSwift: Returns a Foundation object from given JSON data.
+    ///
+    /// - Parameter options: Options for reading the JSON data and creating the Foundation object.
+    ///
+    ///   For possible values, see `JSONSerialization.ReadingOptions`.
+    /// - Returns: A Foundation object from the JSON data in the receiver, or `nil` if an error occurs.
+    /// - Throws: An `NSError` if the receiver does not represent a valid JSON object.
+    public func jsonObject(options: JSONSerialization.ReadingOptions = []) throws -> Any {
```
guykogus: I'me really tempted to use `[.allowFragments]` as the default value, but that would be inconsistent with `JSONSerialization.jsonObject(with:options:)`

LucianoPAlmeida:
Hey @guykogus :))
I believe it will be a useful thing adding to string too, but only concern is that even we using #canImport(Foundation) it will be a extension that adds a little dependency on Foundation. For something like that I'll probably prefer a JSONSerialization extension `JSONSerialization.jsonObject(with string: String, options opt: JSONSerialization.ReadingOptions = [])`(Not sure if we already have that). But it is just a option, as I said we already use canImport statements in some String extensions that already use some foundation types, so ... I think there's no problem to add it to string too :)) Just pointing it out to see it worth a discussion 👍

PR #546, opened 2018-08-15
Create mapping functions for Dictionary

guykogus:
🚀

These versions of `map` and `compactMap` return a `Dictionary`, rather than an `Array`.
I'm a bit unsure of the function names because of the clash with the built-in `map` and `compactMap` functions.

------------------------------------------------

`map` seems to prioritise the new function. E.g. in
```Swift
let intToString = [0: "0", 1: "1", 2: "2"]
let stringToInt = intToString.map { (key, value) in
    return (String(describing: key), Int(value)!)
}
```
`stringToInt ` is of type `[String: Int]`. If you want it to return `[(String, Int)]`, as the original function would do, you need to define the type explicitly. I.e.
```Swift
let stringToInt: [(String, Int)] = intToString.map { (key, value) in
```

------------------------------------------------

`compactMap` actually won't compile in a similar situation because it prioritises the built-in function. E.g.
```Swift
let stringToInt = intToString.compactMap { (key, value) in
    guard let intValue = Int(value) else { return nil }
    return (String(describing: key), intValue)
}
```
will throw a compilation error because it thinks that the the closure passed to `compactMap` should contain a single argument, it won't accept a tuple. You need to explicitly define it as
```Swift
let stringToInt: [String: Int] = intToString.compactMap { (key, value) in
// or
let stringToInt = intToString.compactMap { (key, value) -> (String, Int)? in
```
in order for it to compile and use the new function.

To use the original function you again need to explicitly define the type. I.e.
```Swift
let words: [(String, IntWord)] = strings.compactMap { (key, value) in
```

What do you guys think, should we rename the functions? What names sound good to you?

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
   - `beginRefresh(in tableView:, animated:, sendAction:)` UIRefreshControl extension to begin refresh programatically. [#525](https://github.com/SwifterSwift/SwifterSwift/pull/525) by [ratulSharker](https://github.com/ratulSharker)
 - **Dictionary**:
   - Added `removeValueForRandomKey()` to remove a value for a random key from a dictionary. [#497](https://github.com/SwifterSwift/SwifterSwift/pull/497) by [vyax](https://github.com/vyax).
+  - Added `map(_:)` to map a `Dictionary` into a `Dictionary` with different `Key` and `Value` types. [#546](https://github.com/SwifterSwift/SwifterSwift/pull/546) by [guykogus](https://github.com/guykogus)
+  - Added `compactMap(_:)` to map a `Dictionary` into a `Dictionary`, excluding `nil` results, with different `Key` and `Value` types. [#546](https://github.com/SwifterSwift/SwifterSwift/pull/546) by [guykogus](https://github.com/guykogus)
 - **RangeReplaceableCollection**:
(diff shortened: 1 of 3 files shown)


SD10:
@guykogus This is interesting but I don't like how we're overloading `map` and `compactMap` with specific functionality here. The fact that the overload returns a specific type -- dictionary -- is counterintuitive to the general concept of map

guykogus:
For me it makes more sense for `map` to actually return the same type, but you're right that overloading is not a good idea (especially considering the issues I outlined). So my question remains, what names sound good? `Dictionary` has a function called `mapValues` that does something similar, so maybe we could use `mapKeysAndValues`? Thoughts?

guykogus:
OK, updated the names.
I'm confused as to what you mean about `map` being polymorphic. Are these functions not polymorphic? They convert a `Dictionary<K1, V1>` into `Dictionary<K2, V2>`.

SD10:
@guykogus I guess they are over the generic key/value, but the return type is still bound to `Dictionary`

guykogus:
@omaralbeik I stole the documentation from Apple's docs 😄. The complexity of `compactMap` is declared to be _O(m + n)_, so for the dictionary it would probably be more like _O(m + (2 * n))_, but since constants are ignored for big-O notation it's still technically correct.

LucianoPAlmeida:
Sorry couldn't be on the discussion 😭 
But cool extensions @guykogus :))

guykogus:
I'm a big fan, especially for using dictionaries where the keys are enums that I eventually need to pass as the parameters to an HTTP call, so they need to be converted to their raw values. I feel like there'd be lots of useful scenarios for this...

PR #555, opened 2018-08-26
Add `rotated(by:)` functions to `UIImage` for generating rotated images.

guykogus:
🚀

There are 2 versions, one where the angle is passed directly as a `CGFloat` in radians, the other using the `Measurement` class, which is only available for iOS 10+.

## Checklist

### DIFF
--- a/Sources/Extensions/UIKit/UIImageExtensions.swift
+++ b/Sources/Extensions/UIKit/UIImageExtensions.swift
@@ -96,7 +96,7 @@ public extension UIImage {
 		return newImage
 	}
 
-	/// Creates a copy of the receiver rotated by the given angle.
+    /// SwifterSwift: Creates a copy of the receiver rotated by the given angle.
 	///
 	///     // Rotate the image by 180°
 	///     image.rotated(by: Measurement(value: 180, unit: .degrees))
@@ -129,7 +129,7 @@ public extension UIImage {
 		return newImage
 	}
 
-	/// Creates a copy of the receiver rotated by the given angle (in radians).
+	/// SwifterSwift: Creates a copy of the receiver rotated by the given angle (in radians).
 	///
 	///     // Rotate the image by 180°
 	///     image.rotated(by: .pi)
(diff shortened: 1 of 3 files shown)


Sources/Extensions/UIKit/UIImageExtensions.swift
```diff
@@ -96,6 +96,69 @@ public extension UIImage {
...
 	}
 
+    /// Creates a copy of the receiver rotated by the given angle.
+    ///
+    ///     // Rotate the image by 180°
+    ///     image.rotated(by: Measurement(value: 180, unit: .degrees))
+    ///
+    /// - Parameter angle: The angle measurement by which to rotate the image.
+    /// - Returns: A new image rotated by the given angle.
+    @available(iOS 10.0, *)
```
LucianoPAlmeida: @guykogus This availability check only makes it available to iOS 10.0+, but I think we can add it also to tvOS 10.0+, macOS 10.12+ and watchOS 3.0+ :))  Like `@available(OSX 10.12, iOS 10.0, watchOS 3.0, tvOS 10.0, *)` Since Measurement is available for them too.
What you think?

LucianoPAlmeida: Wow, I love CI 💯 , seems like the build is breaking on those other platforms because of those missing availability checks.
Now I think is a requirement 😆

Sources/Extensions/UIKit/UIImageExtensions.swift
```diff
@@ -96,6 +96,69 @@ public extension UIImage {
 		return newImage
 	}
 
+	/// Creates a copy of the receiver rotated by the given angle.
```
SD10: Needs `SwifterSwift:`

Sources/Extensions/UIKit/UIImageExtensions.swift
```diff
@@ -96,6 +96,69 @@ public extension UIImage {
...
+		draw(in: CGRect(origin: CGPoint(x: -size.width / 2,
+										y: -size.height / 2),
+						size: size))
+
+		let newImage = UIGraphicsGetImageFromCurrentImageContext()
+		UIGraphicsEndImageContext()
+		return newImage
+	}
+
+	/// Creates a copy of the receiver rotated by the given angle (in radians).
```
SD10: Needs `SwifterSwift:`

PR #562, opened 2018-09-15
Update for Swift 4.2 and Xcode 10 support

LucianoPAlmeida:
## Checklist


.travis.yml
```diff
@@ -16,6 +16,7 @@ env:
 
 before_install:
   - bundle install
+  - gem install cocoapods --pre
```
LucianoPAlmeida: Install the 1.6.0 beta because of an issue with pod lib lint 
I don't think it should be merged, but otherwise, we will have to wait for the 1.6.0 official release :))
@omaralbeik @SD10 @guykogus

tache:
This is really support for Swift 4.2, separate from Xcode 10.

.travis.yml
guykogus: I think it's fine, however this should be marked in the `Gemfile` with
`gem 'cocoapods', '>= 1.6.0.beta.1'`
rather than forcing the install here. In that case you should run `bundle update` locally and add `Gemfile.lock` to git, cos it looks like we don't have it currently 🤔.

If you're not keen to set the pre-release version and all that, I wouldn't be against having a `swift-4.2` branch and we can merge into there (I also have some new extensions for it that I'd like to add 😁). Then once cocoapods does an official release for 1.6.0 we can merge into `master`.

LucianoPAlmeida: I'm a little bit skeptical about putting this on gem file because it is just a temporary workaround, when they release it we probably will remove it because it's gonna be already installed on Travis machine by default. That's why I just put the script :))

LucianoPAlmeida:
About the random functions @guykogus 
For sure it's a good thing to update :))

guykogus:
@tache not sure if that's correct. Not only are we updating travis to use Xcode 10, but there are also some updated API changes for Xcode 10, e.g.
`NSAttributedStringKey` -> `NSAttributedString.Key`
`.UIApplicationUserDidTakeScreenshot` -> `UIApplication.userDidTakeScreenshotNotification `
etc.

guykogus:
@LucianoPAlmeida there are still a few more places where we can (and should) start using the new `random` extensions. Basically just search the project for `arc4random` and you'll see where.

guykogus:
One of those instances is _really_ bothering me. In `CGFloatExtensions.swift` there's a `randomBetween` function that's using the random functions for integers instead of floats. It's fine to just deprecate, but if you want to give it a little tweak as well that'd be 👍👍

LucianoPAlmeida:
@guykogus That's awesome, doing right now 👍 
Thank's 💯

SD10:
> Swift 4.2 comes with built-in random extensions, as well as randomElement for collections. So perhaps we should deprecate the random functions in...

Thanks for reminding us to clean this up @guykogus 💯 👍

LucianoPAlmeida:
Hey @omaralbeik For sure will remove then :))
But I always have the notion that removals are a good fit for major releases like in this case for example when we release 5.x.x

guykogus:
I like the idea of only removing functions on major updates. Kind of goes well with the whole idea of how semantic versioning works anyway.

Something that would also be nice would be if we could sync our major & minor versions with that of Swift 😄 So when Swift 5 comes out (next year, maybe?) we can sync and try to keep it that way. Actually that would mean only being able to do even minor releases with added features about once a year, wouldn't it? Maybe not the best idea then...

LucianoPAlmeida:
@guykogus @omaralbeik :))
Just removed those, although I prefer removing deprecations on a major release, I just looked at those extensions and realize that was not too impactful and would be nice do a cleanup 👍

drewpitchford:
Has this been uploaded to Cocoapods trunk? I do `pod update SwifterSwift` and I still only get version 4.4.

guykogus:
@drewpitchford if you got to the [SwifterSwift cocoapods page](https://cocoapods.org/pods/SwifterSwift) you can see that it's updated. Not sure what could be causing this problem. By any chance, could your `Podfile` be locked to a specific version like
`pod 'SwifterSwift', '= 4.4'`
or
`pod 'SwifterSwift', '~> 4.4.0'`
?
It's the only obvious thing that comes to mind.

Another alternative would be that you have a pod that's also depending on SwifterSwift, and if that pod declares its SwifterSwift dependency with a specific version in its podspec then your project will try and consolidate the right version to use with what's defined in your `Podfile` (this can fail your `pod install`ation if there's a version mismatch).

drewpitchford:
My pod file is not locked, but a dependency might be. Thanks for that idea. I’ll check. 

Drew

PR #563, opened 2018-09-15
Allow use of Encodable-only and Decodable only in UserDefaults extensions

marcocapano:
This PR substitutes the unnecessary Codable constraint with Encodable or Decodable where needed.
Not sure this is big enough to be mentioned in the CHANGELOG(?)
## Checklist
- [ ] I have added a [changelog]


guykogus:
What's the point of saving to/loading from defaults if the type is not both `Encodable` and `Decodable`?

LucianoPAlmeida:
Hey @marcocapano :))
I agree with @guykogus, I don't see as correct  to `set` an only `Encodable` if you are not able to get it as a `Decodable` later

SD10:
@marcocapano Thanks for taking the time to submit this suggestion 👍

PR #569, opened 2018-09-27
Update the CHANGELOG to correctly display what has been changed for the latest version

guykogus:
🚀

We need to create a public release to fix the cocoapods issue. Unfortunately https://github.com/SwifterSwift/SwifterSwift/pull/560 was merged before we could create release 4.5.1, so we're stuck with updating the version to 4.6.0.

## Checklist


LucianoPAlmeida:
@guykogus I think we should update the readme and the podspec to the new version 4.6.0 :))

Issue #565, opened 2018-09-25
Nested Dictionary String Key value


LucianoPAlmeida:
When having a nested JSON like 

```
{  
   "error":{  
      "statusCode":422,
      "name":"ValidationError",
      "message":"Errors ",
      "details":{  
         "context":"Context",
         "codes":{  
            "password":[  
               "presence"
            ],
            "email":[  
               "uniqueness"
            ]
         },
         "messages":{  
            "password":[  
               "can't be blank"
            ],
            "email":[  
               "Email already exists"
            ]
         }
      }
   }
}

```
And transforming to a dictionary, if we want to check the value of error->details->codes->passwod.
Probably we have to guarding nested dictionaries until we get our value like
```swift
guard let error = json["error"] as? [String: [String: Any]], 
let details = error["details"] .... till we get the value ["presence"].
```
So, with that being said, the proposedProposed extension is

```swift
extension Dictionary where Key == String {
    
    subscript(dotPath path: String) -> Any? {
        ....
    }
}

json[dotPath: "error.details.codes.password"] -> ["presence"]

```

* [x] I've read, understood, and done my best to follow the [Contributing guidelines](https://github.com/SwifterSwift/SwifterSwift/tree/master/CONTRIBUTING.md) before opening this issue.

calebkleveter:
Do we want this to return type `Any` or a generic type `T`?

calebkleveter:
Here's a fix: https://github.com/SwifterSwift/SwifterSwift/pull/574

PR #574, opened 2018-09-28
Dictionary Key-Path Subscript

calebkleveter:
Fixes https://github.com/SwifterSwift/SwifterSwift/issues/565

```swift
var dict =  ["key": ["key1": ["key2": "value"]]]
dict[path: ["key", "key1", "key2"]] = "newValue"
dict[path: ["key", "key1", "key2"]] -> "newValue"
```

## Checklist

### DIFF
--- a/Sources/Extensions/SwiftStdlib/DictionaryExtensions.swift
+++ b/Sources/Extensions/SwiftStdlib/DictionaryExtensions.swift
@@ -145,7 +145,47 @@ public extension Dictionary where Key: StringProtocol {
             }
         }
     }
+}
 
+// MARK: - Subscripts
+public extension Dictionary {
+
+    /// SwifterSwift: Deep fetch or set a value from nested dictionaries.
+    ///
+    ///        var dict =  ["key": ["key1": ["key2": "value"]]]
+    ///        dict[path: ["key", "key1", "key2"]] = "newValue"
+    ///        dict[path: ["key", "key1", "key2"]] -> "newValue"
+    ///
+    /// - Note: Value fetching is iterative, while setting is recursive.
+    ///
+    /// - Complexity: O(N), _N_ being the length of the path passed in.
+    ///
+    /// - Parameter path: An array of keys to the desired value.
+    ///
+    /// - Returns: The value for the key-path passed in. `nil` if no value is found.
+    public subscript(path path: [Key]) -> Any? {
+        get {
+            return path.reduce(self) { result, key -> Any? in
+                if let element = (result as? [Key: Any])?[key] {
+                    return element
+                } else {
+                    return nil
+                }
+            }
+        }
+        set {
+            if let first = path.first {
+                if path.count == 1, let new = newValue as? Value {
+                    return self[first] = new
+                }
+                if var nested = self[first] as? [Key: Any] {
+                    nested[path: Array(path.dropFirst())] = newValue
+                    return self[first] = nested as? Value
+                }
+            }
+            return
+        }
+    }
 }
 
 // MARK: - Operators
--- a/Tests/SwiftStdlibTests/DictionaryExtensionsTests.swift
+++ b/Tests/SwiftStdlibTests/DictionaryExtensionsTests.swift
@@ -78,6 +78,14 @@ final class DictionaryExtensionsTests: XCTestCase {
         XCTAssertEqual(dict, ["testkey": "value"])
     }
 
+    func testSubscriptKeypath() {
+        var json = ["key": ["key1": ["key2": "value"]]]
+
+        XCTAssertEqual(json[path: ["key", "key1", "key2"]] as? String, "value")
+        json[path: ["key", "key1", "key2"]] = "newValue"
+        XCTAssertEqual(json[path: ["key", "key1", "key2"]] as? String, "newValue")
+    }
+
     func testOperatorPlus() {
         let dict: [String: String] = ["key1": "value1"]
         let dict2: [String: String] = ["key2": "value2"]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 # Upcoming release
 
 ### Added
+- **Dictionary**
+  - Added `Dictionary[path:]` subscript for deep fetching/setting nested values. [#573](https://github.com/SwifterSwift/SwifterSwift/pull/573) by [@calebkleveter](https://github.com/calebkleveter)
 ### Changed
 ### Fixed
 ### Deprecated


guykogus:
I prefer more my suggestion in https://github.com/SwifterSwift/SwifterSwift/issues/565, because it aligns more with the [value(forKeyPath:)](https://developer.apple.com/documentation/objectivec/nsobject/1416468-value) method from NSObject.

calebkleveter:
@guykogus Specifically, do you mean passing in a single string instead of an array of keys?

Sources/Extensions/SwiftStdlib/DictionaryExtensions.swift
```diff
@@ -145,7 +145,47 @@ public extension Dictionary where Key: StringProtocol {
...
+    ///
+    /// - Complexity: O(N), _N_ being the length of the path passed in.
+    ///
+    /// - Parameter path: An array of keys to the desired value.
+    ///
+    /// - Returns: The value for the key-path passed in. `nil` if no value is found.
+    public subscript(path path: [Key]) -> Any? {
+        get {
+            return path.reduce(self) { result, key -> Any? in
+                if let element = (result as? [Key: Any])?[key] {
```
LucianoPAlmeida: Using reduce here is more clear, but it makes not possible to an early exit when one of the first keys are not found. 
Example: When dictionary is `["key": ["key1": ["key2": "value"]]]` and we use dic[path: ["keyRandon", "key1", "key2"]] even if it find nil in the first iteration the reduce will continue execute n times. Maybe we can consider a loop implementation here to optimize this case :))

Sources/Extensions/SwiftStdlib/DictionaryExtensions.swift
```diff
@@ -145,7 +145,47 @@ public extension Dictionary where Key: StringProtocol {
...
+            if let first = path.first {
+                if path.count == 1, let new = newValue as? Value {
+                    return self[first] = new
+                }
+                if var nested = self[first] as? [Key: Any] {
+                    nested[path: Array(path.dropFirst())] = newValue
+                    return self[first] = nested as? Value
+                }
+            }
+            return
```
LucianoPAlmeida: I think this return statement here can be removed  :))

Sources/Extensions/SwiftStdlib/DictionaryExtensions.swift
```diff
@@ -145,7 +145,47 @@ public extension Dictionary where Key: StringProtocol {
...
+    /// - Note: Value fetching is iterative, while setting is recursive.
+    ///
+    /// - Complexity: O(N), _N_ being the length of the path passed in.
+    ///
+    /// - Parameter path: An array of keys to the desired value.
+    ///
+    /// - Returns: The value for the key-path passed in. `nil` if no value is found.
+    public subscript(path path: [Key]) -> Any? {
+        get {
+            return path.reduce(self) { result, key -> Any? in
```
LucianoPAlmeida: We can add checks on that to guard empty paths

Tests/SwiftStdlibTests/DictionaryExtensionsTests.swift
```diff
@@ -78,6 +78,14 @@ final class DictionaryExtensionsTests: XCTestCase {
         XCTAssertEqual(dict, ["testkey": "value"])
     }
 
+    func testSubscriptKeypath() {
+        var json = ["key": ["key1": ["key2": "value"]]]
+
+        XCTAssertEqual(json[path: ["key", "key1", "key2"]] as? String, "value")
```
LucianoPAlmeida: We can add an empty path test case also

LucianoPAlmeida:
@guykogus I agree that it aligns more with the NSObject method, but now I was thinking about that, passing a string with separated by a dot, we will not be able to find keys that actually contains a dot because the split will understand it as another level on the hierarchy. But I don't know, maybe we can escape the dot when we want it as part of the key, but to me now seems safer to receive a path as a list. What do you think?

guykogus:
Yeah, I see your point. I guess the dots for key paths worked for `NSObject`, which just uses it to get property names which can't have a dot, but for Swift Dictionaries it can cause issues.

Sources/Extensions/SwiftStdlib/DictionaryExtensions.swift
calebkleveter: If an empty path is passed in, do we want to return `nil` or `self`?

Sources/Extensions/SwiftStdlib/DictionaryExtensions.swift
```diff
@@ -145,7 +145,49 @@ public extension Dictionary where Key: StringProtocol {
...
+    /// - Note: Value fetching is iterative, while setting is recursive.
+    ///
+    /// - Complexity: O(N), _N_ being the length of the path passed in.
+    ///
+    /// - Parameter path: An array of keys to the desired value.
+    ///
+    /// - Returns: The value for the key-path passed in. `nil` if no value is found.
+    public subscript(path path: [Key]) -> Any? {
+        get {
+            guard path.count > 0 else { return nil }
```
LucianoPAlmeida: Not required, but consider use isEmpty for this check :))

Tests/SwiftStdlibTests/DictionaryExtensionsTests.swift
```diff
@@ -78,6 +78,15 @@ final class DictionaryExtensionsTests: XCTestCase {
         XCTAssertEqual(dict, ["testkey": "value"])
     }
 
+    func testSubscriptKeypath() {
+        var json = ["key": ["key1": ["key2": "value"]]]
+
+        XCTAssertEqual(json[path: []] as? String, nil)
```
LucianoPAlmeida: Also a test case for `["key", "key1"]` returning `["key2": "value"]` can be a good addition :))

Tests/SwiftStdlibTests/DictionaryExtensionsTests.swift
```diff
@@ -78,6 +78,16 @@ final class DictionaryExtensionsTests: XCTestCase {
...
```
guykogus: Is it necessary to cast `as? String`? Are any of these castings necessary?

Sources/Extensions/SwiftStdlib/DictionaryExtensions.swift
```diff
@@ -145,7 +145,49 @@ public extension Dictionary where Key: StringProtocol {
...
+                    result = element
+                } else {
+                    return nil
+                }
+            }
+            return result
+        }
+        set {
+            if let first = path.first {
+                if path.count == 1, let new = newValue as? Value {
```
guykogus: Is this cast necessary?

calebkleveter: Yes. If I remove it, I get `Cannot assign value of type '[Dictionary<Key, Value>.Key : Any]' (aka 'Dictionary<Key, Any>') to type 'Value?'`

Tests/SwiftStdlibTests/DictionaryExtensionsTests.swift
calebkleveter: Yes. Because we are returning type `Any` and `Any` and `String` cannot be compared. If we returned generic type `T`, this would not be necessary.

PR #577, opened 2018-10-04
(Maintenance) Random values

guykogus:
🚀
This is more of a maintenance PR than anything. Just wanted to do some cleaning of usage of the random values/elements. I think this is a bit cleaner (and probably more performant in the case of the random strings).

I also updated the version just so that it points to the current version (it should be 1 ahead of the latest tag, right?).

## Checklist

### DIFF
--- a/Sources/Info.plist
+++ b/Sources/Info.plist
@@ -15,7 +15,7 @@
 	<key>CFBundlePackageType</key>
 	<string>FMWK</string>
 	<key>CFBundleShortVersionString</key>
-	<string>4.4.0</string>
+	<string>4.7.0</string>
 	<key>CFBundleVersion</key>
 	<string>$(CURRENT_PROJECT_VERSION)</string>
 	<key>NSPrincipalClass</key>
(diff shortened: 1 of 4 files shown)


guykogus:
Do I need a changelog entry for this?

LucianoPAlmeida:
@guykogus Since is just implementation details I don't think is required :))

Sources/Info.plist
```diff
@@ -15,7 +15,7 @@
 	<key>CFBundlePackageType</key>
 	<string>FMWK</string>
 	<key>CFBundleShortVersionString</key>
-	<string>4.4.0</string>
+	<string>4.7.0</string>
```
LucianoPAlmeida: I think we are in version 4.6.0

guykogus: 4.6.0 has already been released, so we're working on 4.7.0

guykogus: You never know, considering that Swift 5 may be out in not too long... We'll cross that bridge when we get to it 😄

PR #578, opened 2018-10-04
Added a computed property 'yesterday' for Date and a test function for it.

AlexeiGitH:
## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 # Upcoming release
 
 ### Added
+- **Dictionary**
+  - Added `Dictionary[path:]` subscript for deep fetching/setting nested values. [#574](https://github.com/SwifterSwift/SwifterSwift/pull/573) by [@calebkleveter](https://github.com/calebkleveter)
 ### Changed
 ### Fixed
 - **UIImage**:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,7 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 
 ### Added
 - **Date**
-    -Added  `yesterday` computed property to quickly get yesterday's date to avoid calling `adding(_ :, value:)` function.
+    -Added  `yesterday` computed property to quickly get yesterday's date to avoid calling `adding(_ :, value:)` function. [#578](https://github.com/SwifterSwift/SwifterSwift/pull/578) by [AlexeiGitH](https://github.com/AlexeiGitH)
 - **UIView**
   - Added `ancestorView(where:)` and `ancestorView(withClass:)` to search for a view in the superviews. [#560](https://github.com/SwifterSwift/SwifterSwift/pull/560) by [overovermind](https://github.com/overovermind)
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,7 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 
 ### Added
 - **Date**
-    -Added  `yesterday` computed property to quickly get yesterday's date to avoid calling `adding(_ :value:)` function.
+    -Added  `yesterday` computed property to quickly get yesterday's date to avoid calling `adding(_ :, value:)` function.
 - **UIView**
   - Added `ancestorView(where:)` and `ancestorView(withClass:)` to search for a view in the superviews. [#560](https://github.com/SwifterSwift/SwifterSwift/pull/560) by [overovermind](https://github.com/overovermind)
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 # [v4.6.0](https://github.com/SwifterSwift/SwifterSwift/releases/tag/4.6.0)
 
 ### Added
+- **Date**
+    -Added  `yesterday` computed property to quickly get yesterday's date to avoid calling `adding(_ :value:)` function.
 - **UIView**
   - Added `ancestorView(where:)` and `ancestorView(withClass:)` to search for a view in the superviews. [#560](https://github.com/SwifterSwift/SwifterSwift/pull/560) by [overovermind](https://github.com/overovermind)
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,7 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 
 ### Added
 - **Date**
-    -Added  `yesterday` computed property to quickly get yesterday's date to avoid calling `adding(_ :, value:)` function. [#578](https://github.com/SwifterSwift/SwifterSwift/pull/578) by [AlexeiGitH](https://github.com/AlexeiGitH)
+    - Added `yesterday` computed property to quickly get yesterday's date to avoid calling `adding(_:value:)` function. [#578](https://github.com/SwifterSwift/SwifterSwift/pull/578) by [AlexeiGitH](https://github.com/AlexeiGitH)
 - **UIView**
   - Added `ancestorView(where:)` and `ancestorView(withClass:)` to search for a view in the superviews. [#560](https://github.com/SwifterSwift/SwifterSwift/pull/560) by [overovermind](https://github.com/overovermind)
 ### Fixed
--- a/Sources/Extensions/Foundation/DateExtensions.swift
+++ b/Sources/Extensions/Foundation/DateExtensions.swift
@@ -459,11 +459,11 @@ public extension Date {
 
     /// SwifterSwift: Yesterday date.
     ///
-    ///     var date = Date() // "Oct 3, 2018, 10:57:11"
-    ///     var yesterday = yesterday // "Oct 2, 2018, 10:57:11"
+    ///     let date = Date() // "Oct 3, 2018, 10:57:11"
+    ///     let yesterday = date.yesterday // "Oct 2, 2018, 10:57:11"
     ///
     public var yesterday: Date {
-        return self.addingTimeInterval(-86400.0)
+        return addingTimeInterval(-86400.0)
     }
 
     /// SwifterSwift: UNIX timestamp from date.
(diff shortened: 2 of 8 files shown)


CHANGELOG.md
```diff
@@ -19,6 +19,8 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 # [v4.6.0](https://github.com/SwifterSwift/SwifterSwift/releases/tag/4.6.0)
 
 ### Added
+- **Date**
+    -Added  `yesterday` computed property to quickly get yesterday's date to avoid calling `adding(_ :, value:)` function.
```
calebkleveter: You should reference the PR fo this change here:

```
 -Added  `yesterday` computed property to quickly get yesterday's date to avoid calling `adding(_ :, value:)` function. [#578](https://github.com/SwifterSwift/SwifterSwift/pull/578) by [AlexeiGitH}(https://github.com/AlexeiGitH)
```

AlexeiGitH: Got a little confused here, will fix. Thank you!

Sources/Extensions/Foundation/DateExtensions.swift
```diff
@@ -457,6 +457,15 @@ public extension Date {
         return Calendar.current.timeZone
     }
 
+    /// SwifterSwift: Yesterday date.
+    ///
+    ///     var date = Date() // "Oct 3, 2018, 10:57:11"
+    ///     var yesterday = yesterday // "Oct 2, 2018, 10:57:11"
+    ///
+    public var yesterday: Date {
+        return self.addingTimeInterval(-86400.0)
```
guykogus: Remove `self.`

Sources/Extensions/Foundation/DateExtensions.swift
```diff
@@ -457,6 +457,15 @@ public extension Date {
         return Calendar.current.timeZone
     }
 
+    /// SwifterSwift: Yesterday date.
+    ///
+    ///     var date = Date() // "Oct 3, 2018, 10:57:11"
+    ///     var yesterday = yesterday // "Oct 2, 2018, 10:57:11"
```
guykogus: Why use `var`s in the example if you're not going to mutate them?
Also, this should be `= date.yesterday`

Tests/FoundationTests/DateExtensionsTests.swift
```diff
@@ -944,6 +944,10 @@ final class DateExtensionsTests: XCTestCase {
         XCTAssertTrue(randomDate.isBetween(sinceDate, toDate, includeBounds: true))
     }
 
+    func testYesterday() {
+        let yesterday = Date().yesterday
+        XCTAssert(yesterday.isInYesterday)
```
guykogus: In the right timezone on the right date at the right time, this test could actually fail since it's checking the calendar date components. I mean, the chances of that happening are virtually zero, but [here's some fun reading for you](https://stackoverflow.com/questions/24088730/why-nsdateformatter-is-returning-null-for-a-19-10-2014-in-a-brazilian-time-zone) 😅 (TL;DR Imagine that you run this test at 00:30 where the clock was changed due to daylight savings and now 24 hours ago was actually 25 hours ago and you've gone back 2 days, or something weird like that. It may not be possible but you never know...)
Anyway, I think it might be better to check the time interval difference.

Sources/Extensions/Foundation/DateExtensions.swift
guykogus: Actually if the whole purpose of this is to be a shorthand for `adding(_:value:)` why don't you use that?

AlexeiGitH: In the Guidelines it says "Do not use an existing SwifterSwift extension inside another SwifterSwift extension" so I didn't want to use `adding(_:value:)`

Sources/Extensions/Foundation/DateExtensions.swift
```diff
@@ -457,6 +457,15 @@ public extension Date {
         return Calendar.current.timeZone
     }
 
+    /// SwifterSwift: Yesterday date.
+    ///
+    ///     let date = Date() // "Oct 3, 2018, 10:57:11"
+    ///     let yesterday = date.yesterday // "Oct 2, 2018, 10:57:11"
+    ///
+    public var yesterday: Date {
+        return addingTimeInterval(-86400.0)
```
LucianoPAlmeida: Hey @AlexeiGitH :))
Using `Calendar.current.date(byAdding: .day, value: -1, to: self)` is also a implementation option. To me seems more correct. What you think?

AlexeiGitH: So it seems like using `Calendar.current.date(byAdding: .day, value: -1, to: self)` covers the DST case, and instead of returning a nil date it just gives back a Date that was 23h ago (-82800 seconds), while `addingTimeInterval(_:)` will return a Date that was exactly 24h ago, even if it didn't exists (like in your example). 
I guess it may be more useful to return a real date, which will always give a 24h difference, except for the explained DST scenario.
May I update the method and use `Calendar.current.date(byAdding: .day, value: -1, to: self)`?

guykogus: `addingTimeInterval(_:)` doesn't return a non-existent date. It very much exists, the only difference is that the hour of the day may not be the same.
In my example March 25 at 2:00AM doesn't exist, so when you go back with `addingTimeInterval` it skips back to 1:00AM, which is exactly 24 hours, whereas using `date(byAdding...)` moves it forward to 3:00AM.

My concern with using `date(byAdding...)` is that it _should_ return `nil` (which is why the return value is optional) but doesn't. Instead, it manipulates the hour in order to not return `nil`. While I think that it's a good thing, it's unreliable considering the API and purpose. Does that make sense? It's why I prefer `addingTimeInterval(_:)`.

AlexeiGitH: Got it @guykogus. I think I will leave it as it is with `addingTimeInterval(_:)`.

SD10:

Sources/Extensions/Foundation/DateExtensions.swift
MosheBerman: Using a fixed second constant is not correct. A "real" date is not always 24 hours. See the very first entry on: http://yourcalendricalfallacyis.com

> `addingTimeInterval(_:)` doesn't return a non-existent date. It very much exists, the only difference is that the hour of the day may not be the same.

I would argue that returning yesterday without the same hour component is expected behavior in daylight savings and "leap unit" cases. The computed property is not called `sameTimeYesterday`, but rather `yesterday`. If you wanted `sameTimeYesterday`, you can still use `dateByAdding(...` and then use `DateComponents` to manipulate the sub-day components. (hour, minute, etc.) 

> My concern with using `date(byAdding...)` is that it should return `nil` (which is why the return value is optional) but doesn't. 

I don't think `date(byAdding:)` should return `nil` unless it absolutely has to. That's not the reason why the return value is optional. My guess as to why this API is optional is because a date might not be expressible in a given calendar system. (A really old Julian date on a Gregorian calendar system, a 13th month on a 12 month calendar system, etc.)

The "missing" hour isn't missing. Consider an absolute moment in time as an offset from some epoch. We usually refer to a date by how many seconds, hours, etc have passed since that time. The way daylight savings and leap-years work is we (somewhat arbitrarily) adjust how we describe that offset by pretending that some time was added or subtracted even though it has not.

> Instead, it manipulates the hour in order to not return `nil`. While I think that it's a good thing, it's unreliable considering the API and purpose.

This is deterministic behavior, so I don't think it's completely accurate to call it unreliable. Given an input that spans a leap-unit this API will always account for "leapness". Given any other day, it will behave as a constant time would. 

The purpose of this API is to account for DST and leap units, so that developers on Apple platforms don't have to. By writing this API this way, you are transferring that responsibility back to developers and accomplishing the exact opposite of what this library is intended for.

guykogus: > If you wanted `sameTimeYesterday`, you can still use `dateByAdding(...)`

But we don't want the same time yesterday because that may not exist. Whether you use the `Calendar` or `dateByAdding(...)` you're going to get back a different time yesterday (with DST or leap units).

I guess you're right that since a day is not always 24 hours, we shouldn't expect the end result to be 24 hours different, but jumping back by a fixed 24 hours is as "consistent" as we can get. If you want to use the `Calendar` to go back a day, which calendar do you use? Do we stick to the Gregorian? Do we use the system device's calendar?

I think it's also important to ask what this function will be used for. My guess is that generally it will be something like, "Let's get all the values since yesterday", so we run something like:
```Swift
let now = Date()
let results = getValues(from: now.yesterday, to: now)
```
If we were collecting these results every day and used the `Calendar` to jump back each day, the results would spike or dip over DST/leap units, which would be inconsistent.

There are potential complications regardless of what solution we would've gone for. In the end I was of the opinion that this way was simpler. If you can think of real-world examples where this will cause people problems I'm willing to be convinced otherwise.

davedelong: It is incorrect to use `-86400`. It's quite easy to construct a `Date` such that, when you add `-86400` to it, you get a `Date` that is _two_ days ago.

For example: in 4 days (Mar 10th), the United States will perform its spring-time Daylight Saving Time transition and "spring forward" an hour. In this case, Mar 10th will only be 23 hours long. Thus, if I had a `Date` that referred to March 11th at 12:01 AM and I added `-86400` seconds to it, I would get March 9th at 11:01 PM. This is, semantically, not what your method is trying to convey.

The proper way to do this, as has been suggested, is to use the method on `Calendar` to add -1 days to the day. Adding time intervals is fundamentally a semantically different calculation.

guykogus: Thanks @davedelong, I finally managed to find a way to prove it in the code.
```Swift
let timeZone = TimeZone(identifier: "America/Denver")!

let formatter = ISO8601DateFormatter()
formatter.timeZone = timeZone
let date = formatter.date(from: "2019-03-11T00:00:00-06:00")! // 11 Mar 2019 at 00:00
formatter.string(from: date) // 2019-03-11T00:00:00-06:00 (👆, obviously)

var calendar = Calendar.current
calendar.timeZone = timeZone
let yesterdayCalendar = calendar.date(byAdding: .day, value: -1, to: date)! // 10 Mar 2019 at 00:00
formatter.string(from: yesterdayCalendar) // 2019-03-10T00:00:00-07:00
yesterdayCalendar.timeIntervalSince(date) // -82800

let yesterday24Hours = date.addingTimeInterval(-86400) // 9 Mar 2019 at 23:00
formatter.string(from: yesterday24Hours) // 2019-03-09T23:00:00-07:00
```
I'm now in agreement, this should be changed :)

guykogus: Opened https://github.com/SwifterSwift/SwifterSwift/issues/640

PR #579, opened 2018-10-07
Removed self. when not needed

staffler-xyz:
Removed "self." calls where not needed. According to most code style guides (for example https://github.com/raywenderlich/swift-style-guide#use-of-self) they should be avoided where possible. What is your opinion about this?

I didn't add a changelog because it's only a style change (or should I?).

## Checklist

### DIFF
--- a/Sources/Extensions/AppKit/NSViewExtensions.swift
+++ b/Sources/Extensions/AppKit/NSViewExtensions.swift
@@ -132,7 +132,7 @@ extension NSView {
     ///
     /// - Parameter subviews: array of subviews to add to self.
     public func addSubviews(_ subviews: [NSView]) {
-        subviews.forEach({self.addSubview($0)})
+        subviews.forEach({addSubview($0)})
     }
 
     /// SwifterSwift: Remove all subviews in view.
--- a/Sources/Extensions/AppKit/NSViewExtensions.swift
+++ b/Sources/Extensions/AppKit/NSViewExtensions.swift
@@ -132,12 +132,12 @@ extension NSView {
     ///
     /// - Parameter subviews: array of subviews to add to self.
     public func addSubviews(_ subviews: [NSView]) {
-        subviews.forEach({addSubview($0)})
+        subviews.forEach { addSubview($0) }
     }
 
     /// SwifterSwift: Remove all subviews in view.
     public func removeSubviews() {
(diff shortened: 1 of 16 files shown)


Sources/Extensions/AppKit/NSViewExtensions.swift
```diff
@@ -132,7 +132,7 @@ extension NSView {
     ///
     /// - Parameter subviews: array of subviews to add to self.
     public func addSubviews(_ subviews: [NSView]) {
-        subviews.forEach({self.addSubview($0)})
+        subviews.forEach({addSubview($0)})
```
LucianoPAlmeida: I know is not the purpose of the PR but can you also fix here to use trailing syntax and adding space between like `subviews.forEach { addSubview($0) }`? :))

Sources/Extensions/UIKit/UIViewExtensions.swift
```diff
@@ -280,7 +280,7 @@ public extension UIView {
     ///
     /// - Parameter subviews: array of subviews to add to self.
     public func addSubviews(_ subviews: [UIView]) {
-        subviews.forEach({ self.addSubview($0) })
+        subviews.forEach({ addSubview($0) })
```
LucianoPAlmeida: Here too :))

Sources/Extensions/UIKit/UITextFieldExtensions.swift
```diff
@@ -152,9 +152,9 @@ public extension UITextField {
     public func addPaddingLeftIcon(_ image: UIImage, padding: CGFloat) {
         let imageView = UIImageView(image: image)
         imageView.contentMode = .center
-        self.leftView = imageView
-        self.leftView?.frame.size = CGSize(width: image.size.width + padding, height: image.size.height)
-        self.leftViewMode = UITextField.ViewMode.always
+        leftView = imageView
+        leftView?.frame.size = CGSize(width: image.size.width + padding, height: image.size.height)
+        leftViewMode = UITextField.ViewMode.always
```
guykogus: Can this be changed to `leftViewMode = .always`?

guykogus:
Very happy to see this! Good stuff :) Just a few comments left here for you.

staffler-xyz:
sure, changes from comments added :)

PR #582, opened 2018-10-12
Add `BidirectionalCollection[offset:]` subscript for getting element with negative index.

jianstm:
It allows you to access elements in a more intuitive way. 

```swift
let arr = [1, 2, 3, 4, 5]
arr[-1] // -> 5
arr[-2] // -> 4
```

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
   - Added `Dictionary[path:]` subscript for deep fetching/setting nested values. [#574](https://github.com/SwifterSwift/SwifterSwift/pull/573) by [@calebkleveter](https://github.com/calebkleveter)
 - **UIColor**
   - Added `whatsApp` color constant. [#581](https://github.com/SwifterSwift/SwifterSwift/pull/581) by [staffler-xyz](https://github.com/staffler-xyz)
-  - Added `Collection[nth:]` subscript for getting element with negative index.
+  - Added `Collection[nth:]` subscript for getting element with negative index. [#582](https://github.com/SwifterSwift/SwifterSwift/pull/582) by [jianstm](https://github.com/jianstm)
 ### Changed
 ### Fixed
 - **UIImage**:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
   - Added `Dictionary[path:]` subscript for deep fetching/setting nested values. [#574](https://github.com/SwifterSwift/SwifterSwift/pull/573) by [@calebkleveter](https://github.com/calebkleveter)
 - **UIColor**
   - Added `whatsApp` color constant. [#581](https://github.com/SwifterSwift/SwifterSwift/pull/581) by [staffler-xyz](https://github.com/staffler-xyz)
-  - Added `Collection[nth:]` subscript for getting element with negative index. [#582](https://github.com/SwifterSwift/SwifterSwift/pull/582) by [jianstm](https://github.com/jianstm)
+  - Added `BidirectionalCollection[offset:]` subscript to get element with negative offset. [#582](https://github.com/SwifterSwift/SwifterSwift/pull/582) by [jianstm](https://github.com/jianstm)
 ### Changed
 ### Fixed
 - **UIImage**:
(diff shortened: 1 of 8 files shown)


Sources/Extensions/SwiftStdlib/CollectionExtensions.swift
```diff
@@ -122,6 +122,21 @@ public extension Collection where Index == Int {
...
+    /// negative, the nth element from the end will be returned. If index is out
+    /// of the indices, nil will be returned.
+    ///
+    ///        let arr = [1, 2, 3, 4, 5]
+    ///        arr[nth: 1] -> 2
+    ///        arr[nth: -2] -> 4
+    ///        arr[nth: -6] -> nil
+    ///
+    /// - Parameter index: The position of the element to access.
+    public subscript(nth index: Index) -> Element? {
```
guykogus: `nth` is not descriptive enough for a parameter name. I'm not sure what would be a better name, maybe `reverseIndex`, although you're using the index for both positive and negative values so `bidirectionalIndex`, although that may be a bit of a mouthful. Can you think of any suggestions?

guykogus: The regular `subscript` doesn't return an `Optional`, so I think it would be consistent to do the same. If it's out of bounds then let the program crash.

guykogus:
You can kind of make this generic by doing it like this:
```Swift
extension Collection {
    public subscript(nth offset: Int) -> Element {
        let index = offset >= 0 ? startIndex : endIndex
        return self[indices.index(index, offsetBy: offset)]
    }
}
```

Sources/Extensions/SwiftStdlib/CollectionExtensions.swift
jianstm: @guykogus I'm not sure, since `subscript` doesn't have auto-complete, i think the shorter the method name, the better.

jianstm: @guykogus OK. I personally like javascript style, we know what developer want to do, crash makes people annoyed. But you are right, this is Swift.

jianstm:
@guykogus Hi, this feature is very common in other languages(ruby, python). It allows you to access elements in a more intuitive way. For example, if you want to get the third element from the end, you can write `arr[-3]` instead of `arr[length - 3]`.

jianstm:
> You can kind of make this generic by doing it like this:
> 
> ```swift
> extension Collection {
>     public subscript(nth offset: Int) -> Element {
>         let index = offset >= 0 ? startIndex : endIndex
>         return self[indices.index(index, offsetBy: offset)]
>     }
> }
> ```

Unless it's `BidirectionalCollection`

Sources/Extensions/SwiftStdlib/CollectionExtensions.swift
```diff
@@ -185,3 +185,20 @@ public extension Collection where Element: FloatingPoint {
     }
 
 }
+
+public extension BidirectionalCollection {
```
LucianoPAlmeida: Can you move the `BidirectionalCollection` extension to its own file?

LucianoPAlmeida:
cc @SD10 @omaralbeik 
Pinging for your review :))

Sources/Extensions/SwiftStdlib/BidirectionalCollectionExtensions.swift
```diff
@@ -0,0 +1,25 @@
...
+    ///        let arr = [1, 2, 3, 4, 5]
+    ///        arr[offset: 1] -> 2
+    ///        arr[offset: -2] -> 4
+    ///
+    /// - Parameter distance: The distance to offset.
+    public subscript(offset distance: Int) -> Element {
+        let index = distance >= 0 ? startIndex : endIndex
+        return self[indices.index(index, offsetBy: distance)]
+    }
+}
```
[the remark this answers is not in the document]
jianstm: Added, very thanks for your help!

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
   - Added `Dictionary[path:]` subscript for deep fetching/setting nested values. [#574](https://github.com/SwifterSwift/SwifterSwift/pull/573) by [@calebkleveter](https://github.com/calebkleveter)
 - **UIColor**
   - Added `whatsApp` color constant. [#581](https://github.com/SwifterSwift/SwifterSwift/pull/581) by [staffler-xyz](https://github.com/staffler-xyz)
+- **BidirectionalCollection**
+  - Added `BidirectionalCollection[offset:]` subscript to get element with negative offset. [#582](https://github.com/SwifterSwift/SwifterSwift/pull/582) by [jianstm](https://github.com/jianstm)
```
guykogus: This should say:
> Added \`subscript[offset:]\` to get...

jianstm: @guykogus Done!

PR #588, opened 2018-10-15
Add contains(_:times:) to Sequence

MrLotU:
🚀
Adds `contains(_:times:)` to Sequence to check if a Sequence contains specified element exactly `n` times

## Checklist

### DIFF
--- a/Sources/Extensions/SwiftStdlib/SequenceExtensions.swift
+++ b/Sources/Extensions/SwiftStdlib/SequenceExtensions.swift
@@ -180,6 +180,26 @@ public extension Sequence where Element: Equatable {
         return true
     }
 
+    /// SwifterSwift: Check if sequence contains element exactly `n` times
+    ///
+    ///     [1, 2, 3, 1].contains(1, times: 1) -> False
+    ///     [1, 2, 3, 1].contains(1, times: 2) -> True
+    ///     ["h", "e", "l", "l", "o"].contains("l", times: 2) -> true
+    ///
+    /// - Parameter element: element to check
+    /// - Parameter n: amount of times element should appear in the sequence
+    /// - Returns: true if sequence contains element n times
+    public func contains(_ element: Element, times n: Int) -> Bool {
+        var matches = 0
+        
+        for item in self {
+            if item == element {
+                matches += 1
+            }
+        }
+        
+        return matches == n
+    }
 }
 
 public extension Sequence where Element: Hashable {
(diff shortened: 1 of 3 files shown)


MrLotU:
#589 Implements this for Strings

guykogus:
I have to agree with @LucianoPAlmeida's points.
I'm also wondering why you have this in `Sequence` and not in `Collection`? That way it would be more generic and you wouldn't have to copy it for `String`, like in https://github.com/SwifterSwift/SwifterSwift/pull/589

Sources/Extensions/SwiftStdlib/SequenceExtensions.swift
```diff
@@ -180,6 +180,25 @@ public extension Sequence where Element: Equatable {
...
+    ///     [1, 2, 3, 1].contains(1, times: 1) -> false
+    ///     [1, 2, 3, 1].contains(1, times: 2) -> true
+    ///     ["h", "e", "l", "l", "o"].contains("l", times: 2) -> true
+    ///
+    /// - Parameters:
+    ///     - element: element to check
+    ///     - times: amount of times element should appear in the sequence
+    /// - Returns: true if sequence contains element `times` times
+    public func contains(_ element: Element, times: Int) -> Bool {
+        var matches = 0
```
guykogus: You can break this all down to 1 line:
```Swift
return filter { $0 == element }.count == times
```

calebkleveter:
The reason you have a different version for `String` is because in this PR where are checking a single element in the sequence, for `String` we are checking multiple elements that must all be adjacent and in the correct order. Maybe the generic implementation should allow for that? I'm not sure what the interface would look like.

Also, @guykogus. From my experience of the standard library, `Sequence` is more generic than `Collection`, not the other way around. Is there a reason you say `Collection` is more generic?

guykogus:
🤦‍♂️ sorry, of course `Sequence` is more generic.
I see the difference now in `String`. I'll comment on it for you there.

Sources/Extensions/SwiftStdlib/SequenceExtensions.swift
guykogus: Actually, `filter` is a bad choice as it will create another array. Better to use `reduce`. I.e.
```Swift
public func count(_ element: Element) -> Int {
    return reduce(0) { ($1 == element) ? $0 + 1 : $0 }
}
```

guykogus:
Closing due to inactivity. Feel free to reopen 😄

PR #589, opened 2018-10-15
Add `contains(_:times:)` to String

MrLotU:
🚀 
Adds contains(_:times:) to String to check if a String contains specified substring exactly `times` times
## Checklist

### DIFF
--- a/Sources/Extensions/SwiftStdlib/StringExtensions.swift
+++ b/Sources/Extensions/SwiftStdlib/StringExtensions.swift
@@ -677,6 +677,34 @@ public extension String {
         return range(of: string) != nil
     }
     #endif
+    
+    /// SwifterSwift: Check if string contains substring exactly `times` times
+    ///
+    ///     "hellohello".contains("hello", times: 2) -> true
+    ///     "hello".contains("h", times: 2) -> false
+    ///     "hello.contains("l", times: 2) -> true
+    /// - Parameters:
+    ///     - string: substring to search for
+    ///     - times: amount of times substring should appear in the string
+    /// - Returns: true if string contains substring `times` times
+    public func contains<S>(_ string: S, times: Int) -> Bool where S: StringProtocol {
+        var index = self.startIndex
+        var matchEnd = self.index(self.startIndex, offsetBy: string.count - 1)
+        var matches = 0
+        
+        while matchEnd < self.endIndex {
+            if self[index] == string.first {
+                if self[index...matchEnd] == string {
+                    matches += 1
+                }
+            }
+            
+            index = self.index(index, offsetBy: 1)
+            matchEnd = self.index(index, offsetBy: string.count - 1)
+        }
+        
+        return matches == times
+    }
 
     #if canImport(Foundation)
     /// SwifterSwift: Count of substring in string.
--- a/Sources/Extensions/SwiftStdlib/StringExtensions.swift
+++ b/Sources/Extensions/SwiftStdlib/StringExtensions.swift
@@ -677,7 +677,7 @@ public extension String {
         return range(of: string) != nil
     }
     #endif
-    
+
     /// SwifterSwift: Check if string contains substring exactly `times` times
     ///
     ///     "hellohello".contains("hello", times: 2) -> true
@@ -691,18 +691,18 @@ public extension String {
         var index = self.startIndex
         var matchEnd = self.index(self.startIndex, offsetBy: string.count - 1)
         var matches = 0
-        
+
         while matchEnd < self.endIndex {
             if self[index] == string.first {
                 if self[index...matchEnd] == string {
                     matches += 1
                 }
             }
-            
+
             index = self.index(index, offsetBy: 1)
             matchEnd = self.index(index, offsetBy: string.count - 1)
         }
-        
+
         return matches == times
     }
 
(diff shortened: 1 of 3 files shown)


MrLotU:
#588 Implements this for Sequence

Sources/Extensions/SwiftStdlib/StringExtensions.swift
```diff
@@ -678,6 +678,34 @@ public extension String {
     }
     #endif
 
+    /// SwifterSwift: Check if string contains substring exactly `times` times
+    ///
+    ///     "hellohello".contains("hello", times: 2) -> true
+    ///     "hello".contains("h", times: 2) -> false
+    ///     "hello.contains("l", times: 2) -> true
```
calebkleveter: Looks like you missed a quote for the string value.

MrLotU: OOPS! Let me add that in! Thanks 👍

Sources/Extensions/SwiftStdlib/StringExtensions.swift
```diff
@@ -678,6 +678,34 @@ public extension String {
...
+    ///     - times: amount of times substring should appear in the string
+    /// - Returns: true if string contains substring `times` times
+    public func contains<S>(_ string: S, times: Int) -> Bool where S: StringProtocol {
+        var index = self.startIndex
+        var matchEnd = self.index(self.startIndex, offsetBy: string.count - 1)
+        var matches = 0
+
+        while matchEnd < self.endIndex {
+            if self[index] == string.first {
+                if self[index...matchEnd] == string {
```
guykogus: For long strings this would be incredibly inefficient, since you'd be creating copies of the string buffer, unless there's some optimisation going on in with `Substring` in which case it's not too bad, but I recommend limiting the length just to be safe.
```Swift
func count<S>(_ string: S) -> Int where S: StringProtocol {
    guard !string.isEmpty else { return 0 }

    var matches = 0

    var matchStart = startIndex
    while distance(from: matchStart, to: endIndex) >= string.count {
        if self[matchStart..<index(matchStart, offsetBy: string.count)] == string {
            matches += 1
        }
        matchStart = index(matchStart, offsetBy: 1)
    }

    return matches
}
```

LucianoPAlmeida: I think Substring and slices are just references for the underlying storage, I think there's no copy for that :))

LucianoPAlmeida:
@guykogus I'm not sure about using `NSRegularExpression` that is a `Foundation` class into a String extension inside our stdlib subspec. I think it adds Foundation dependency into the stdlib module. 
Maybe it is not so bad and we can just use canImport statement since we have few already :)

Sources/Extensions/SwiftStdlib/StringExtensions.swift
guykogus: Oh and I didn't realise that `matchEnd` is set as being that length anyway. I swear there's something wrong with my brain lately 😬

OK, so if you want, @MrLotU, you can skip changing the whole thing to my way, but at the very least I suggest getting rid of `if self[index] == string.first`, since you're then following it by comparing the strings. My guess is that `==` will go character by character anyway and check for equality, so I think it's redundant.

guykogus:
@LucianoPAlmeida you're right, but we already have some `Foundation` classes in `StringExtensions.swift` because `Foundation` itself contains `String` extensions, so they work closely together.

guykogus:
Hmm, now I realise that it's limiting the availability if we move it to `Foundation` when there's no need for it.
Anyway, like I mentioned the results from using regular expressions doesn't actually match the desired results.

Sources/Extensions/SwiftStdlib/StringExtensions.swift
calebkleveter: I decided to test the speed on this method with and without the `if self[index] == string.first`.

```swift
func testStringContainsCountSpeed() {
    let lorum =
        "Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua." +
        "Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat." +
        "Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur."
    let str = String(repeating: lorum, count: 10_000)
        
    XCTAssert(str.contains("reprehenderit", times: 10_000))
        
    measure {
        _ = str.contains("reprehenderit", times: 10_000)
    }
}
```
The string tested ended up being 3,320,000 characters long. Without the additional `if` statement, it took `1.899` seconds, but with it it was `1.409` seconds. I believe this has to do with matching a range being an operation of `O(n)`, where `n` is the length of the range, where checking a single index is `O(1)`. If we get `false` on the single character, we can skip a longer and unnecessary operation.

If you still want to get rid of that `if` statement, that's fine. I just thought I'd post my findings.

guykogus:
Sorry about the multiple reviews. GitHub is having some serious issues today...

guykogus:
Closing due to inactivity. Feel free to reopen 😄

PR #590, opened 2018-10-17
Update `podspec` to make the group paths of SwifterSwift correct in Pods project

dklinzh:
## Checklist

Fix up `source_files` in `podspec` to remove the duplicate reference of source files that make the group paths of SwifterSwift correct in `Pods` project. I also create an example project with Cocoapods installation and can change `Podfile` with different pod situations for test.
```
# The correct group paths of SwifterSwift in Development Pods.
pod 'SwifterSwift', :path => '../../' 
```

```
# The group paths of SwifterSwifte in Pods before version 4.6.0 is incorrect. The source files are not contained in their own subspec groups.
pod 'SwifterSwift', '4.6.0'
```

```
# The correct group paths of SwifterSwift in Pods as same as that in Development Pods. 
pod 'SwifterSwift', :git => 'https://github.com/dklinzh/SwifterSwift.git', :branch => 'pod_install'
```


dklinzh:
@LucianoPAlmeida 
Yes, I think so. I may remove the example project and update the PR if needed.

guykogus:
Agreed, I don't think the example project is necessary.

SwifterSwift.podspec
```diff
@@ -35,12 +35,14 @@ Pod::Spec.new do |s|
 
   # UIKit Extensions
   s.subspec 'UIKit' do |sp|
-    sp.source_files  = 'Sources/Extensions/UIKit/*.swift', 'Sources/Extensions/Shared/ColorExtensions.swift'
+    sp.dependency 'SwifterSwift/Shared/Color'
```
guykogus: Should/Could this (and the other instances) be
```
sp.dependency `Shared`
```
?
That way if we want to add more shared dependencies they would only need to be updated in one place.

dklinzh: I am afraid that it can't be supported by the current version of Cocoapods. [Here](https://guides.cocoapods.org/syntax/podspec.html#subspec) is the usage of `supspec` with 'dependency'. And I revoke the `subspec 'Shared/Color'` and use its parent  `subspec 'Shared'` instead.

PR #592, opened 2018-10-20
Fix unit test fail of testing date beginning

cruisediary:
👻 Fix unit test failed cause by timezone problem not new feature

## Checklist

### DIFF
--- a/Tests/FoundationTests/DateExtensionsTests.swift
+++ b/Tests/FoundationTests/DateExtensionsTests.swift
@@ -638,12 +638,12 @@ final class DateExtensionsTests: XCTestCase {
         XCTAssertNotNil(beginningOfWeek)
         XCTAssertEqual(date.beginning(of: .weekOfMonth)?.day, beginningOfWeek?.day)
 
-        let beginningOfMonth = Date(year: 2016, month: 8, day: 1, hour: 5)
+        let beginningOfMonth = Date(timeZone: Date().timeZone, year: 2016, month: 8, day: 1, hour: 5)
         XCTAssertNotNil(date.beginning(of: .month))
         XCTAssertNotNil(beginningOfMonth)
         XCTAssertEqual(date.beginning(of: .month)?.day, beginningOfMonth?.day)
 
-        let beginningOfYear = Date(year: 2016, month: 1, day: 1, hour: 5)
+        let beginningOfYear = Date(timeZone: Date().timeZone, year: 2016, month: 1, day: 1, hour: 5)
         XCTAssertNotNil(date.beginning(of: .year))
         XCTAssertNotNil(beginningOfYear)
         XCTAssertEqual(date.beginning(of: .year)?.day, beginningOfYear?.day)


cruisediary:
@jianstm I used `Date().timeZone` because it is used in some tests already
```swift
func testNewDateFromComponenets() { ... }
func testTimezone() { ... }
```

guykogus:
Honestly, I think that the `Date.timeZone` property shouldn't exist at all. It doesn't make sense semantically. `Date` objects are timezone-agnostic.
@jianstm is right, please use `NSTimeZone.default` instead.

Another option, which might not be a bad idea, is to replace instances of `TimeZone.current` with `NSTimeZone.default`. If you do this in `Date.init(calendar:timeZone:...)` and since the value is being set in the `setup` function, that value will then be used correctly during the tests.

guykogus:
Honestly, I think that the `Date.timeZone` property shouldn't exist at all. It doesn't make sense semantically. `Date` objects are timezone-agnostic.
@jianstm is right, please use `NSTimeZone.default` instead.

Another option, which might not be a bad idea, is to replace instances of `TimeZone.current` with `NSTimeZone.default`. If you do this in `Date.init(calendar:timeZone:...)` and since the value is being set in the `setup` function, that value will then be used correctly during the tests.

guykogus:
Honestly, I think that the `Date.timeZone` property shouldn't exist at all. It doesn't make sense semantically. `Date` objects are timezone-agnostic.
@jianstm is right, please use `NSTimeZone.default` instead.

Another option, which might not be a bad idea, is to replace instances of `TimeZone.current` with `NSTimeZone.default`. If you do this in `Date.init(calendar:timeZone:...)` and since the value is being set in the `setup` function, that value will then be used correctly during the tests.

guykogus:
@jianstm is right, please use `NSTimeZone.default` instead.

Honestly, I think that the `Date.timeZone` property shouldn't exist at all. It doesn't make sense semantically. `Date` objects are timezone-agnostic.

Another option, which might not be a bad idea, is to replace instances of `TimeZone.current` with `NSTimeZone.default`. If you do this in `Date.init(calendar:timeZone:...)` and since the value is being set in the `setup` function, that value will then be used correctly during the tests. It will also allow people in other situations to benefit from setting a default timezone.

guykogus:
Sorry about the multiple comments. GitHub is having some serious issues today...

LucianoPAlmeida:
The API is very unstable since yesterday 😭

cruisediary:
@jianstm @guykogus Updated 🏄 
Use `NSTimeZone.default` instead `Timezone.current` or `Date().timezone`

guykogus:
@cruisediary seems `pod lib lint` failed for some unknown reason. It does this sometimes. Try and re-trigger it, if you can. If not, just make an empty git commit:
```sh
git commit --amend
git push -f
```
That will trigger it too.

Tests/FoundationTests/DateExtensionsTests.swift
```diff
@@ -459,7 +459,7 @@ final class DateExtensionsTests: XCTestCase {
     }
 
     func testTimezone() {
-        XCTAssertEqual(Date().timeZone, Calendar.current.timeZone)
+        XCTAssertEqual(NSTimeZone.default, Calendar.current.timeZone)
```
guykogus: Please revert this as it's testing the actual extension.
```suggestion
        XCTAssertEqual(Date().timeZone, Calendar.current.timeZone)
```

PR #593, opened 2018-10-22
Adding convenience method to present a popover.

marcocapano:
This PR adds a methods to present a view controller as a popover, which usually requires multiple lines of code.
I'm not sure the test is exhaustive enough, so I'm happy to receive suggestions.
## Checklist
- [ x] New extensions are written in Swift 4.
- [x ] I have added tests for new extensions, and they passed.

### DIFF
--- a/Sources/Extensions/UIKit/UIViewControllerExtensions.swift
+++ b/Sources/Extensions/UIKit/UIViewControllerExtensions.swift
@@ -96,5 +96,29 @@ public extension UIViewController {
         removeFromParent()
         view.removeFromSuperview()
     }
+
+    /// SwifterSwift: Helper method to present a UIViewController as a popover.
+    ///
+    /// - Parameters:
+    ///   - popoverContent: the view controller to add as a popover.
+    ///   - sourcePoint: the point in which to anchor the popover.
+    ///   - size: the size of the popover. Default uses the popover preferredContentSize.
+    ///   - delegate: the popover's presentationController delegate. Default is nil.
+    ///   - completion: The block to execute after the presentation finishes. Default is nil.
+    public func presentPopover(_ popoverContent: UIViewController, sourcePoint: CGPoint, size: CGSize? = nil, delegate: UIPopoverPresentationControllerDelegate? = nil, completion: ( () -> Void)? = nil) {
+
+        popoverContent.modalPresentationStyle = .popover
+
+        if let size = size {
+            popoverContent.preferredContentSize = size
+        }
+
+        let popoverPresentationVC = popoverContent.popoverPresentationController!
+        popoverPresentationVC.sourceView = self.view
+        popoverPresentationVC.sourceRect = CGRect(origin: sourcePoint, size: .zero)
+        popoverPresentationVC.delegate = delegate
+
+        self.present(popoverContent, animated: true, completion: completion)
+    }
 }
 #endif
--- a/Tests/UIKitTests/UIViewControllerExtensionsTests.swift
+++ b/Tests/UIKitTests/UIViewControllerExtensionsTests.swift
@@ -119,6 +119,17 @@ final class UIViewControllerExtensionsTests: XCTestCase {
 
         XCTAssertNil(childViewController.parent)
     }
+    
+    func testPresentPopover() {
+        let popover = UIViewController()
+        let vc = UIViewController()
+        
+        vc.presentPopover(popover, sourcePoint: vc.view.center) {
+            XCTAssertEqual(vc.presentedViewController, popover)
+            XCTAssertEqual(popover.presentingViewController, vc)
+            XCTAssertEqual(popover.modalPresentationStyle, .popover)
+        }
+    }
 
 }
 #endif


Sources/Extensions/UIKit/UIViewControllerExtensions.swift
```diff
@@ -96,5 +96,29 @@ public extension UIViewController {
...
+    public func presentPopover(_ popoverContent: UIViewController, sourcePoint: CGPoint, size: CGSize? = nil, delegate: UIPopoverPresentationControllerDelegate? = nil, completion: ( () -> Void)? = nil) {
+
+        popoverContent.modalPresentationStyle = .popover
+
+        if let size = size {
+            popoverContent.preferredContentSize = size
+        }
+
+        let popoverPresentationVC = popoverContent.popoverPresentationController!
+        popoverPresentationVC.sourceView = self.view
```
LucianoPAlmeida: Self in this context is not needed :))

Sources/Extensions/UIKit/UIViewControllerExtensions.swift
```diff
@@ -96,5 +96,29 @@ public extension UIViewController {
...
+    ///   - completion: The block to execute after the presentation finishes. Default is nil.
+    public func presentPopover(_ popoverContent: UIViewController, sourcePoint: CGPoint, size: CGSize? = nil, delegate: UIPopoverPresentationControllerDelegate? = nil, completion: ( () -> Void)? = nil) {
+
+        popoverContent.modalPresentationStyle = .popover
+
+        if let size = size {
+            popoverContent.preferredContentSize = size
+        }
+
+        let popoverPresentationVC = popoverContent.popoverPresentationController!
```
LucianoPAlmeida: Not sure, but this property can be nil on iPhones I think. 
As far as I remember popovers are iPad only

Sources/Extensions/UIKit/UIViewControllerExtensions.swift
```diff
@@ -96,5 +96,29 @@ public extension UIViewController {
...
+
+    /// SwifterSwift: Helper method to present a UIViewController as a popover.
+    ///
+    /// - Parameters:
+    ///   - popoverContent: the view controller to add as a popover.
+    ///   - sourcePoint: the point in which to anchor the popover.
+    ///   - size: the size of the popover. Default uses the popover preferredContentSize.
+    ///   - delegate: the popover's presentationController delegate. Default is nil.
+    ///   - completion: The block to execute after the presentation finishes. Default is nil.
+    public func presentPopover(_ popoverContent: UIViewController, sourcePoint: CGPoint, size: CGSize? = nil, delegate: UIPopoverPresentationControllerDelegate? = nil, completion: ( () -> Void)? = nil) {
```
LucianoPAlmeida: I think it is not available on tvOS so will need iOS only compile conditional check :))

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 # Upcoming release
 
 ### Added
+- **UIViewController**
+    - Added `presentPopover(_:,sourcePoint:,size:,delegate:,completion:)` method to quickly present a `UIViewController` as a popover. [#593](https://github.com/SwifterSwift/SwifterSwift/pull/593) by [marcocapano](https://github.com/marcocapano)
```
guykogus: Please remove the commas
```suggestion
    - Added `presentPopover(_:sourcePoint:size:delegate:completion:)` method to quickly present a `UIViewController` as a popover. [#593](https://github.com/SwifterSwift/SwifterSwift/pull/593) by [marcocapano](https://github.com/marcocapano)
```

Sources/Extensions/UIKit/UIViewControllerExtensions.swift
guykogus: Minor white space change
```suggestion
    public func presentPopover(_ popoverContent: UIViewController, sourcePoint: CGPoint, size: CGSize? = nil, delegate: UIPopoverPresentationControllerDelegate? = nil, completion: (() -> Void)? = nil) {
```

Sources/Extensions/UIKit/UIViewControllerExtensions.swift
guykogus: Agreed. If anything, you might want to start the function with
```Swift
guard let popoverPresentationVC = popoverContent.popoverPresentationController else { return }
```
Also, you should almost never use `!`, it's a serious code smell in Swift.

Sources/Extensions/UIKit/UIViewControllerExtensions.swift
```diff
@@ -96,5 +96,29 @@ public extension UIViewController {
...
+        if let size = size {
+            popoverContent.preferredContentSize = size
+        }
+
+        let popoverPresentationVC = popoverContent.popoverPresentationController!
+        popoverPresentationVC.sourceView = self.view
+        popoverPresentationVC.sourceRect = CGRect(origin: sourcePoint, size: .zero)
+        popoverPresentationVC.delegate = delegate
+
+        self.present(popoverContent, animated: true, completion: completion)
```
guykogus: `self` is not needed here, either.
I also recommend adding `animated` as a parameter for the function with a default value of `true`.

Tests/UIKitTests/UIViewControllerExtensionsTests.swift
```diff
@@ -120,5 +120,16 @@ final class UIViewControllerExtensionsTests: XCTestCase {
         XCTAssertNil(childViewController.parent)
     }
 
+    func testPresentPopover() {
+        let popover = UIViewController()
+        let presentingViewController = UIViewController()
+
+        presentingViewController.presentPopover(popover, sourcePoint: presentingViewController.view.center) {
+            XCTAssertEqual(presentingViewController.presentedViewController, popover)
```
guykogus: Are you sure that these even get called? The test may shut down before even getting here. You may need to use an `XCTestExpectation` to ensure that they're called.

Sources/Extensions/UIKit/UIViewControllerExtensions.swift
marcocapano: 👍

Sources/Extensions/UIKit/UIViewControllerExtensions.swift
marcocapano: May I suggest using an `if let popoverPresentationVC = popoverContent.popoverPresentationController` ? 
This way this function will present a popover when on iPad or on iPhone when the delegate specifies so, and will not when the `presentationControllerVC` is `nil` (default situation on iPhone)

Sources/Extensions/UIKit/UIViewControllerExtensions.swift
```diff
@@ -96,5 +96,33 @@ public extension UIViewController {
...
+    ///
+    /// - Parameters:
+    ///   - popoverContent: the view controller to add as a popover.
+    ///   - sourcePoint: the point in which to anchor the popover.
+    ///   - size: the size of the popover. Default uses the popover preferredContentSize.
+    ///   - delegate: the popover's presentationController delegate. Default is nil.
+    ///   - animated: Pass true to animate the presentation; otherwise, pass false.
+    ///   - completion: The block to execute after the presentation finishes. Default is nil.
+    public func presentPopover(_ popoverContent: UIViewController, sourcePoint: CGPoint, size: CGSize? = nil, delegate: UIPopoverPresentationControllerDelegate? = nil, animated: Bool = true, completion: (() -> Void)? = nil) {
+
```
guykogus: Please remove this empty line, it's not necessary.

Tests/UIKitTests/UIViewControllerExtensionsTests.swift
```diff
@@ -120,5 +120,22 @@ final class UIViewControllerExtensionsTests: XCTestCase {
...
+            XCTAssertEqual(popover.presentingViewController, presentingViewController)
+            XCTAssertEqual(popover.modalPresentationStyle, .popover)
+
+            exp.fulfill()
+        }
+
+        waitForExpectations(timeout: 15)
```
guykogus: It's a bit too long. 1 second should be enough.

marcocapano:
I'm not sure why the completion block is not called, even though I pass it into the present function:

```swift
public func presentPopover(_ popoverContent: UIViewController, sourcePoint: CGPoint, size: CGSize? = nil, delegate: UIPopoverPresentationControllerDelegate? = nil, animated: Bool = true, completion: (() -> Void)? = nil) {
        popoverContent.modalPresentationStyle = .popover

        if let size = size {
            popoverContent.preferredContentSize = size
        }

        if let popoverPresentationVC = popoverContent.popoverPresentationController {
            popoverPresentationVC.sourceView = view
            popoverPresentationVC.sourceRect = CGRect(origin: sourcePoint, size: .zero)
            popoverPresentationVC.delegate = delegate
        }

        present(popoverContent, animated: animated, completion: completion)
    }`

**Comment by guykogus** [comment]
My guess is that animations are disabled during unit tests, and therefore it's not calling the completion block. Run your function with `animated: false` and then run the `XCAssert...` calls directly after it. It means that you won't need the `XCExpectation` as well.

Also, for your tests you should use test the different possible parameter values, otherwise the code coverage will warn you that you're not testing all paths.

**Comment by marcocapano** [comment]
Thank you!

> My guess is that animations are disabled during unit tests, and therefore it's not calling the completion block. Run your function with `animated: false` and then run the `XCAssert...` calls directly after it. It means that you won't need the `XCExpectation` as well.
> 
> Also, for your tests you should use test the different possible parameter values, otherwise the code coverage will warn you that you're not testing all paths.

Thank you! Does that mean that I cannot test the case with `animated: true` ?

**Comment by guykogus** [comment]
> Does that mean that I cannot test the case with `animated: true` ?

Probably not

**Comment by guykogus** [comment]
Your tests are failing...

**Comment by marcocapano** [comment]
> Your tests are failing...

Yeah I'm working on it, seems like presentedViewController and presentingViewController are not being set

File: Tests/UIKitTests/UIViewControllerExtensionsTests.swift [hunk 9]
```diff
@@ -120,5 +120,22 @@ final class UIViewControllerExtensionsTests: XCTestCase {
         XCTAssertNil(childViewController.parent)
     }
 
+    func testPresentPopover() {
```
LucianoPAlmeida: Hey @marcocapano :))
Seems like its falling on tvOS, I think you should also constraint the tests to #if os(iOS) 👍

marcocapano: You're right! Did it for the source but not for the test

marcocapano: What do you think of this? I would really appreciate a suggestion 😊🙏🏻
>  If I add another test for example to check that the size of the popover is the one set in the parameter, should I still test other things like presentedViewController, presentingViewController etc.. that I already test in the other test?

guykogus: Yes, it ensures that you have consistent behaviour in your function.

LucianoPAlmeida: For sure 👍

marcocapano: Done, thanks for your help!

Tests/UIKitTests/UIViewControllerExtensionsTests.swift
```diff
@@ -120,5 +120,73 @@ final class UIViewControllerExtensionsTests: XCTestCase {
...
+    #if os(iOS)
+    func testPresentPopover() {
+        let popover = UIViewController()
+        let presentingViewController = UIViewController()
+
+        let window = UIWindow()
+        window.rootViewController = presentingViewController
+        window.rootViewController = presentingViewController
+        window.addSubview(presentingViewController.view)
+        RunLoop.current.run(until: Date())
```
guykogus: Is this your way of waiting for something to happen asynchronously? Generally you would use `XCTestExpectation`, which allows you to wait for callback methods. Maybe this is what it's doing internally, I'm not sure, but it may help to make it clearer what are your intentions.

Tests/UIKitTests/UIViewControllerExtensionsTests.swift
```diff
@@ -120,5 +120,73 @@ final class UIViewControllerExtensionsTests: XCTestCase {
...
     }
 
+    #if os(iOS)
+    func testPresentPopover() {
+        let popover = UIViewController()
+        let presentingViewController = UIViewController()
+
+        let window = UIWindow()
+        window.rootViewController = presentingViewController
+        window.rootViewController = presentingViewController
```
guykogus: You're setting this twice.

Tests/UIKitTests/UIViewControllerExtensionsTests.swift
marcocapano: Unfortunately I cannot use the callback because the completion is not called in unit tests due to disabled animations. Also the window + runloop thing is to make sure the presentation is as similar to a real app use case as it can. I found that this is the only way for the `presentedViewController` and `presentingViewController` properties to be set.
I'll write a comment to make it clear

guykogus: What I mean is calling `DispatchQueue.main.async { ... }`. That's my usual trick for letting the run loop perform whatever it needs to. I didn't know about this trick you're using here. I figured that the main thread would be blocked otherwise and you couldn't get the rest of the UI to run without breaking away from it and coming back later. I really like your way!

Tests/UIKitTests/UIViewControllerExtensionsTests.swift
```diff
@@ -120,5 +120,76 @@ final class UIViewControllerExtensionsTests: XCTestCase {
         XCTAssertNil(childViewController.parent)
     }
 
+    #if os(iOS)
+    func testPresentPopover() {
+        let popover = UIViewController()
+        let presentingViewController = UIViewController()
+
+        //Putting the view controller in a window and running a RunLoop seems to be the only way to make
```
guykogus: A small and finicky note, but please put a space after the `/`s (in all the other comments too).
```suggestion
        // Putting the view controller in a window and running a RunLoop seems to be the only way to make
```

Tests/UIKitTests/UIViewControllerExtensionsTests.swift
marcocapano: Thanks! Also I try to avoid blocks when I can use another solutions, consecutive lines of code without additional indentation looks "cleaner"

marcocapano: Related: when running all the tests some day ago I found that another test gives a warning (not failing though) for a similar reason (UI not properly set up), but I can't remember which one. I'll give it a look

Tests/UIKitTests/UIViewControllerExtensionsTests.swift
guykogus: @marcocapano Please look at all your other lines of comments where you've done this as well.

PR #597, opened 2018-10-30
Added encode(to url:) method for Encodable protocol

AlexeiGitH:
This method allows to save an object that conforms to Encodable protocol to a file at specified URL.
Example:
```Swift
func saveArrayTo(fileURL: URL) {
   let array = [1,2,3,4,5] // array conforms to Encodable
   do {
       try array.encode(to: fileURL)
   }
   catch {
      print(error)
   }
}
```

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 # Upcoming release
 
 ### Added
+- **Encodable**
+- Added  `encode(to url:)` method to save Encodable object to file at provided URL. [#597](https://github.com/SwifterSwift/SwifterSwift/pull/597) by [AlexeiGitH](https://github.com/AlexeiGitH)
 - **Date**
 - Added `tomorrow` computed property to get tomorrow's date avoiding calling `adding(_:value:)` function. (Completes PR #578) [#587](https://github.com/SwifterSwift/SwifterSwift/pull/587) by [AlexeiGitH](https://github.com/AlexeiGitH)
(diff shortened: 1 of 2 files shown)


AlexeiGitH:
Don't know why my previous commits appeared in this pr. 
I am pretty sure my last one was merged and I pulled all the changes from the upstream repo.

Sources/Extensions/Foundation/EncodableExtensions.swift
```diff
@@ -0,0 +1,27 @@
...
+//  EncodableExtensions.swift
+//  SwifterSwift
+//
+//  Created by Alexei Kozachenko on 10/29/2018.
+//  Copyright © 2018 SwifterSwift
+//
+
+import Foundation
+
+public extension Encodable {
```
LucianoPAlmeida: Hey @AlexeiGitH :))
This to me feels more like a FileManager extension then a Codable. I think Codable abstracts a representation that can be encoded/decoded and we should not add file saving(or FileManager responsibility) logic or even encoding logic as an extension to this type. To me an API that maybe will make more sense is something like:
```swift
extension FileManager {
  func createFile(at path: URL, contentsEncodable: Codable, encoder: JSONEncoder = JSONEncoder()){ ...
}
```
Let me know your thoughts about that :))

AlexeiGitH: Hi @LucianoPAlmeida,

I actually have this method as a part of FileManager extensions in my repo (I think I updated the file :D) , though when I was thinking about adding it here I decided to make it as an extension of Encodable. 

It is implemented as 
```Swift
public static func encode<T: Encodable>(_ object: T, to url: URL) throws {
       ...
}
```
I think it would be convenient having it as a method of encodable class, but I totally agree that saving to a file is something that FileManager should handle. 

I think I should return to my original idea and implement it in a FileManager.

PR #598, opened 2018-10-30
Added encode(_:to:) method to FileManager 

AlexeiGitH:
This method allows to save an object that conforms to Encodable protocol to a file at specified URL.
Example:
```Swift
func saveArrayTo(fileURL: URL) {
   let array = [1,2,3,4,5] // array conforms to Encodable
   do {
       try try FileManager.encodeJSON(array, to: fileURL)
   }
   catch {
      print(error)
   }
}
```
## Checklist


Sources/Extensions/Foundation/FileManagerExtensions.swift
```diff
@@ -56,5 +56,19 @@ public extension FileManager {
         return nil
     }
 
+    /// SwifterSwift: Save Encodable object to a given URL.
+    ///
+    /// - Parameters:
```
LucianoPAlmeida: Missing 'T' parameter on docs

Sources/Extensions/Foundation/FileManagerExtensions.swift
```diff
@@ -56,5 +56,19 @@ public extension FileManager {
...
+    ///   - url: URL to save object to.
+    /// - Throws: Throws any errors thrown by JSONEncoder or FileManager.
+    public static func encode<T: Encodable>(_ object: T, to url: URL) throws {
```
LucianoPAlmeida: @AlexeiGitH What do you think about adding a JSONEncoder parameter with default value JSONEncoder() so people can use custom encoders here?

Sources/Extensions/Foundation/FileManagerExtensions.swift
AlexeiGitH: added

Tests/FoundationTests/FileManagerExtensionsTests.swift
```diff
@@ -80,6 +80,19 @@ final class FileManagerExtensionsTests: XCTestCase {
         } catch {}
     }
 
+    func testEncode() {
+        let point = CGPoint(x: 1, y: 2) //CGPoint conforms to Encodable
+        let tmpDirPath = NSTemporaryDirectory()
```
guykogus: See [here](https://nshipster.com/temporary-files/). Interesting update regarding how to correctly use temporary directories.

Sources/Extensions/Foundation/FileManagerExtensions.swift
```diff
@@ -56,5 +56,21 @@ public extension FileManager {
...
     }
 
+    /// SwifterSwift: Save Encodable object to a given URL.
+    ///
+    /// - Parameters:
+    ///   - object: Object of type T that conforms to Encodable protocol.
+    ///   - url: URL to save object to.
+    ///   - encoder: JSONEncoder.
+    /// - Throws: Throws any errors thrown by JSONEncoder or FileManager.
+    public static func encode<T: Encodable>(_ object: T, to url: URL, using encoder: JSONEncoder=JSONEncoder()) throws {
```
guykogus: This should not be a static function. Add spacing for the encoder param.
```suggestion
    public func encode<T: Encodable>(_ object: T, to url: URL, using encoder: JSONEncoder = JSONEncoder()) throws {
```

Sources/Extensions/Foundation/FileManagerExtensions.swift
```diff
@@ -56,5 +56,21 @@ public extension FileManager {
...
+        do {
```
guykogus: There's no need to use `do-catch` because this function `throws`.

Sources/Extensions/Foundation/FileManagerExtensions.swift
```diff
@@ -56,5 +56,21 @@ public extension FileManager {
...
+            let encoder = encoder
+            let data = try encoder.encode(object)
+            FileManager.default.createFile(atPath: url.path, contents: data, attributes: nil)
```
guykogus: When this function is not `static` you can call the local `createFile` function.
```suggestion
            createFile(atPath: url.path, contents: data, attributes: nil)
```

Sources/Extensions/Foundation/FileManagerExtensions.swift
guykogus: You should add the parameter
```Swift
attributes: [FileAttributeKey: Any]? = nil
```
and pass it to `createFile`

Sources/Extensions/Foundation/FileManagerExtensions.swift
guykogus: `createFile` returns a boolean value to indicate if it succeeded or failed. I think it would make sense to return it as the return value of this function so that users will be informed of success/failure.
In my project I have a custom `Error` that I `throw` if the call to `createFile` fails, however adding this isn't possible given the structure of SwifterSwift.

Tests/FoundationTests/FileManagerExtensionsTests.swift
```diff
@@ -80,6 +80,19 @@ final class FileManagerExtensionsTests: XCTestCase {
...
+        let fileURL = URL(fileURLWithPath: tmpDirPath.appending("testEncodeFile"))
+
+        do {
```
guykogus: Instead of using `do-catch` here, you should use `XCTAssertNoThrow`s.

Tests/FoundationTests/FileManagerExtensionsTests.swift
```diff
@@ -80,6 +80,19 @@ final class FileManagerExtensionsTests: XCTestCase {
...
+            try FileManager.encode(point, to: fileURL)
+            try FileManager.default.removeItem(at: fileURL) //delete created file
```
guykogus: No need for the comment at the end.

SwifterSwift.xcodeproj/project.pbxproj
```diff
@@ -333,13 +333,6 @@
 		278CA0911F9A9679004918BD /* NSImageExtensionsTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 278CA0901F9A9679004918BD /* NSImageExtensionsTests.swift */; };
 		42F53FEC2039C5AC0070DC11 /* UIStackViewExtensions.swift in Sources */ = {isa = PBXBuildFile; fileRef = 42F53FEB2039C5AC0070DC11 /* UIStackViewExtensions.swift */; };
 		42F53FF02039C7140070DC11 /* UIStackViewExtensionsTest.swift in Sources */ = {isa = PBXBuildFile; fileRef = 42F53FEF2039C7140070DC11 /* UIStackViewExtensionsTest.swift */; };
-		664CB983217243D200FC87B4 /* DispatchQueueExtensions.swift in Sources */ = {isa = PBXBuildFile; fileRef = 664CB982217243D200FC87B4 /* DispatchQueueExtensions.swift */; };
```
guykogus: Please revert all these changes.

Tests/FoundationTests/FileManagerExtensionsTests.swift
AlexeiGitH: I have 3 calls that have try in them (make temporaryDirectoryURL, encode, removeItem).
Should I only use `XCTAssertNoThrow` for `encode` method but inside do{} block?

guykogus: Or actually, if you want you can keep it like this, just get rid of the `XCTAssert(true)` and instead create other assertions to check:
1. The new file exists
2. It contains the expected JSON value
3. It no longer exists after calling `removeItem`

Sources/Extensions/Foundation/FileManagerExtensions.swift
```diff
@@ -56,5 +56,19 @@ public extension FileManager {
...
+    /// SwifterSwift: Save Encodable object to a given URL.
+    ///
+    /// - Parameters:
+    ///   - object: Object of type T that conforms to Encodable protocol.
+    ///   - url: URL to save object to.
+    ///   - encoder: JSONEncoder.
+    /// - Throws: Throws any errors thrown by JSONEncoder or FileManager.
+    public func encode<T: Encodable>(_ object: T, to url: URL, using encoder: JSONEncoder = JSONEncoder(), attributes: [FileAttributeKey: Any]? = nil) throws -> Bool {
+        let data = try encoder.encode(object)
+        if createFile(atPath: url.path, contents: data, attributes: attributes) {
```
guykogus: Why not just call:
```suggestion
        return createFile(atPath: url.path, contents: data, attributes: attributes)
```
?

AlexeiGitH: @guykogus, To perform a second check (Check that file contains the expected JSON value), I am assuming I need to decode the file? (I will add a decode() extension method in another PR so I don't add more than 1 extension in this one)

AlexeiGitH:
@guykogus, I have reverted change for getting tempDir url. 
`let temporaryDirectoryURL = FileManager.default.temporaryDirectory` works for all devices, but the previous implementation with try block fails on iOS.

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 # Upcoming release
 
 ### Added
+- **Encodable**
+- Added  `encode(_:to:)` method to save Encodable object to file at provided URL. [#598](https://github.com/SwifterSwift/SwifterSwift/pull/598) by [AlexeiGitH](https://github.com/AlexeiGitH)
```
guykogus: I know I'm being pedantic but... 😛 
```suggestion
- Added `encode(_:to:)` method to save Encodable object to file at provided URL. [#598](https://github.com/SwifterSwift/SwifterSwift/pull/598) by [AlexeiGitH](https://github.com/AlexeiGitH)
```

AlexeiGitH: Fixed. It's so easy to miss because of that `

Sources/Extensions/Foundation/FileManagerExtensions.swift
```diff
@@ -56,5 +56,16 @@ public extension FileManager {
         return nil
     }
 
+    /// SwifterSwift: Save Encodable object to a given URL.
+    ///
+    /// - Parameters:
+    ///   - object: Object of type T that conforms to Encodable protocol.
+    ///   - url: URL to save object to.
+    ///   - encoder: JSONEncoder.
```
LucianoPAlmeida: Missing attributes parameter on docs :)

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 # Upcoming release
 
 ### Added
+- **Encodable**
```
LucianoPAlmeida: Isn't a `FileManager` extension?

Sources/Extensions/Foundation/FileManagerExtensions.swift
```diff
@@ -56,5 +56,16 @@ public extension FileManager {
...
+    /// - Throws: Throws any errors thrown by JSONEncoder or FileManager.
+    public func encode<T: Encodable>(_ object: T, to url: URL, using encoder: JSONEncoder = JSONEncoder(), attributes: [FileAttributeKey: Any]? = nil) throws -> Bool {
```
guykogus: I think you should switch `encoder` and `url` parameters. It reads more naturally because of the `attributes` parameter that it would paired with `url`.
Also, you need to add `attributes` documentation.

CHANGELOG.md
AlexeiGitH: @LucianoPAlmeida Fixed

Tests/FoundationTests/FileManagerExtensionsTests.swift
```diff
@@ -80,6 +80,28 @@ final class FileManagerExtensionsTests: XCTestCase {
...
     }
 
+    func testEncode() {
+        let point = CGPoint(x: 1, y: 2) //CGPoint conforms to Encodable
+        do {
+            let temporaryDirectoryURL = FileManager.default.temporaryDirectory
+            let temporaryFilename = ProcessInfo().globallyUniqueString
+            let temporaryFileURL = temporaryDirectoryURL.appendingPathComponent(temporaryFilename)
+
+            XCTAssertNoThrow(try FileManager.default.encode(point, to: temporaryFileURL))
```
guykogus: No need for the assertion here since you're `catch`-ing. It won't make sense to run the rest of the test anyway if this fails, so better that it be caught.

Tests/FoundationTests/FileManagerExtensionsTests.swift
```diff
@@ -80,6 +80,28 @@ final class FileManagerExtensionsTests: XCTestCase {
...
+            XCTAssert(FileManager.default.fileExists(atPath: temporaryFileURL.path))
+
+            let data = FileManager.default.contents(atPath: temporaryFileURL.path)
+            XCTAssertNotNil(data)
+
+            let decoder = JSONDecoder()
+            XCTAssertNoThrow(try decoder.decode(CGPoint.self, from: data!))
```
guykogus: Again, no assertion. Get the value and check that it matches the original.

Sources/Extensions/Foundation/DateExtensions.swift
```diff
@@ -455,7 +455,9 @@ public extension Date {
     ///     let yesterday = date.yesterday // "Oct 2, 2018, 10:57:11"
     ///
     public var yesterday: Date {
-        return addingTimeInterval(-86400.0)
+        let cal = Calendar.current
```
guykogus: Please do this in a different PR

guykogus: @AlexeiGitH please remove these changes and fix the conflicts so that we can finally merge this in 😅

Tests/FoundationTests/FileManagerExtensionsTests.swift
```diff
@@ -80,6 +80,30 @@ final class FileManagerExtensionsTests: XCTestCase {
...
+
+            let data = FileManager.default.contents(atPath: temporaryFileURL.path)
+            XCTAssertNotNil(data)
+
+            let decoder = JSONDecoder()
+            let decodedPoint = try decoder.decode(CGPoint.self, from: data!)
+            XCTAssertTrue(point == decodedPoint)
+
+            try FileManager.default.removeItem(at: temporaryFileURL)
+            XCTAssertFalse(FileManager.default.fileExists(atPath: temporaryFileURL.path))
```
guykogus: You don't really need this check, you can actually change the line above to:
```Swift
XCTAssertTrue(try FileManager.default.removeItem(at: temporaryFileURL))
```
Even that you don't really need, because if the removal fails it will throw an exception, which you will catch.
Sorry that I didn't catch this in the last review... This one seems to be going on forever 😅

guykogus:
@omaralbeik @LucianoPAlmeida can you please check?

LucianoPAlmeida:
Sorry, I just notice that we missed that one 😅, just cleaning up those old PR's. 
@AlexeiGitH Closing this one due inactivity, feel free to reopen :))

PR #599, opened 2018-11-01
Added FileManager extension method "decode"

AlexeiGitH:
This method allows to decode an object that was previously save to a file.

## Checklist

- [+] I checked the [**Contributing Guidelines**](https://github.com/SwifterSwift/SwifterSwift/blob/master/CONTRIBUTING.md) before creating this request.
- [+] New extensions are written in Swift 4.
- [+] New extensions support iOS 8.0+ / tvOS 9.0+ / macOS 10.10+ / watchOS 2.0+.
- [+] I have added tests for new extensions, and they passed.
- [+] All extensions have a **clear** comments explaining their functionality, all parameters and return type in English.
- [+] All extensions are declared as **public**.
- [+] I have added a [changelog](https://github.com/SwifterSwift/SwifterSwift/blob/master/CHANGELOG_GUIDELINES.md) entry describing my changes.


AlexeiGitH:
Not sure what exactly failed here.

Sources/Extensions/Foundation/FileManagerExtensions.swift
```diff
@@ -56,5 +56,20 @@ public extension FileManager {
...
 
+    /// SwifterSwift: Decode Encodable object from file.
+    ///
+    /// - Parameters:
+    ///   - url: URL a file with saved object data.
+    ///   - type: type of the object that is being decoded.
+    ///   - decoder: JSONDecoder.
+    /// - Throws: Throws any errors thrown by JSONDecoder.
+    /// - Returns: An object of specified type. Returns nil if failed to read from URL.
+    public func decode<T: Decodable>(from url: URL, as type: T.Type, using decoder: JSONDecoder = JSONDecoder()) throws -> T? {
```
guykogus: I think that since your `encode` function's parameters are ordered as
`encode(_ object: T, to url: URL, using encoder: JSONEncoder...)`
then it would make sense for this to be
`decode(_ type: T.Type, from url: URL, using decoder: JSONDecoder)`

Tests/FoundationTests/FileManagerExtensionsTests.swift
```diff
@@ -80,6 +80,31 @@ final class FileManagerExtensionsTests: XCTestCase {
...
 
+    func testDecode() {
+        let point = CGPoint(x: 1, y: 2) //CGPoint conforms to Encodable
+        do {
+            let temporaryDirectoryURL = FileManager.default.temporaryDirectory
+            let temporaryFilename = ProcessInfo().globallyUniqueString
+            let temporaryFileURL = temporaryDirectoryURL.appendingPathComponent(temporaryFilename)
+
+            XCTAssertNoThrow(try JSONEncoder().encode(point))
+            let data1 = try JSONEncoder().encode(point)
```
guykogus: So I was wondering about all these, and I think that if you're doing `try-catch` then you don't need the `XCTAssertNoThrow`s. It seems wasteful having to call all these functions multiple times.
Alternatively I was thinking about having a helper function like this:
```Swift
func XCTAssertNoThrowAndReturn<T>(_ expression: @autoclosure () throws -> T,
                                  _ message: @autoclosure () -> String = "",
                                  file: StaticString = #file,
                                  line: UInt = #line) -> T? {
    do {
        return try expression()
    } catch {
        XCTFail(message())
        return nil
    }
}
```
Then you could just call
```Swift
let data1 = XCTAssertNoThrowAndReturn(try JSONEncoder().encode(point))
```
What do you think?

AlexeiGitH: Helper `XCTAssertNoThrowAndReturn` looks a lot cleaner than doing multiple `XCTAssertNoThrow`.

Will this method be useful in future? If you think it's the only case where it is going to be used, then I probably should just get rid of `XCTAssertNoThrow`.

guykogus: As much as I like it, I think it's better the way that you initially did it using do-catch. If you're trying to call and get a value, the next steps probably don't make sense without it if the method you're calling `throw`s. In that case it's fine to let the `do-catch` fail and force the rest of the tests to be skipped. I think `XCTAssertNoThrowAndReturn` may be useful in some scenarios, but perhaps here you don't need it.

AlexeiGitH: Sounds good to me. 
Any other changes I should make?

guykogus: Yes, please look at the other comments.

Sources/Extensions/Foundation/FileManagerExtensions.swift
```diff
@@ -56,5 +56,20 @@ public extension FileManager {
...
+        if let data = FileManager.default.contents(atPath: url.path) {
```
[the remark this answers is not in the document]
AlexeiGitH: Changed

Sources/Extensions/Foundation/FileManagerExtensions.swift
AlexeiGitH: The order of arguments now matched to `encode` method

Tests/FoundationTests/FileManagerExtensionsTests.swift
```diff
@@ -80,6 +80,31 @@ final class FileManagerExtensionsTests: XCTestCase {
...
     }
 
+    func testDecode() {
+        let point = CGPoint(x: 1, y: 2) //CGPoint conforms to Encodable
+        do {
+            let temporaryDirectoryURL = FileManager.default.temporaryDirectory
+            let temporaryFilename = ProcessInfo().globallyUniqueString
+            let temporaryFileURL = temporaryDirectoryURL.appendingPathComponent(temporaryFilename)
+
+            XCTAssertNoThrow(try JSONEncoder().encode(point))
```
guykogus: If you're catching the `try` then there's no need to do an assert and _then_ get the value.

Tests/FoundationTests/FileManagerExtensionsTests.swift
```diff
@@ -80,6 +80,31 @@ final class FileManagerExtensionsTests: XCTestCase {
...
+            let data1 = try JSONEncoder().encode(point)
+            XCTAssertTrue(FileManager.default.createFile(atPath: temporaryFileURL.path, contents: data1, attributes: nil))
+
+            XCTAssert(FileManager.default.fileExists(atPath: temporaryFileURL.path))
+
+            let data2 = FileManager.default.contents(atPath: temporaryFileURL.path)
+            XCTAssertNotNil(data2)
+
+            XCTAssertNoThrow(try FileManager.default.decode(as: CGPoint.self, from: temporaryFileURL))
```
guykogus: Again, since you're `catch`-ing you don't want to make this assertion. Instead you actually want to get the result back and test that it matches the original point that you saved.

Sources/Extensions/Foundation/FileManagerExtensions.swift
```diff
@@ -56,5 +56,20 @@ public extension FileManager {
...
+    ///   - decoder: JSONDecoder.
+    ///   - url: URL a file with saved object data.
+    /// - Throws: Throws any errors thrown by JSONDecoder.
+    /// - Returns: An object of specified type. Returns nil if failed to read from URL.
+    public func decode<T: Decodable>(as type: T.Type, using decoder: JSONDecoder = JSONDecoder(), from url: URL) throws -> T? {
+        guard let data = FileManager.default.contents(atPath: url.path) else {
+            return nil
+        }
+        let model = try decoder.decode(type, from: data)
+        return model
```
LucianoPAlmeida: Here you can just return the result `return try decoder.decode(type, from: data)` :))

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 # Upcoming release
 
 ### Added
+- **FileManager**
+- Added `decode(from:as:using:)'` method to decode Decodable object previously saved to a file. [#599](https://github.com/SwifterSwift/SwifterSwift/pull/599) by [AlexeiGitH](https://github.com/AlexeiGitH)
```
LucianoPAlmeida: Here `to decode Decodable object` could be `to decode a Decodable object`? English is not my native language so this is more a question 😆

Sources/Extensions/Foundation/FileManagerExtensions.swift
```diff
@@ -56,5 +56,20 @@ public extension FileManager {
...
+    /// SwifterSwift: Decode Encodable object from file.
+    ///
+    /// - Parameters:
+    ///   - type: type of the object that is being decoded.
+    ///   - decoder: JSONDecoder.
+    ///   - url: URL a file with saved object data.
+    /// - Throws: Throws any errors thrown by JSONDecoder.
+    /// - Returns: An object of specified type. Returns nil if failed to read from URL.
+    public func decodeJSON<T: Decodable>(as type: T.Type, using decoder: JSONDecoder = JSONDecoder(), from url: URL) throws -> T? {
+        guard let data = FileManager.default.contents(atPath: url.path) else {
```
guykogus: You should be using the receiving `FileManager` object.
```suggestion
        guard let data = contents(atPath: url.path) else {
```

Tests/FoundationTests/FileManagerExtensionsTests.swift
```diff
@@ -80,6 +80,31 @@ final class FileManagerExtensionsTests: XCTestCase {
...
+            let encodedPoint = try JSONEncoder().encode(point)
+            XCTAssertTrue(FileManager.default.createFile(atPath: temporaryFileURL.path, contents: encodedPoint, attributes: nil))
+
+            XCTAssert(FileManager.default.fileExists(atPath: temporaryFileURL.path))
+
+            let data2 = FileManager.default.contents(atPath: temporaryFileURL.path)
+            XCTAssertNotNil(data2)
+
+            let decodedPoint = try FileManager.default.decodeJSON(as: CGPoint.self, from: temporaryFileURL)
+            XCTAssertTrue(point == decodedPoint)
```
guykogus: Use `XCTAssertEqual`
```suggestion
            XCTAssertEqual(point, decodedPoint)
```

Tests/FoundationTests/FileManagerExtensionsTests.swift
```diff
@@ -80,6 +80,31 @@ final class FileManagerExtensionsTests: XCTestCase {
...
 
+    func testDecode() {
+        let point = CGPoint(x: 1, y: 2) //CGPoint conforms to Encodable
+        do {
+            let temporaryDirectoryURL = FileManager.default.temporaryDirectory
+            let temporaryFilename = ProcessInfo().globallyUniqueString
+            let temporaryFileURL = temporaryDirectoryURL.appendingPathComponent(temporaryFilename)
+
+            let encodedPoint = try JSONEncoder().encode(point)
+            XCTAssertTrue(FileManager.default.createFile(atPath: temporaryFileURL.path, contents: encodedPoint, attributes: nil))
```
guykogus: No need fro `True`
```suggestion
            XCTAssert(FileManager.default.createFile(atPath: temporaryFileURL.path, contents: encodedPoint, attributes: nil))
```

Sources/Extensions/Foundation/FileManagerExtensions.swift
AlexeiGitH: fixed

Tests/FoundationTests/FileManagerExtensionsTests.swift
AlexeiGitH: fixed

Tests/FoundationTests/FileManagerExtensionsTests.swift
AlexeiGitH: @guykogus applied changes to all comments

Sources/Extensions/Foundation/FileManagerExtensions.swift
```diff
@@ -56,6 +56,22 @@ public extension FileManager {
...
+    /// SwifterSwift: Decode Encodable object from file.
+    ///
+    /// - Parameters:
+    ///   - type: type of the object that is being decoded.
+    ///   - decoder: JSONDecoder.
+    ///   - url: URL a file with saved object data.
+    /// - Throws: Throws any errors thrown by JSONDecoder.
+    /// - Returns: An object of specified type. Returns nil if failed to read from URL.
+    public func decodeJSON<T: Decodable>(as type: T.Type, using decoder: JSONDecoder = JSONDecoder(), from url: URL) throws -> T? {
+        guard let data = contents(atPath: url.path) else {
```
guykogus: Put this all in 1 line
```suggestion
        guard let data = contents(atPath: url.path) else { return nil }
```

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 
 ### Added
 - **FileManager**
+- Added `decodeJSON(from:as:using:)'` method to decode a Decodable object previously saved to a file. [#599](https://github.com/SwifterSwift/SwifterSwift/pull/599) by [AlexeiGitH](https://github.com/AlexeiGitH)
```
guykogus: The signature was changed
```suggestion
- Added `decodeJSON(as:using:from:)'` method to decode a `Decodable` object previously saved to a file. [#599](https://github.com/SwifterSwift/SwifterSwift/pull/599) by [AlexeiGitH](https://github.com/AlexeiGitH)
```

guykogus:
@omaralbeik @LucianoPAlmeida can you please check?

LucianoPAlmeida:
Sorry, I just notice that we missed that one 😅, not remember why we didn't merge it 
@AlexeiGitH Closing this one due to inactivity, feel free to reopen :))

PR #606, opened 2018-11-17
Improvement common suffix extension

LucianoPAlmeida:
Notice that the old implementation was not so performant because it was eagerly reversing the string and creating new strings copies. Just improving implementation and tests based on [Core libs Foundation](https://github.com/apple/swift-corelibs-foundation/blob/c82398aa779dd820c0f664b17197d9ed78e7052d/TestFoundation/TestNSString.swift#L1311) 

## Checklist


.travis.yml
```diff
@@ -31,4 +31,4 @@ script:
   - xcodebuild clean build -project "$PROJECT" -scheme "$WATCHOS_SCHEME" -destination "$WATCHOS_DESTINATION" | XCPRETTY_JSON_FILE_OUTPUT="xcodebuild-watchos.json" xcpretty -f `xcpretty-json-formatter`
   - swiftlint lint
   - bundle exec danger --verbose
-  - pod lib lint
+  - pod lib lint --skip-tests
```
LucianoPAlmeida: Should we do this? The podlib lint was often failing the build with timeout because it was taking too long to run. Since we already build and tests all targets before run this command, I believe we can skip build and testing for the lint and in fact only lint the podspec and subspecs. What do you guys think?
cc @omaralbeik @SD10 @guykogus

Sources/Extensions/SwiftStdlib/StringProtocolExtensions.swift
```diff
@@ -19,8 +19,33 @@ public extension StringProtocol where Index == String.Index {
     ///     - Parameter options: Options for the comparison.
     /// - Returns: The longest common suffix of the receiver and the given String
     public func commonSuffix<T: StringProtocol>(with aString: T, options: String.CompareOptions = []) -> String {
-        let reversedSuffix = String(reversed()).commonPrefix(with: String(aString.reversed()), options: options)
```
guykogus: You don't need to do this all by hand.
```Swift
public func commonSuffix<T: StringProtocol>(with aString: T, options: String.CompareOptions = []) -> String {
    var suffix = ""
    suffix.reserveCapacity(count)
    for (lhs, rhs) in zip(reversed(), aString.reversed()) {
        guard String(lhs).compare(String(rhs), options: options) == .orderedSame else { break }
        suffix.append(lhs)
    }
    return String(suffix.reversed())
}
```

.travis.yml
guykogus: It makes sense to me if we're performing the tests already.

Sources/Extensions/SwiftStdlib/StringProtocolExtensions.swift
LucianoPAlmeida: That makes sense to use zip since `Zip2Sequence` and `ReversedCollection` are "lazy".
But I think is not an issue handle the compare by hand since we avoid allocate new strings and you only handle options that make sense in this context of single character comparation (case Insentive and literal).  And maybe avoid reverse it at the end by insert(at: 0 ...) instead of the append on the suffix. 
Also I'm not sure about reserve capacity for count, maybe it make more sense to use `min(count, aString.count)`, but since String and StringProtocol don't conform to `RandomAccessCollection` the counts probably are not O(1), and also maybe we reserve space that we don't use e. g. Two large strings that don't have any common suffix. 
I'll try take a look at this later to check the options, thank's for the review  @guykogus  :))

LucianoPAlmeida: In the end, I think both implementations seem to have the same performance gain (avoid eager reverse). And using for with lazy reversed and Zip2Sequence to walk the strings by the index are just a simple implementation detail and more personal preference. So I think there's not a big problem to let this implementation here. :)) Let me know what you think.
cc @omaralbeik @SD10 Thoughts?

guykogus: I don't think the current implementation is correct. What if the caller of this function wanted to call with `.diacriticInsensitive`, for example? There's no need to check the options manually. I think that's the main benefit to my suggestion.
Also, I wouldn't suggest inserting at `0`/`startIndex` since appending to the beginning (assuming an array is used internally, which is pretty common practice) is O(n) while appending to the end (if you've reserved enough memory from the beginning) is O(1).

LucianoPAlmeida: I see the point, is just that it bothers me a bit have to convert the character to string to compare 🙃 and since I thought the only options that make sense character wise comparation were those two... but right I'll change :))

.travis.yml
omaralbeik: Makes sense to me

Sources/Extensions/SwiftStdlib/StringProtocolExtensions.swift
```diff
@@ -19,8 +19,28 @@ public extension StringProtocol where Index == String.Index {
...
+
+            if String(self[idx]).compare(String(aString[strIdx]), options: options) != .orderedSame { break }
+
+        } while idx > startIndex && strIdx > aString.startIndex
+
+        if idx == startIndex {
+            return String(self)
+        }
+
+        if strIdx == aString.startIndex {
```
guykogus: All these index edge cases and whatnot is why I think it would've been easier to just use my implementation. I guess your imp potentially saves some memory usage and whatnot, but there are still obviously some potential pitfalls.

LucianoPAlmeida:
@guykogus I think is simple and nice now 😆
That was fun to think about and good improvement
Thank you for all reviews and discussions :))

PR #607, opened 2018-11-18
Adding convenience initializer and test for UILabel

marcocapano:
This PR adds a convenience initializer for UILabel that makes a label decide the font based on the type of content.

``` swift
//This
let label = UILabel(text: text, style: .body)

//is the same as this:
let label = UILabel()
label.font = UIFont.preferredFont(forTextStyle: .body)
label.text = text
```

## Checklist

- [x] New extensions support iOS 8.0+


marcocapano:
I've added a test but I have a doubt: is testing with one style enough?
UIFont.TextStyle has about 11 values, it would be a little

marcocapano:
Should I just use another style?

marcocapano:
> I think just using a common option is fine :))

👍🏻

LucianoPAlmeida:
That's weird, let's try to rebuild and see what happens :)

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 # Upcoming release
 
 ### Added
+- **UILabel**
+    - Added `convenience init(text:style)` to create a `UILabel` with a text and an appropriate font. [#607](https://github.com/SwifterSwift/SwifterSwift/pull/607) by [marcocapano](https://github.com/marcocapano)
```
guykogus: No need to include `convenience` for the signature. Also, the argument isn't a font, that would have been of type `UIFont`, it's more a font _style_. I actually think it would make more sense to pass a `font` instead than a `style`, but if this is useful for you then it's probably good for us to have as well :)
```suggestion
    - Added `init(text:style)` to create a `UILabel` with a text and font style. [#607](https://github.com/SwifterSwift/SwifterSwift/pull/607) by [marcocapano](https://github.com/marcocapano)
```

Sources/Extensions/UIKit/UILabelExtensions.swift
```diff
@@ -18,6 +18,17 @@ public extension UILabel {
         self.text = text
     }
 
+    /// SwifterSwift: Initialize a UILabel with a text and font style.
+    ///
+    /// - Parameters:
+    ///   - text: the label's text.
+    ///   - style: the text style of the label, used to determine which font should be used.
+    public convenience init(text: String, style: UIFont.TextStyle) {
+        self.init()
```
guykogus: What would happen if you didn't call `self.init` and removed `convenience`? Would this still work?

marcocapano: You cannot (or at least from what I know) create a designated initializer of UILabel in an extension.

marcocapano: Ok checked now, the error is "Designated initializer cannot be declared in an extension of 'UILabel'; did you mean this to be a convenience initializer?"

So maybe a convenience initializer is better. Not sure is this is true with all types or just for types we don't "own" because declared in UIKit, because I don't recall having problems in creating initializers in extensions🤔

marcocapano:
I'm not sure why the build keeps "erroring" instead of "failing", seems like something not related to tests(?). Also I ran the test on all platforms and everything is ok. Any idea?

guykogus:
@marcocapano yeah for some reason a tvOS test is failing: https://travis-ci.org/SwifterSwift/SwifterSwift/builds/457830891#L2872
Can't see any reason why... Restarting the build

marcocapano:
Whats odd is that it did the same 3 days ago then everything went ok after @LucianoPAlmeida restarted it. 

> That's weird, let's try to rebuild and see what happens :)

PR #608, opened 2018-11-20
Added extension 'isLeapYear' to Date

dylancfe15:
## Checklist

### DIFF
--- a/Sources/Extensions/Foundation/DateExtensions.swift
+++ b/Sources/Extensions/Foundation/DateExtensions.swift
@@ -457,6 +457,15 @@ public extension Date {
         return Calendar.current.timeZone
     }
 
+    /// SwifterSwift: Yesterday date.
+    ///
+    ///     let date = Date() // "Oct 3, 2018, 10:57:11"
+    ///     let yesterday = date.yesterday // "Oct 2, 2018, 10:57:11"
+    ///
+    public var yesterday: Date {
+        return addingTimeInterval(-86400.0)
+    }
+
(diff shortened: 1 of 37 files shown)


Sources/Extensions/Foundation/DateExtensions.swift
```diff
@@ -344,6 +344,12 @@ public extension Date {
         return Calendar.current.isDate(self, equalTo: Date(), toGranularity: .year)
     }
 
+    /// SwifterSwift: Check if the current year is a leap year.
+    public var isLeapYear: Bool {
```
guykogus: Oddly this is trickier than it seems, because what you're checking here isn't whether or not a date is a leap year, but rather whether or not it's a leap year **in the Gregorian calendar**. The only thing a `Date` object does is represent a specific point in time. It's the `Calendar` class that knows more about days/months/years/etc. It would really make more sense, therefore, to have this as an extension of `Calendar` like this:
```Swift
extension Calendar {
    public func isLeapYear(date: Date) -> Bool {
        switch identifier {
        case .gregorian:
            return range(of: .day, in: .year, for: date)?.upperBound == 367
        default:
            // Other calendars not yet implemented
            return false
        }
    }
}
```
Other calendars handle leap years in different ways, e.g. the Hebrew calendar periodically [adds an extra month](https://en.wikipedia.org/wiki/Hebrew_calendar#Leap_years) to synchronise with the solar cycle. You don't have to try and implement those now, obviously 😬

PR #612, opened 2018-11-26
Added extension 'setDistanceToRightEdge' to UIView

dylancfe15:
## Checklist

### DIFF
--- a/Sources/Extensions/UIKit/UIViewExtensions.swift
+++ b/Sources/Extensions/UIKit/UIViewExtensions.swift
@@ -572,5 +572,21 @@ public extension UIView {
         return ancestorView(where: { $0 is T }) as? T
     }
 
+    /// SwifterSwift: Adjust frame to meet the distance from a UIView to the right edge.
+    ///
+    /// - Parameter keepWith: to keep the original width or not.
+    /// - Parameter distance: distance between the UIView and the right edge.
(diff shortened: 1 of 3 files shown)


Sources/Extensions/UIKit/UIViewExtensions.swift
```diff
@@ -572,5 +572,21 @@ public extension UIView {
         return ancestorView(where: { $0 is T }) as? T
     }
 
+    /// SwifterSwift: Adjust frame to meet the distance from a UIView to the right edge.
+    ///
+    /// - Parameter keepWith: to keep the original width or not.
+    /// - Parameter distance: distance between the UIView and the right edge.
+    public func setDistanceToRightEdge(keepWidth: Bool, distance: CGFloat) {
```
LucianoPAlmeida: Hey @dylancfe15 :))
Few points here:
* This extensions is specific to right edge. Maybe we can do something more generic that work right, bottom, top, left using UIEdge insets. 
* But I don't know if it's worth it 
Because it will be an extension that we can do very easily like:
```swift
view.frame = view.frame.inset(by: UIEdgeInsets(top: 0, left: 0, bottom: 0, right: 5.0))
```
Let me know your thoughts on that :))

dylancfe15: You are right!!! Thanks for your comment.

PR #615, opened 2018-12-07
Added FileManager.temporaryFileDirectory() to create a directory for saving temporary files

guykogus:
`FileManager` provides a way to get the user's temporary files directory, but for saving temporary files it's necessary to actually create a new temp folder within it. This helper method allows that to be done correctly.

```Swift
let tempDirectory = try fileManager.temporaryFileDirectory()
let tempFile1URL = tempDirectory.appendingPathComponent(ProcessInfo().globallyUniqueString)
let tempFile2URL = tempDirectory.appendingPathComponent(ProcessInfo().globallyUniqueString)
```

## Checklist

### DIFF
--- a/Sources/Extensions/Foundation/FileManagerExtensions.swift
+++ b/Sources/Extensions/Foundation/FileManagerExtensions.swift
@@ -56,5 +56,28 @@ public extension FileManager {
         return nil
     }
 
+    /// Creates a unique directory for saving temporary files.
+    ///
+    /// The directory can be used to create multiple temporary files used for a common purpose.
+    ///
+    ///     let tempDirectory = try fileManager.temporaryFileDirectory()
+    ///     let tempFile1URL = tempDirectory.appendingPathComponent(ProcessInfo().globallyUniqueString)
+    ///     let tempFile2URL = tempDirectory.appendingPathComponent(ProcessInfo().globallyUniqueString)
+    ///
+    /// - Returns: A URL to a new directory for saving temporary files.
+    /// - Throws: An error if a temporary directory cannot be found or created.
+    public func temporaryFileDirectory() throws -> URL {
+        let temporaryDirectoryURL: URL
+        if #available(iOS 10.0, OSX 10.12, tvOS 10.0, watchOS 3.0, *) {
+            temporaryDirectoryURL = temporaryDirectory
+        } else {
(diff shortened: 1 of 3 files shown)


guykogus:
@LucianoPAlmeida I was basically following [this NSHipster article](https://nshipster.com/temporary-files/), as well as Apple's `NSTemporaryDirectory` documentation, which says:

> See the `FileManager` method `url(for:in:appropriateFor:create:)` for the preferred means of finding the correct temporary directory.

The idea is that the root temp directory isn't really where you want to save temp files. You're better off creating a temp folder so that you can clean it off in one go. If we just use `temporaryDirectory` then trying to remove the folder will throw an error since the user won't have permission to delete it.

It's also meant to ensure that you're saving to the right volume so that if you want to move the file to a permanent location then the transfer is faster and smoother. I think that this is more important for macOS than iOS, since iOS will generally just have a single volume and you're always working in the sandbox.

Sources/Extensions/Foundation/FileManagerExtensions.swift
```diff
@@ -56,5 +56,28 @@ public extension FileManager {
...
+    ///
+    /// The directory can be used to create multiple temporary files used for a common purpose.
+    ///
+    ///     let tempDirectory = try fileManager.temporaryFileDirectory()
+    ///     let tempFile1URL = tempDirectory.appendingPathComponent(ProcessInfo().globallyUniqueString)
+    ///     let tempFile2URL = tempDirectory.appendingPathComponent(ProcessInfo().globallyUniqueString)
+    ///
+    /// - Returns: A URL to a new directory for saving temporary files.
+    /// - Throws: An error if a temporary directory cannot be found or created.
+    public func temporaryFileDirectory() throws -> URL {
```
LucianoPAlmeida: My only concern here is that the naming doens't express the idea that a new directory is going to be created every time you call this. Even with just being a method kinda give this idea, I'm not sure if it's clear enough. What do you think? Maybe `makeTemporaryFileDirectory`?

guykogus: I was thinking a similar thing. How about `createTemporaryDirectory`? That would make it in line with the `create` parameter of `url(for:in:appropriateFor:create:)`.

PR #439, opened 2018-04-10
update UITableView and UICollectionView Extensions

jdisho:
In this PR I am updating:
`dequeueReusableCell(withClass:for)`,
`dequeueReusableHeaderFooterView(withClass)`,
`dequeueReusableCell(withClass)` in **UITableViewExtentions** and  `dequeueReusableCell(withClass:for)`,
`dequeueReusableSupplementaryView(ofKind:withClass:for)` in **UICollectionViewExtentions** 
to return not an optional value. In my opinion is better to get a `fatalError(...)` if the nib file could not be found.

## Checklist

- [x] I checked the [**Contributing Guidelines**](https://github.com/SwifterSwift/SwifterSwift/blob/master/CONTRIBUTING.md) before creating this request.
- [x] New extensions are written in Swift 4.
- [x] New extensions support iOS 8.0+ / tvOS 9.0+ / macOS 10.10+ / watchOS 2.0+.
- [x] I have added tests for new extensions, and they passed.
- [x] All extensions have a **clear** comments explaining their functionality, all parameters and return type in English.
- [x] All extensions are declared as **public**.
- [x] I have added a [changelog](https://github.com/SwifterSwift/SwifterSwift/blob/master/CHANGELOG_GUIDELINES.md) entry describing my changes.

PR #629, opened 2019-01-27
Added MKMapViewExtensions

staffler-xyz:
Similar to `UITableView`, `MapView` should have the same methods for registering and dequeueing. What do you think about this?

## Checklist

### DIFF
--- a/Sources/Extensions/MapKit/MKMapViewExtensions.swift
+++ b/Sources/Extensions/MapKit/MKMapViewExtensions.swift
@@ -25,7 +25,7 @@ public extension MKMapView {
     /// SwifterSwift: Register MKAnnotationView using class type
     ///
     /// - Parameter name: MKAnnotationView type.
-    @available(iOS 11.0, *, tvOS 11.0, *, macOS 10.13)
+    @available(iOS 11.0, tvOS 11.0, macOS 10.13, *)
     public func register<T: MKAnnotationView>(annotationViewWithClass name: T.Type) {
         register(T.self, forAnnotationViewWithReuseIdentifier: String(describing: name))
     }
@@ -36,7 +36,7 @@ public extension MKMapView {
     ///   - name: MKAnnotationView type.
     ///   - annotation: annotation of the mapView.
     /// - Returns: optional MKAnnotationView object.
-    @available(iOS 11.0, *, tvOS 11.0, *, macOS 10.13)
+    @available(iOS 11.0, tvOS 11.0, macOS 10.13, *)
     public func dequeueReusableAnnotationView<T: MKAnnotationView>(withClass name: T.Type, for annotation: MKAnnotation) -> T? {
         guard let annotationView = dequeueReusableAnnotationView(withIdentifier: String(describing: name), for: annotation) as? T else {
             fatalError("Couldn't find MKAnnotationView for \(String(describing: name))")
--- a/Sources/Extensions/MapKit/MKMapViewExtensions.swift
+++ b/Sources/Extensions/MapKit/MKMapViewExtensions.swift
@@ -9,9 +9,8 @@
 #if canImport(MapKit)
 import MapKit
 
-// MARK: - Initializers
 #if !os(watchOS)
-@available(iOS 3.0, *, tvOS 9.2, *, macOS 10.9)
+@available(tvOS 9.2, *)
 public extension MKMapView {
 
     /// SwifterSwift: Dequeue reusable MKAnnotationView using class type
--- a/Sources/Extensions/MapKit/MKMapViewExtensions.swift
+++ b/Sources/Extensions/MapKit/MKMapViewExtensions.swift
@@ -1,0 +1,52 @@
+//
+//  MKMapViewExtensions.swift
+//  SwifterSwift
+//
+//  Created by Hannes Staffler on 24.01.19.
+//  Copyright © 2019 SwifterSwift
+//
+
+#if canImport(MapKit)
+import MapKit
+
+// MARK: - Initializers
+#if !os(watchOS)
+@available(iOS 3.0, *, tvOS 9.2, *, macOS 10.9)
+public extension MKMapView {
+
+    /// SwifterSwift: Dequeue reusable MKAnnotationView using class type
+    ///
+    /// - Parameters:
+    ///   - name: MKAnnotationView type.
+    /// - Returns: optional MKAnnotationView object.
+    public func dequeueReusableAnnotationView<T: MKAnnotationView>(withClass name: T.Type) -> T? {
+        return dequeueReusableAnnotationView(withIdentifier: String(describing: name)) as? T
+    }
+
+    /// SwifterSwift: Register MKAnnotationView using class type
+    ///
+    /// - Parameter name: MKAnnotationView type.
+    @available(iOS 11.0, *, tvOS 11.0, *, macOS 10.13)
+    public func register<T: MKAnnotationView>(annotationViewWithClass name: T.Type) {
+        register(T.self, forAnnotationViewWithReuseIdentifier: String(describing: name))
+    }
+
+    /// SwifterSwift: Dequeue reusable MKAnnotationView using class type
+    ///
+    /// - Parameters:
+    ///   - name: MKAnnotationView type.
+    ///   - annotation: annotation of the mapView.
+    /// - Returns: optional MKAnnotationView object.
+    @available(iOS 11.0, *, tvOS 11.0, *, macOS 10.13)
+    public func dequeueReusableAnnotationView<T: MKAnnotationView>(withClass name: T.Type, for annotation: MKAnnotation) -> T? {
+        guard let annotationView = dequeueReusableAnnotationView(withIdentifier: String(describing: name), for: annotation) as? T else {
+            fatalError("Couldn't find MKAnnotationView for \(String(describing: name))")
+        }
+
+        return annotationView
+    }
+
+}
+#endif
+
+#endif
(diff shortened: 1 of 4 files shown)


Sources/Extensions/MapKit/MKMapViewExtensions.swift
```diff
@@ -0,0 +1,47 @@
...
+//
+
+#if canImport(MapKit)
+import MapKit
+
+// MARK: - Initializers
+#if !os(watchOS)
+public extension MKMapView {
+
+    /// SwifterSwift: Register MKAnnotationView using class name
```
LucianoPAlmeida: The description wouldn't be class type instead of class name?

Sources/Extensions/MapKit/MKMapViewExtensions.swift
```diff
@@ -0,0 +1,47 @@
...
+
+    /// SwifterSwift: Dequeue reusable MKAnnotationView using class name
+    ///
+    /// - Parameters:
+    ///   - name: MKAnnotationView type.
+    ///   - annotation: annotation of the mapView.
+    /// - Returns: optional MKAnnotationView object.
+    @available(iOS 11.0, *)
+    public func dequeueReusableAnnotationView<T: MKAnnotationView>(withClass name: T.Type, for annotation: MKAnnotation) -> T? {
+        return dequeueReusableAnnotationView(withIdentifier: String(describing: name), for: annotation) as? T
```
LucianoPAlmeida: Should we trap here instead of returning an option like #439? 
The one above is fine because `dequeueReusableAnnotationView(withIdentifier: )` returns optional already. 
But this one maybe could be an option :))

guykogus:
@staffler-xyz will you please look at the comments?

SwifterSwift.xcodeproj/project.pbxproj
```diff
@@ -333,13 +333,6 @@
 		278CA0911F9A9679004918BD /* NSImageExtensionsTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 278CA0901F9A9679004918BD /* NSImageExtensionsTests.swift */; };
 		42F53FEC2039C5AC0070DC11 /* UIStackViewExtensions.swift in Sources */ = {isa = PBXBuildFile; fileRef = 42F53FEB2039C5AC0070DC11 /* UIStackViewExtensions.swift */; };
 		42F53FF02039C7140070DC11 /* UIStackViewExtensionsTest.swift in Sources */ = {isa = PBXBuildFile; fileRef = 42F53FEF2039C7140070DC11 /* UIStackViewExtensionsTest.swift */; };
-		664CB983217243D200FC87B4 /* DispatchQueueExtensions.swift in Sources */ = {isa = PBXBuildFile; fileRef = 664CB982217243D200FC87B4 /* DispatchQueueExtensions.swift */; };
```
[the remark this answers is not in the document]
staffler-xyz: appearently XCode is reordering the names by alphabet when I add new files. I tried to reset the project and re-add them several times, but I got the same output every time.

staffler-xyz:
Sorry for the delay -> PR was updated just now :)

Sources/Extensions/MapKit/MKMapViewExtensions.swift
```diff
@@ -0,0 +1,52 @@
...
+//  Created by Hannes Staffler on 24.01.19.
+//  Copyright © 2019 SwifterSwift
+//
+
+#if canImport(MapKit)
+import MapKit
+
+// MARK: - Initializers
+#if !os(watchOS)
+@available(iOS 3.0, *, tvOS 9.2, *, macOS 10.9)
```
guykogus:
```suggestion
@available(iOS 3.0, tvOS 9.2, macOS 10.9, *)
```

SwifterSwift.xcodeproj/project.pbxproj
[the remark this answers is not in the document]
guykogus: Removing and adding the files won't reset them to their original state, it will continuously set a new random key. You need to fix these manually using a text editor .

Tests/MapKitTests/MKMapViewTests.swift
```diff
@@ -0,0 +1,36 @@
...
+
+import XCTest
+@testable import SwifterSwift
+
+#if canImport(MapKit)
+import MapKit
+
+import struct CoreLocation.CLLocationCoordinate2D
+
+final class MKMapViewTests: XCTestCase {
```
guykogus: The conditional compilation and attributes should match the source file.
```suggestion
#if !os(watchOS)
@available(iOS 3.0, tvOS 9.2, macOS 10.9, *)
final class MKMapViewTests: XCTestCase {
```
Don't forget the closing `#endif`

Sources/Extensions/MapKit/MKMapViewExtensions.swift
```diff
@@ -0,0 +1,52 @@
...
+//  SwifterSwift
+//
+//  Created by Hannes Staffler on 24.01.19.
+//  Copyright © 2019 SwifterSwift
+//
+
+#if canImport(MapKit)
+import MapKit
+
+// MARK: - Initializers
```
guykogus: Remove this

Tests/MapKitTests/MKMapViewTests.swift
```diff
@@ -0,0 +1,36 @@
...
+
+    func testRegisterAndDequeue() {
```
guykogus: We generally want 1 test per function, although here it kind of makes sense to combine them. Nevertheless, can you please split them into 2 functions, `testDequeue` and `testRegisterAndDequeue`, where the first one tests just the `dequeueReusableAnnotationView(withClass:)` function so that it doesn't need the `@available` attribute.

Tests/MapKitTests/MKMapViewTests.swift
```diff
@@ -0,0 +1,36 @@
...
+        guard #available(iOS 11.0, *, tvOS 11.0, *, macOS 10.13) else {
```
guykogus: Better to use `@available` on the function than guarding like this.

Tests/MapKitTests/MKMapViewTests.swift
```diff
@@ -0,0 +1,36 @@
...
+            return
+        }
+
+        let mapView = MKMapView()
+        let annotation = MKPlacemark(coordinate: CLLocationCoordinate2DMake(0, 0))
```
guykogus: `CLLocationCoordinate2DMake` is soooo Objective-C 😜
```suggestion
        let annotation = MKPlacemark()
```
This change will also mean that you can remove `import struct CoreLocation.CLLocationCoordinate2D`

staffler-xyz:
Hi! Thanks for the feedback @guykogus ! I updated the PR. I also noticed that this gives an error in SwiftLint:
```
@available(iOS 3.0, tvOS 9.2, macOS 10.9, *)
```
It should be enough to use this, as the deployment target is already set higher:
```
@available(tvOS 9.2, *)
```

Sources/Extensions/MapKit/MKMapViewExtensions.swift
```diff
@@ -0,0 +1,51 @@
...
+    ///   - name: MKAnnotationView type.
+    /// - Returns: optional MKAnnotationView object.
+    public func dequeueReusableAnnotationView<T: MKAnnotationView>(withClass name: T.Type) -> T? {
+        return dequeueReusableAnnotationView(withIdentifier: String(describing: name)) as? T
+    }
+
+    /// SwifterSwift: Register MKAnnotationView using class type
+    ///
+    /// - Parameter name: MKAnnotationView type.
+    @available(iOS 11.0, *, tvOS 11.0, *, macOS 10.13)
```
guykogus:
```suggestion
    @available(iOS 11.0, tvOS 11.0, macOS 10.13, *)
```

Sources/Extensions/MapKit/MKMapViewExtensions.swift
```diff
@@ -0,0 +1,51 @@
...
+        register(T.self, forAnnotationViewWithReuseIdentifier: String(describing: name))
+    }
+
+    /// SwifterSwift: Dequeue reusable MKAnnotationView using class type
+    ///
+    /// - Parameters:
+    ///   - name: MKAnnotationView type.
+    ///   - annotation: annotation of the mapView.
+    /// - Returns: optional MKAnnotationView object.
+    @available(iOS 11.0, *, tvOS 11.0, *, macOS 10.13)
```
guykogus:
```suggestion
    @available(iOS 11.0, tvOS 11.0, macOS 10.13, *)
```

Tests/MapKitTests/MKMapViewTests.swift
```diff
@@ -0,0 +1,42 @@
...
+@available(tvOS 9.2, *)
+final class MKMapViewTests: XCTestCase {
+
+    func testRegister() {
+        let mapView = MKMapView()
+
+        mapView.register(annotationViewWithClass: MKPinAnnotationView.self)
+    }
+
+    @available(iOS 11.0, *, tvOS 11.0, *, macOS 10.13)
```
guykogus:
```suggestion
    @available(iOS 11.0, tvOS 11.0, macOS 10.13, *)
```

Tests/MapKitTests/MKMapViewTests.swift
```diff
@@ -0,0 +1,42 @@
...
+import struct CoreLocation.CLLocationCoordinate2D
+
+#if !os(watchOS)
+@available(tvOS 9.2, *)
+final class MKMapViewTests: XCTestCase {
+
+    func testRegister() {
+        let mapView = MKMapView()
+
+        mapView.register(annotationViewWithClass: MKPinAnnotationView.self)
```
guykogus: You aren't actually testing anything in here. You can use the same test as in the other function:
```Swift
let annotationView = mapView.dequeueReusableAnnotationView(withClass: MKPinAnnotationView.self)
XCTAssertNotNil(annotationView)
```

Tests/MapKitTests/MKMapViewTests.swift
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    @available(iOS 11.0, *, tvOS 11.0, *, macOS 10.13)
+    func testRegisterAndDequeue() {
+        let mapView = MKMapView()
+        let annotation = MKPlacemark(coordinate: CLLocationCoordinate2D(latitude: 0, longitude: 0))
```
guykogus: Why can't you just replace the call with:
```suggestion
        let annotation = MKPlacemark()
```
which does the same thing, and then you can remove the `import`?

staffler-xyz: tried it, but i get a `EXC_BAD_ACCESS` in `dequeueReusableAnnotationView`. Seems like it requires the annotation to have a lat/lng.

guykogus: You're right. I was sure that I ran that in a playground and it worked 🤷‍♂️

Tests/MapKitTests/MKMapViewTests.swift
staffler-xyz: ok, updated :)

PR #634, opened 2019-02-23
Making some array extensions, generic collection extensions

LucianoPAlmeida:
Making keep(while: ), take(while: ) and skip(while:) array extensions, collection extensions.

Just want to use them with other types 😅 

## Checklist

### DIFF
--- a/Sources/Extensions/SwiftStdlib/RangeReplaceableCollectionExtensions.swift
+++ b/Sources/Extensions/SwiftStdlib/RangeReplaceableCollectionExtensions.swift
@@ -6,10 +6,6 @@
 //  Copyright © 2018 SwifterSwift
 //
 
-#if canImport(Foundation)
-import Foundation
-#endif
-
 // MARK: - Initializers
 extension RangeReplaceableCollection {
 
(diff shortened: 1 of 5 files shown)


Sources/Extensions/SwiftStdlib/RangeReplaceableCollectionExtensions.swift
```diff
@@ -91,12 +87,64 @@ extension RangeReplaceableCollection {
...
+
+    /// SwifterSwift: Keep elements of Array while condition is true.
+    ///
+    ///        [0, 2, 4, 7].keep(while: { $0 % 2 == 0 }) -> [0, 2, 4]
+    ///
+    /// - Parameter condition: condition to evaluate each element against.
+    /// - Returns: self after applying provided condition.
+    /// - Throws: provided condition exception.
+    @discardableResult
+    public mutating func keep(while condition: (Element) throws -> Bool) rethrows -> Self {
```
guykogus: This can be simplified with:
```Swift
public mutating func keep(while condition: (Element) throws -> Bool) rethrows -> Self {
    if let lastIndex = try firstIndex(where: { try !condition($0) }) {
        removeSubrange(lastIndex...)
    }
    return self
}
```
I also don't see why a `mutating` function should return itself, but no need to break backwards compatibility.

Sources/Extensions/SwiftStdlib/RangeReplaceableCollectionExtensions.swift
```diff
@@ -91,12 +87,64 @@ extension RangeReplaceableCollection {
...
+        return self
+    }
+
+    /// SwifterSwift: Take element of Array while condition is true.
+    ///
+    ///        [0, 2, 4, 7, 6, 8].take( where: {$0 % 2 == 0}) -> [0, 2, 4]
+    ///
+    /// - Parameter condition: condition to evaluate each element against.
+    /// - Returns: All elements up until condition evaluates to false.
+    public func take(while condition: (Element) throws -> Bool) rethrows -> Self {
```
guykogus: This is basically `prefix(while:)` and can be simplified with:
```Swift
public func take(while condition: (Element) throws -> Bool) rethrows -> Self {
    return Self(try prefix(while: condition))
}
```

Sources/Extensions/SwiftStdlib/RangeReplaceableCollectionExtensions.swift
```diff
@@ -91,12 +87,64 @@ extension RangeReplaceableCollection {
...
+        return self
+    }
+
+    /// SwifterSwift: Skip elements of Array while condition is true.
+    ///
+    ///        [0, 2, 4, 7, 6, 8].skip( where: {$0 % 2 == 0}) -> [6, 8]
+    ///
+    /// - Parameter condition: condition to evaluate each element against.
+    /// - Returns: All elements after the condition evaluates to false.
+    public func skip(while condition: (Element) throws-> Bool) rethrows -> Self {
```
guykogus: This can be simplified with:
```Swift
public func skip(while condition: (Element) throws-> Bool) rethrows -> Self {
    let idx = try firstIndex(where: { try !condition($0) }) ?? startIndex
    return Self(self[idx...])
}
```

Sources/Extensions/SwiftStdlib/RangeReplaceableCollectionExtensions.swift
LucianoPAlmeida: You're right, so maybe we should deprecate it in the future :)

PR #647, opened 2019-03-29
Add conformance for Optional to Comparable

guykogus:
It would be nice if we were able to compare `Optional` values, like so:
```Swift
struct Player {
    let name: String
    let score: Int?
}

let players = [Player(name: "Omar", score: 10),
               Player(name: "Steven", score: 20),
               Player(name: "Guy", score: nil)]
players.sorted(by: { $0.score < $1.score }) // Guy, Omar, Steven
```

Does this make sense to do? Is there a situation where this doesn't make sense?
Also, I can't mark the extension as `public`, because I get the following error:
> 'public' modifier cannot be used with extensions that declare protocol conformances

## Checklist

### DIFF
--- a/Tests/SwiftStdlibTests/OptionalExtensionsTests.swift
+++ b/Tests/SwiftStdlibTests/OptionalExtensionsTests.swift
@@ -116,4 +116,39 @@ final class OptionalExtensionsTests: XCTestCase {
         XCTAssertEqual(collection.nonEmpty!, [1, 2, 3])
     }
 
+    func testComparable() {
+        let minusOne: Int? = -1
+        let zero: Int? = 0
+        let one: Int? = 1
+
+        // Int?.some vs Int
+        XCTAssert(minusOne < 0)
+        XCTAssert(zero < 1)
+        XCTAssert(one < 2)
+        XCTAssert(-2 < minusOne)
+        XCTAssert(-1 < zero)
+        XCTAssert(0 < one)
+
+        // Int?.none vs Int
+        XCTAssert(nil < -1)
+        XCTAssert(nil < 0)
+        XCTAssert(nil < 1)
+        XCTAssertFalse(-1 < nil)
+        XCTAssertFalse(0 < nil)
+        XCTAssertFalse(1 < nil)
+
+        // Int?.some vs Int?.some
+        XCTAssert(minusOne < zero)
+        XCTAssert(minusOne < one)
+        XCTAssert(zero < one)
+
+        // Int?.some vs Int?.none
+        XCTAssertFalse(minusOne < nil)
+        XCTAssertFalse(zero < nil)
+        XCTAssertFalse(one < nil)
+        XCTAssert(nil < minusOne)
+        XCTAssert(nil < zero)
+        XCTAssert(nil < one)
+    }
+
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 ## Upcoming Release
 
 ### Added
+- **Optional**
+  - Conformance to `Comparable`. [#647](https://github.com/SwifterSwift/SwifterSwift/pull/647) by [guykogus](https://github.com/guykogus)
 
 ### Changed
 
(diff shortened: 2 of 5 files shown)


guykogus:
To add to my question of whether or not this makes sense: This functionality was available in python 2 but removed in python 3. There's a bit of an explanation [in this SO question](https://stackoverflow.com/questions/2214194/is-everything-greater-than-none). They were trying to be a bit more strict with comparisons in Python 3 because previously you could compare things that didn't make sense. I think that here it does.

Sources/Extensions/SwiftStdlib/OptionalExtensions.swift
```diff
@@ -104,6 +104,21 @@ public extension Optional where Wrapped: Collection {
...
 }
 
+// MARK: - Methods (Comparable)
+
+extension Optional: Comparable where Wrapped: Comparable {
+    public static func < (lhs: Optional, rhs: Optional) -> Bool {
+        switch (lhs, rhs) {
+        case (.some(let lhsValue), .some(let rhsValue)):
+            return lhsValue < rhsValue
+        case (_, .none):
```
LucianoPAlmeida: @guykogus What's the criteria to return false and true in both those cases?

Tests/SwiftStdlibTests/OptionalExtensionsTests.swift
```diff
@@ -105,4 +105,39 @@ final class OptionalExtensionsTests: XCTestCase {
...
+        XCTAssert(one < 2)
+        XCTAssert(-2 < minusOne)
+        XCTAssert(-1 < zero)
+        XCTAssert(0 < one)
+
+        // Int?.none vs Int
+        XCTAssert(nil < -1)
+        XCTAssert(nil < 0)
+        XCTAssert(nil < 1)
+        XCTAssertFalse(-1 < nil)
```
LucianoPAlmeida: This really feels weird to me, I mean for correctness saying that an integer or any value is greater or lesser than nil, because, at least for me it can't be true or false because nil is not actually a value. Maybe that was part of the reason for this not to be on the stdlib, but I'm not sure, is just a feeling that I have ... so let me know your thoughts on that :))

Sources/Extensions/SwiftStdlib/OptionalExtensions.swift
guykogus: Do you mean why is it `false` if we have `.some < .none` and `.none < .none`? For me the idea was that nothing will ever be less than `.none`. Turns out that this is how Python 2 implements it, as well.

Tests/SwiftStdlibTests/OptionalExtensionsTests.swift
guykogus: I see it as `nil` being the absence of a value. Maybe it reads better if I use `Int?.none` instead of `nil`.
```Swift
XCTAssert(Int?.none < 0)
```

guykogus:
@LucianoPAlmeida look at the change I made in https://github.com/SwifterSwift/SwifterSwift/pull/647/commits/9bdc2deaf93badac6b4456edb6616c25e66095da. The only change I actually made to the unit test was to make Rose's age `nil`. If I had left the original implementation then the test would have failed because it wasn't handling comparison correctly with a `nil` value. Thanks to my new conformance implementation it works as expected. This is the kind of thing that we can avoid, by ensuring that `nil` values are correctly grouped together in comparisons.

Tests/SwiftStdlibTests/OptionalExtensionsTests.swift
LucianoPAlmeida: The issue is not how it reads better, but actually with semantics and correctness of comparing something to nil. Even as you mention, if we assume nothing is less than nil like you mention python does.

Sources/Extensions/SwiftStdlib/OptionalExtensions.swift
LucianoPAlmeida: Got it 👍 
Personally, I see the point of why it is useful, only concern is if that is correct considering the Swift perspective. By the way, the point you raised is making me think if it should be a `sort(keypath:)` where the key is optional. It is, in fact, useful but again not sure if it's correct.

Tests/SwiftStdlibTests/OptionalExtensionsTests.swift
guykogus: This is the tricky bit about the PR, why I wanted to get more opinions about it. Because yeah, semantically it doesn't make sense to compare to `nil` as a non-value, but `nil` as `Int?.none` can be semantically true. @SD10 @BennX @omaralbeik can I get your opinions about this too, please?

chulbert:
It's worth mentioning this was specifically removed in Swift 3:

https://github.com/apple/swift-evolution/blob/master/proposals/0121-remove-optional-comparison-operators.md

calebkleveter:
@chulbert True, but the proposal specifically states that it is because there wasn't conditional conformance. That exists now, so I think it would make sense to add it back in until it's added back to the Swift standard library.

guykogus:
@chulbert thanks for pointing that out. Interesting to see that they removed it because of missing conditional conformances, which of course we now have, so it doesn't sound like a bad idea to re-add it. However, the example provided is actually a great use case for a potential bug with the comparison. Still, I personally think it's worth it.

chulbert:
@guykogus @calebkleveter I think there were two main rationales.  First, as you state, was conditional protocol conformance and that's been resolved.  However, the second, that comparisons between optional and non-optional values is error-prone remains a concern.  I would feel uncomfortable making that decision for all consumers of SwifterSwift.

guykogus:
@chulbert exactly, the problem is that implementing this conformance means that it's automatically, silently included for any user including SwifterSwift.
I think I'm leaning against this now. I'm going to close the PR, if anyone can convince me I'm happy to re-open it :)

PR #648, opened 2019-03-30
v5.0.0

omaralbeik:
This PR brings Swift 5.0 support, Linux support, and many other improvements.
Refer to the CHANGELOG for further details.


Sources/Extensions/SwiftStdlib/BinaryFloatingPointExtensions.swift
```diff
@@ -6,9 +6,13 @@
 //  Copyright © 2018 SwifterSwift
 //
 
+#if canImport(Foundation)
+import Foundation
+
 // MARK: - Methods
 public extension BinaryFloatingPoint {
 
+    #if canImport(Foundation)
```
LucianoPAlmeida: Does it need this check? Since it don't use any foundation API only stdlib

omaralbeik: We need it because `pow` is coming from `Foundation` ¯\_(ツ)_/¯

Sources/SwifterSwift/SwiftStdlib/ArrayExtensions.swift
```diff
@@ -18,32 +15,74 @@ public extension Array {
...
-    public mutating func safeSwap(from index: Index, to otherIndex: Index) {
+    mutating func safeSwap(from index: Index, to otherIndex: Index) {
         guard index != otherIndex else { return }
         guard startIndex..<endIndex ~= index else { return }
         guard startIndex..<endIndex ~= otherIndex else { return }
         swapAt(index, otherIndex)
     }
 
-    /// SwifterSwift: Separates an array into 2 arrays based on a predicate.
+    /// SwifterSwift: Keep elements of array while condition is true.
```
LucianoPAlmeida: How this got added back? I remember this being moved to Collection.

omaralbeik: It could be that I added it by accident while updating from `master` 😅

PR #656, opened 2019-04-04
Add Array Extension for filtered

fanglinwei:
Add Array Extension for filtered

### DIFF
--- a/Sources/SwifterSwift/SwiftStdlib/ArrayExtensions.swift
+++ b/Sources/SwifterSwift/SwiftStdlib/ArrayExtensions.swift
@@ -163,31 +163,31 @@ public extension Array where Element: Equatable {
     }
 }
 
-extension Array where Element: Hashable {
+extension Array where Element: Equatable {
     
-    public func filtered<E: Equatable>(duplication path: KeyPath<Element, E>) -> [Element] {
-        return reduce(into: [Element]()) { (result, e) in
-            let contains = result.contains { $0[keyPath: path] == e[keyPath: path] }
-            result += contains ? [] : [e]
+    public func withoutDuplicates<E: Equatable>(by keyPath: KeyPath<Element, E>) -> [Element] {
+        return reduce(into: [Element]()) { (result, element) in
+            let contains = result.contains { $0[keyPath: keyPath] == element[keyPath: keyPath] }
+            result += contains ? [] : [element]
         }
     }
     
-    public func filtered<E: Equatable>(duplication closure: (Element) throws -> E) rethrows -> [Element] {
-        return try reduce(into: [Element]()) { (result, e) in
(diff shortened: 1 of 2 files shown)


Sources/SwifterSwift/SwiftStdlib/ArrayExtensions.swift
```diff
@@ -161,5 +161,34 @@ public extension Array where Element: Equatable {
             }
         }
     }
+}
 
+extension Array where Element: Hashable {
+    
+    public func filtered<E: Equatable>(duplication path: KeyPath<Element, E>) -> [Element] {
+        return reduce(into: [Element]()) { (result, e) in
```
guykogus: Don't use `e` as a variable name, make it more meaningful.
```suggestion
        return reduce(into: [Element]()) { (result, element) in
```

Sources/SwifterSwift/SwiftStdlib/ArrayExtensions.swift
```diff
@@ -161,5 +161,34 @@ public extension Array where Element: Equatable {
...
+            let contains = result.contains { $0[keyPath: path] == e[keyPath: path] }
+            result += contains ? [] : [e]
+        }
+    }
+    
+    public func filtered<E: Equatable>(duplication closure: (Element) throws -> E) rethrows -> [Element] {
```
guykogus: Try to avoid variable names like `array`, `string`, `closure`, etc. I think here a good name is `property` because you're getting a calculated property of the element.
```suggestion
    public func filtered<E: Equatable>(by property: (Element) throws -> E) rethrows -> [Element] {
```

Sources/SwifterSwift/SwiftStdlib/ArrayExtensions.swift
```diff
@@ -161,5 +161,34 @@ public extension Array where Element: Equatable {
...
+        return try reduce(into: [Element]()) { (result, e) in
```
guykogus: Again, use meaningful variable names.
```suggestion
        return try reduce(into: [Element]()) { (result, element) in
```

chulbert:
As I said in the other issue, I think the essential functionality of this proposal is _deduplication_ and not filtering.  As such, I would recommend this be named something like `withoutDuplicates(for keyPath:)`

Sources/SwifterSwift/SwiftStdlib/ArrayExtensions.swift
chulbert: Why not reduce into a dictionary keyed by keypath value to avoid the expense of contains?

LucianoPAlmeida:
just cleaning up those old PR's. 
@fanglinwei   Closing this one due to inactivity, feel free to reopen :))

PR #662, opened 2019-04-07
Add aspectRatio and max properties to CGSize

Vyax:
## Checklist

- [/] New extensions are written in Swift 4.
- [/] New extensions support iOS 8.0+ / tvOS 9.0+ / macOS 10.10+ / watchOS 2.0+.
- [/] I have added tests for new extensions, and they passed.
- [/] All extensions have a **clear** comments explaining their functionality, all parameters and return type in English.
- [/] All extensions are declared as **public**.

### DIFF
--- a/Sources/SwifterSwift/CoreGraphics/CGSizeExtensions.swift
+++ b/Sources/SwifterSwift/CoreGraphics/CGSizeExtensions.swift
@@ -20,6 +20,21 @@ import Cocoa
 // MARK: - Methods
 public extension CGSize {
 
+    /// SwifterSwift: Returns the aspect ratio.
+    var aspectRatio: CGFloat {
+        return height == 0 ? 0 : width / height
+    }
+
+    /// SwifterSwift: Returns width or height, whatever is the bigger value.
+    var max: CGFloat {
+        return Swift.max(width, height)
+    }
+
+}
+
+// MARK: - Methods
+public extension CGSize {
+
     /// SwifterSwift: Aspect fit CGSize.
     ///
     ///     let rect = CGSize(width: 120, height: 80)
--- a/Sources/SwifterSwift/CoreGraphics/CGSizeExtensions.swift
+++ b/Sources/SwifterSwift/CoreGraphics/CGSizeExtensions.swift
@@ -26,8 +26,13 @@ public extension CGSize {
     }
 
     /// SwifterSwift: Returns width or height, whatever is the bigger value.
-    var max: CGFloat {
-        return Swift.max(width, height)
+    var maxDimension: CGFloat {
+        return max(width, height)
+    }
+
+    /// SwifterSwift: Returns width or height, whatever is the smaller value.
+    var minDimension: CGFloat {
+        return min(width, height)
     }
 
 }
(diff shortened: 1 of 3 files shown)


Sources/SwifterSwift/CoreGraphics/CGSizeExtensions.swift
```diff
@@ -17,6 +17,21 @@ import UIKit
 import Cocoa
 #endif
 
+// MARK: - Methods
+public extension CGSize {
+
+    /// SwifterSwift: Returns the aspect ratio.
+    var aspectRatio: CGFloat {
```
LucianoPAlmeida: Nice one :))

Sources/SwifterSwift/CoreGraphics/CGSizeExtensions.swift
```diff
@@ -17,6 +17,21 @@ import UIKit
...
+        return height == 0 ? 0 : width / height
+    }
+
+    /// SwifterSwift: Returns width or height, whatever is the bigger value.
+    var max: CGFloat {
```
LucianoPAlmeida: Not sure about this tho, because calling that don't seem readable. Example `size.max` is the max size? I mean I couldn't tell just looking to the property or a call of that what it does. Don't know, let me know if you think it makes sense :)

Vyax: I agree. Do you think _maxDimension_ would be a better name? If we want to avoid _max_, we could name it _largerDimension_.

LucianoPAlmeida: `maxDimension` sounds better to me :))

chulbert: If you're going to add `maxDimension`, you might as well add `minDimension` for parity.

Vyax: Done :)

Sources/SwifterSwift/CoreGraphics/CGSizeExtensions.swift
```diff
@@ -17,6 +17,26 @@ import UIKit
...
 
+// MARK: - Methods
+public extension CGSize {
+
+    /// SwifterSwift: Returns the aspect ratio.
+    var aspectRatio: CGFloat {
+        return height == 0 ? 0 : width / height
+    }
+
+    /// SwifterSwift: Returns width or height, whatever is the bigger value.
```
guykogus:
```suggestion
    /// SwifterSwift: Returns `width` or `height`, whichever is the bigger value.
```

Sources/SwifterSwift/CoreGraphics/CGSizeExtensions.swift
```diff
@@ -17,6 +17,26 @@ import UIKit
...
+    var maxDimension: CGFloat {
+        return max(width, height)
+    }
+
+    /// SwifterSwift: Returns width or height, whatever is the smaller value.
```
guykogus:
```suggestion
    /// SwifterSwift: Returns `width` or `height`, whichever is the smaller value.
```

Vyax: Thanks

Sources/SwifterSwift/CoreGraphics/CGSizeExtensions.swift
Vyax: Thanks

Issue #676, opened 2019-05-20
Don't have access to RandomAccessCollection 's extension function `func indices(of item: Element)`


WeMadeCode:
> Please fill out this template when filing an issue.
>
> All comment lines beginning with an **>** instruct you with what info we expect. You can delete those lines once you've filled in the info.
>

[ ] I've read, understood, and done my best to follow the [Contributing guidelines](https://github.com/SwifterSwift/SwifterSwift/tree/master/CONTRIBUTING.md) before opening this issue.

### What did you do?
in viersion 4.6.0
I can compile `[1, 2, 2, 3, 4, 2, 5].indices(of: 2)`
but in version 5.0.0
i can't compile `[1, 2, 2, 3, 4, 2, 5].indices(of: 2)`

### What did you expect to happen?
have access to RandomAccessCollection 's extension function `func indices(of item: Element)`

### What happened instead?
Don't have access to RandomAccessCollection 's extension function `func indices(of item: Element)` 

### SwifterSwift Environment

- SwifterSwift version:5.0.0
- Xcode version:10.2.1
- macOS version running Xcode:
- Swift version:5.0
- Platform(s) running SwifterSwift: Simulator  iPhone Xr

### Demo Project
>Please link to or upload a project we can download that reproduces the issue.

PR #677, opened 2019-05-20
Fixing missing public modifiers

LucianoPAlmeida:
Fixing #676 
## Checklist


LucianoPAlmeida:
The pod lib lint was causing issues with the CI for quite some time, I'm not sure why but I think it should not be a big deal to just remove it for now, and check this later.
As far as I can check for the logs, Travis is aborting the build because it is taking too long without an output(10 min) since it was running silent. But if we actually put verbose to have an output it complains about too long logs and fails too. So ... ¯\_(ツ)_/¯ let's just remove it for now?
cc @omaralbeik

PR #387, opened 2018-02-02
Double extensions for units of measurement and duration string

jason-ingenuity:
## Checklist

- [x] I checked the [**Contributing Guidelines**](https://github.com/SwifterSwift/SwifterSwift/blob/master/CONTRIBUTING.md) before creating this request.
- [x] New extensions are written in Swift 4.
- [x] New extensions support iOS 8.0+ / tvOS 9.0+ / macOS 10.10+ / watchOS 2.0+.
- [x] I have added tests for new extensions, and they passed.
- [x] All extensions have a **clear** comments explaining their functionality, all parameters and return type in English.
- [x] All extensions are declared as **public**.
- [x] I have added a changelog entry describing my changes.

PR #680, opened 2019-05-24
Add temperature conversion to Int

ChristopherAaronBrown:
:rocket: Added C -> F and F -> C for Ints

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,8 +44,6 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 ### Deprecated
 
 ### Removed
-- **Character**:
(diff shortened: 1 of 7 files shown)


LucianoPAlmeida:
Hey @ChristopherAaronBrown Thanks to taking the time for that PR 👍 
But I think it is the same case of #387 where I think those are not a good fit for Int/Double extensions because of the following reasons:

* Not every Int represents a measure. So assuming that Int represents a temperature measure makes the API not concise. For example:
let distance: Int = 20
distance.fahrenheitToCelsius 
In the example above we can see that feels strange a Int in one level of abstraction that represents a distance have a property that converts it to celsius.

* Double is a trivial type that it's only responsibility is to represent a number and because of separation of concerns in API's it makes more sense separate measure or any other responsibility in other layers of abstraction :)

* Foundation already has a really great and simple API to deal with measurement. Example:

let measure = Measurement<UnitTemperature>(value: 20, unit: UnitTemperature.celsius) 
measure.converted(to: UnitLength.fahrenheit) 

TLDR; The reason is just to maintain the API concise and correct 
Again thank you for taking your time to submit this PR :))

ChristopherAaronBrown:
I see, that makes sense. Thanks for reviewing my pull request!

PR #681, opened 2019-05-24
Added a convenience to UIColor 

DrBeta:
## Checklist 🚀 

## Changes

My extension allows you to use UIColor without dividing every number. It uses a convenience init to do this. This is also my first time writing a test, and contributing Swift to a open source Project. Here is the before and after of the color white:

#### Before:

`UIColor(red: 255/255, green: 255/255, blue: 255/255, a: 1)` 

#### After:

`UIColor(red: 255, green: 255, blue: 255, a: 1)` 

I complied to all SwiftLint Guidelines in writing this code.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 
 ### Added
 - **UIColor**:
-    -You can now init UIColor with `UIColor(red: 255, green: 255, blue: 255, alpha: 1)`. That means no dividing!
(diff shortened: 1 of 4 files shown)


guykogus:
Hi @DrBeta, thanks for submitting this PR but we already have this functionality. In SwifterSwift we have the `ColorExtensions.swift` file, which handles the different colour classes used for each platform, e.g. in macOS and iOS the former uses `NSColor` and the latter uses `UIColor`. There you will find that we already have `init(red:green:blue:transparency:)`, which fulfils the purpose of your PR.

PR #685, opened 2019-06-03
Add isNilOrZero and nonZero properties to the optional type BinaryFloatingPoint and BinaryInteger

fanglinwei:
1. Add the`isZero ` property to BinaryInteger and Tests
2. Add the`isNilOrZero ` and`nonZero ` properties to the optional type BinaryFloatingPoint and Tests
3. Add the`isNilOrZero ` and`nonZero ` properties to the optional type BinaryInteger and Tests

## Checklist


Sources/SwifterSwift/SwiftStdlib/BinaryIntegerPointExtensions.swift
```diff
@@ -0,0 +1,28 @@
...
+//  Copyright © 2019 SwifterSwift
+//
+
+#if canImport(Foundation)
+import Foundation
+
+// MARK: - Properties
+public extension BinaryInteger {
+    
+    #if canImport(Foundation)
```
LucianoPAlmeida: Since we are not relying on any Foundation API, I think all of those #if canImport(Foundation) aren't necessary.

LucianoPAlmeida:
Hey @fanglinwei 
Thank's to taking the time, just two little things:
Can you add a changelog entry?
Can you run `swift test --generate-linuxmain` to update linux tests too?
Also a run `swiftlint autocorrect`

Sources/SwifterSwift/SwiftStdlib/BinaryIntegerPointExtensions.swift
fanglinwei: Yes, so I resubmitted one and removed those `#if canImport(Foundation)`😊

fanglinwei:
> Hey @fanglinwei
> Thank's to taking the time, just two little things:
> Can you add a changelog entry?
> Can you run `swift test --generate-linuxmain` to update linux tests too?
> Also a run `swiftlint autocorrect`

Ok, I got a ci test error, I am trying to modify

fanglinwei:
> Hey @fanglinwei
> Thank's to taking the time, just two little things:
> Can you add a changelog entry?
> Can you run `swift test --generate-linuxmain` to update linux tests too?
> Also a run `swiftlint autocorrect`

Ok, I got some test errors, I am trying to modify

LucianoPAlmeida:
Have you run swiftlint autocorrect? Danger is still pointing some lint warnings.

Sources/SwifterSwift/SwiftStdlib/OptionalExtensions.swift
```diff
@@ -111,6 +111,40 @@ public extension Optional where Wrapped: Collection {
...
+public extension Optional where Wrapped: BinaryInteger {
+    
+    /// SwifterSwift: Check if optional is nil or zero integer.
+    var isNilOrZero: Bool {
+        guard let integer = self else { return true }
+        return integer == 0
+    }
+    
+    /// SwifterSwift: Returns the integer only if it is not nill and not zero.
+    var nonZero: Wrapped? {
```
LucianoPAlmeida: The `isNilOrZero` fine but, can you provide some use cases for this one? I imagine it can be used for `if let i = optionalInt.nonZero { }` but I think `if let int = optionalInt, !int.isZero { }` serve the same purpose and is more clear, so I'm not sure about this. What you think?

guykogus:
@fanglinwei I'm curious about the use cases for these extensions. Where and how are you using them? Why is `someInt.isZero` preferable to `someInt == 0`?

Sources/SwifterSwift/SwiftStdlib/OptionalExtensions.swift
fanglinwei: Sorry, I may not use swiftlint autocorrection to eliminate these space warnings.😅
I think`isNilOrZero` still has to exist, just like`isNilOrEmpty` is on`String`
I don't know why I have to add a file to add the`isZero` property to`BinaryIntegerExtensions`, there will always be errors about `Codecov changes.`
I tried to solve it in the past, but it didn't work.
I am sorry that I am not good at this aspect of the pr code, Can you help me?

chulbert: Agree.  Equating optionality and a magic value seems very un-Swifty.  If you want to perform the same action when a value is missing or when it is zero, that seems more like business logic than a general purpose extension to me.

fanglinwei:
@guykogus 
Because there is an isZero attribute in the standard library `BinaryFloatingPoint` protocol, I think there can be one in BinaryInteger. What do you think

Sources/SwifterSwift/SwiftStdlib/OptionalExtensions.swift
fanglinwei: @chulbert 
You have some facts that the judgment of 0 and null does have more to do with business logic.😄

LucianoPAlmeida: > I think`isNilOrZero` still has to exist, just like`isNilOrEmpty` is on`String`

I agree, I think is fine, the one I was talking is `nonZero` one :)) 



> I don't know why I have to add a file to add the`isZero` property to`BinaryIntegerExtensions`, there will always be errors about `Codecov changes.`

I don't think you have to add another file, can you provide more info about the error? 


> I am sorry that I am not good at this aspect of the pr code, Can you help me?

Sure :)) Just let us know what the error was and we can figure out together

LucianoPAlmeida:
@guykogus I think we fall into the same case as why people prefer `array.isEmpty` instead or `array.count == 0` 😆

guykogus:
@fanglinwei I like the balancing of `isZero` for integers, but I don't get the purpose of the `Optional` extensions. I was considering once something similar for `Optional where Wrapped == Bool` to have `isTrueOrNil` and similar functions, but I ended up finding it more confusing than testing `foo == true`, `foo != false`, etc. That may be a personal taste, though. But it seems to me that if you have `isNilOrZero` then you should also have `isNotNilOrZero`, `isNilOrNotZero`, `isNotNilOrNotZero`, otherwise it's incomplete. I don't think it's a rabbit hole down which you want to go.

@LucianoPAlmeida using `isEmpty` is recommended according Apple's documentation. For an `Array` it doesn't matter, but if you want to change your `Collection` type you shouldn't have to go and replace `.count == 0` to `.isEmpty`.

> When you need to check whether your collection is empty, use the `isEmpty` property instead of checking that the` count` property is equal to zero. For collections that don’t conform to `RandomAccessCollection`, accessing the `count` property iterates through the elements of the collection.

fanglinwei:
@guykogus 
I think the`bool` type has only simple`true` and`false`, but for numeric types, it has positive, negative, and zero cases.
Just like the standard library provides`iszero` for the`BinaryFloatingPoint` protocol, but there is no`iszero` attribute for`BinaryInteger`. I think the`Int` type needs such a property. What do you think?

Sources/SwifterSwift/SwiftStdlib/OptionalExtensions.swift
fanglinwei: I think`isZero` is added for the`BinaryInteger` protocol. When I looked at the original file structure, I didn't find a file about`BinaryIntegerExtensions`. If I don't add a new file, I should put it there.

fanglinwei: @LucianoPAlmeida 
It seems that`isNilOrZero` and`nonZero` are really more biased towards the business module, I think that`swifterswift` has `Optional where Wrapped == Collection` to expand the`isNilOrEmpty` and`nonEmpty` properties, then you should need to add`isNilOrZero` and`nonZero` for `Optional where Wrapped == BinaryFloatingPoint` Attributes, in fact, the discussion of`isNilOrEmpty`,`nonEmpty`,`isNilOrZero``nonZero` and are more biased towards business logic. What do you think?

LucianoPAlmeida: Sorry, just note that I've missed a lot of discussions lately here 😅, but if you still like to continue from here.
>  Attributes, in fact, the discussion of`isNilOrEmpty`,`nonEmpty`,`` isNilOrZero``nonZero `` and are more biased towards business logic. What do you think?

I just think that maybe those are over abstracting very simple operations, although some people may think that is more readable and more business logic as you said. I'm not sure is a good fit for here, since we try to gather general propose extension that may be useful in a general way.

Also, if I miss the notification here on GitHub for to much time(like a week or two), feel free to ping me on the SwifterSwift slack :))

PR #690, opened 2019-06-26
Fix UIImageView download

novpeter:
### Fixed
- Fix UIImageView download method. If it's deallocated, this causes a crash when call **self**.


guykogus:
Can you please add a CHANGELOG entry?

PR #704, opened 2019-08-15
Array.withoutDuplicates(keyPath:)

### LINKED ISSUE #653: Add Array Extension for filtered on the  KeyPath
```swift
public extension Array where Element: Equatable {
    
     // Returns a deduplicated array based on the optional key path.
    public func filtered<E: Equatable>(duplication path: KeyPath<Element, E>) -> [Element] {
        return reduce(into: [Element]()) { (result, e) in
            let contains = result.contains { $0[keyPath: path] == e[keyPath: path] }
            result += contains ? [] : [e]
        }
    }
}
```

RomanPodymov:
🚀 Fixed #653. Added **withoutDuplicates(keyPath:)** to **Array**.
## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 
 ### Added
 - **Array**:
-    - Added `removeDuplicates(by:)`, `removeDuplicates(keyPath:)`, `withoutDuplicates(by:)`, `withoutDuplicates(keyPath:)` for remove duplicate elements based on comparator or key path. [#703](https://github.com/SwifterSwift/SwifterSwift/pull/703) by [RomanPodymov](https://github.com/RomanPodymov)
+    - Added `removeDuplicates(by:)`, `removeDuplicates(keyPath:)`, `withoutDuplicates(by:)`, `withoutDuplicates(keyPath:)` for remove duplicate elements based on comparator or key path. [#704](https://github.com/SwifterSwift/SwifterSwift/pull/704) by [RomanPodymov](https://github.com/RomanPodymov).
 - **Sequence**:
     - Added `withoutDuplicates(transform:)` for remove duplicate elements based on condition in a sequence. [#666](https://github.com/SwifterSwift/SwifterSwift/pull/666) by [saucym](https://github.com/saucym)
 - **String**
@@ -44,6 +44,8 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
   - Added `init(from:to:)`, `init(points:)`, `init(polygonWithPoints:)`, `init(ovalOf:centered:)` and `init(rectOf:centered:)` convenience initializers. [#659](https://github.com/SwifterSwift/SwifterSwift/pull/659) by [vyax](https://github.com/vyax)
 
 ### Changed
+- **Array**:
+  - - `removeDuplicates` and `withoutDuplicates` are implemented based on `removeDuplicates(by:)` and `withoutDuplicates(by:)`. [#704](https://github.com/SwifterSwift/SwifterSwift/pull/704) by [RomanPodymov](https://github.com/RomanPodymov).
 - **UIApplication**:
   - - `queryValue(for:)` extension for URL is refactored. [#668](https://github.com/SwifterSwift/SwifterSwift/pull/668) by [LucianoPAlmeida](https://github.com/ratulSharker).
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 
 ### Added
 - **Array**:
-    - Added `removeDuplicates(by:)`, `removeDuplicates(keyPath:)`, `withoutDuplicates(by:)`, `withoutDuplicates(keyPath:)` for remove duplicate elements based on comparator or key path. [#704](https://github.com/SwifterSwift/SwifterSwift/pull/704) by [RomanPodymov](https://github.com/RomanPodymov).
+    - Added `removeDuplicates(by:)`, `removeDuplicates(keyPath:)`, `withoutDuplicates(by:)`, `withoutDuplicates(keyPath:)` for remove duplicate elements based on comparator or key path. Different implementations for `Element: Equatable` and `Element: Hashable`. [#704](https://github.com/SwifterSwift/SwifterSwift/pull/704) by [RomanPodymov](https://github.com/RomanPodymov).
 - **Sequence**:
     - Added `withoutDuplicates(transform:)` for remove duplicate elements based on condition in a sequence. [#666](https://github.com/SwifterSwift/SwifterSwift/pull/666) by [saucym](https://github.com/saucym)
 - **String**
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,9 +5,6 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 ## Upcoming Release
 
 ### Added
-- **Array**:
-    - Added `removeDuplicates(by:)`, `removeDuplicates(keyPath:)`, `withoutDuplicates(by:)`, `withoutDuplicates(keyPath:)` for remove duplicate elements based on comparator or key path. [#704](https://github.com/SwifterSwift/SwifterSwift/pull/704) by [RomanPodymov](https://github.com/RomanPodymov).
-    - Added `removeDuplicates()`, `withoutDuplicates()` where `Element: Hashable` for remove duplicate elements. [#704](https://github.com/SwifterSwift/SwifterSwift/pull/704) by [RomanPodymov](https://github.com/RomanPodymov).
 - **Sequence**:
     - Added `withoutDuplicates(transform:)` for remove duplicate elements based on condition in a sequence. [#666](https://github.com/SwifterSwift/SwifterSwift/pull/666) by [saucym](https://github.com/saucym)
 - **String**
@@ -46,8 +43,6 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
   - Added `init(from:to:)`, `init(points:)`, `init(polygonWithPoints:)`, `init(ovalOf:centered:)` and `init(rectOf:centered:)` convenience initializers. [#659](https://github.com/SwifterSwift/SwifterSwift/pull/659) by [vyax](https://github.com/vyax)
 
 ### Changed
-- **Array**:
-  - - `removeDuplicates` and `withoutDuplicates` where `Element: Equatable` are implemented based on `removeDuplicates(by:)` and `withoutDuplicates(by:)`. [#704](https://github.com/SwifterSwift/SwifterSwift/pull/704) by [RomanPodymov](https://github.com/RomanPodymov).
 - **UIApplication**:
   - - `queryValue(for:)` extension for URL is refactored. [#668](https://github.com/SwifterSwift/SwifterSwift/pull/668) by [LucianoPAlmeida](https://github.com/ratulSharker).
 - **Sequence**:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,8 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 
 ### Added
 - **Array**:
-    - Added `removeDuplicates(by:)`, `removeDuplicates(keyPath:)`, `withoutDuplicates(by:)`, `withoutDuplicates(keyPath:)` for remove duplicate elements based on comparator or key path. Different implementations for `Element: Equatable` and `Element: Hashable`. [#704](https://github.com/SwifterSwift/SwifterSwift/pull/704) by [RomanPodymov](https://github.com/RomanPodymov).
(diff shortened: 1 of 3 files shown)


Sources/SwifterSwift/SwiftStdlib/ArrayExtensions.swift
```diff
@@ -97,6 +97,56 @@ public extension Array {
...
+    ///
+    ///     [1, -1, 2, -4, 3, 3, 3, 4, -5].withoutDuplicates { abs($0) == abs($1) } -> [1, 2, -4, 3, -5]
+    ///
+    /// - Parameters:
+    ///   - comparator: Returns true if 2 elements of array are equal.
+    /// - Returns: an array of unique elements.
+    ///
+    func withoutDuplicates(by comparator: (Element, Element) -> Bool) -> [Element] {
+        // Thanks to https://github.com/sairamkotha for improving the method
+        return reduce(into: [Element]()) { result, element in
```
LucianoPAlmeida: We could provide a custom more performant version of this for Hashable?

RomanPodymov: Hello @LucianoPAlmeida 
Do you mean [the version](https://github.com/SwifterSwift/SwifterSwift/issues/653#issuecomment-479406641) provided by @guykogus?

RomanPodymov: @LucianoPAlmeida Also, the implementation based on `Set` is changing elements ordering. I am not sure if all programmers expect such behaviour.

LucianoPAlmeida: Kinda... I was thinking in something like:
```swift
func withoutDuplicates(by comparator: (Element, Element) -> Bool) -> [Element] where Element: Hashable {
    var set: Set<Element> = []
    return filter { set.insert($0) } 
}
// And for the mutating
mutating func withoutDuplicates(by comparator: (Element, Element) -> Bool) -> [Element] where Element: Hashable {
    var set: Set<Element> = []
    removeAll(where: { set.insert($0) }) // Since it makes a partition based removal is more optimized.
}
```
I think maybe this is valid because most usages will be with types like Int, String, double ... which are normally hashable. 
What do you think? @guykogus

RomanPodymov: @LucianoPAlmeida But you don't call `comparator` in these functions. Could you please provide a version with this function?

LucianoPAlmeida: Ohh, forget about it I was thinking about the one with no params. Not sure if there is a good way to do it with a custom comparator closure :))

Sources/SwifterSwift/SwiftStdlib/ArrayExtensions.swift
```diff
@@ -154,12 +198,83 @@ public extension Array where Element: Equatable {
     /// - Returns: an array of unique elements.
     ///
     func withoutDuplicates() -> [Element] {
-        // Thanks to https://github.com/sairamkotha for improving the method
```
LucianoPAlmeida: So here that would be applicable as optimization
```swift
func withoutDuplicates() -> [Element] where Element: Hashable {
    var set: Set<Element> = []
    return filter { set.insert($0).inserted }
}
```

Sources/SwifterSwift/SwiftStdlib/ArrayExtensions.swift
```diff
@@ -137,13 +187,7 @@ public extension Array where Element: Equatable {
     /// - Returns: Return array with all duplicate elements removed.
     @discardableResult
     mutating func removeDuplicates() -> [Element] {
-        // Thanks to https://github.com/sairamkotha for improving the method
```
LucianoPAlmeida: Here that would be applicable as optimization:
```swift
mutating func withoutDuplicates() -> [Element] {
      var set: Set<Element> = []
      removeAll(where: { !set.insert($0).inserted }) // Since it makes a partition based removal is more optimized.
      return self
}
```

RomanPodymov: Thank you @LucianoPAlmeida. Added your version to `public extension Array where Element: Hashable { }`.

Sources/SwifterSwift/SwiftStdlib/ArrayExtensions.swift
RomanPodymov: Thank you @LucianoPAlmeida. Added your version to `public extension Array where Element: Hashable { }`.

Sources/SwifterSwift/SwiftStdlib/ArrayExtensions.swift
RomanPodymov: Hello @LucianoPAlmeida 
With `extension Array where Element: Hashable` Codecov reporting that my pull request decreases coverage by some percents. That's why I removed this extension. Should I revert it? Or maybe I should remove `removeDuplicates` and `withoutDuplicates` in `extension Array where Element: Equatable`?

RomanPodymov:
> > The problem is that I don't know how to implement `func withoutDuplicates<E: Equatable>(keyPath path: KeyPath<Element, E>) -> [Element]` with the advantages of `Hashable` and keep the same elements ordering in the return value.
> 
> @RomanPodymov I was thinking something like:
> 
> ```swift
>     func withoutDuplicates<E: Hashable>(keyPath path: KeyPath<Element, E>) -> [Element] {
>       var set: Set<E> = []
>     // since set don't allow duplications it never adds two equal key path values, and filter by that and maintaining the check O(1)
>       return filter { set.insert($0[keyPath: path]).inserted }
>     }
> ```
> 
> I think that could work :))
> 
> Edit: Note that the element itself doesn't need to be Hashable, only the keyPath type.

Thank you. I think we can leave only 1 `withoutDuplicates(keyPath:)` method. Or should we also keep `func withoutDuplicates<E: Equatable>(keyPath path: KeyPath<Element, E>) -> [Element]`?

guykogus:
@LucianoPAlmeida good catch on only checking `<E: Hashable>`, that allows for a nice optimisation without having to specialise. I wonder, though, how the compiler will choose between the `<E: Equatable>` and `<E: Hashable>` options if the type conforms to both. From running a simple test it used the `Hashable` option, but I don't know where/how that's defined.

Sources/SwifterSwift/SwiftStdlib/ArrayExtensions.swift
```diff
@@ -146,4 +146,13 @@ public extension Array where Element: Equatable {
         }
     }
 
+    /// SwifterSwift: Returns a set with all duplicate elements removed using KeyPath to compare.
+    ///
+    /// - Parameter path: Key path to compare, the value must be Hashable.
+    /// - Returns: a set of unique elements.
```
guykogus: Now you're returning an array
```suggestion
    /// - Returns: an array of unique elements.
```

Sources/SwifterSwift/SwiftStdlib/ArrayExtensions.swift
```diff
@@ -146,4 +146,13 @@ public extension Array where Element: Equatable {
         }
     }
 
+    /// SwifterSwift: Returns a set with all duplicate elements removed using KeyPath to compare.
```
guykogus:
```suggestion
    /// SwifterSwift: Returns an array with all duplicate elements removed using KeyPath to compare.
```

Sources/SwifterSwift/SwiftStdlib/ArrayExtensions.swift
```diff
@@ -146,4 +146,13 @@ public extension Array where Element: Equatable {
...
+    ///
+    /// - Parameter path: Key path to compare, the value must be Hashable.
+    /// - Returns: a set of unique elements.
+    func withoutDuplicates<E: Hashable>(keyPath path: KeyPath<Element, E>) -> [Element] {
+        var set: Set<E> = []
+        // since set don't allow duplications it never adds two equal key path values, and filter by that and maintaining the check O(1)
```
guykogus: Remove this comment #cleancode 😉

Sources/SwifterSwift/SwiftStdlib/ArrayExtensions.swift
```diff
@@ -146,4 +146,13 @@ public extension Array where Element: Equatable {
         }
     }
 
+    /// SwifterSwift: Returns a set with all duplicate elements removed using KeyPath to compare.
+    ///
+    /// - Parameter path: Key path to compare, the value must be Hashable.
+    /// - Returns: a set of unique elements.
+    func withoutDuplicates<E: Hashable>(keyPath path: KeyPath<Element, E>) -> [Element] {
+        var set: Set<E> = []
```
guykogus:
```suggestion
        var set = Set<E>()
```

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 ## Upcoming Release
 
 ### Added
+- **Array**:
+    - Added `withoutDuplicates(keyPath:)` for remove duplicate elements based on key path. [#704](https://github.com/SwifterSwift/SwifterSwift/pull/704) by [RomanPodymov](https://github.com/RomanPodymov).
```
guykogus: You copied bad english from the other entry 😂
```suggestion
    - Added `withoutDuplicates(keyPath:)` for filtering duplicate elements based on key path. [#704](https://github.com/SwifterSwift/SwifterSwift/pull/704) by [RomanPodymov](https://github.com/RomanPodymov).
```

Sources/SwifterSwift/SwiftStdlib/ArrayExtensions.swift
RomanPodymov: Done.

guykogus:
@RomanPodymov so we're not including the `Equatable` version too? @LucianoPAlmeida what do you think?

LucianoPAlmeida:
> I wonder, though, how the compiler will choose between the <E: Equatable> and <E: Hashable> options if the type conforms to both. From running a simple test it used the Hashable option, but I don't know where/how that's defined.
> 
@guykogus 
Since Hashable is defined as a specialization of Equatable(`protocol Hashable: Equatable`) the compiler will always attempt the overloads and choose for the most specialized in that case.

guykogus:
> Since Hashable is defined as a specialization of Equatable(`protocol Hashable: Equatable`) the compiler will always attempt the overloads and choose for the most specialized in that case.

@LucianoPAlmeida awesome, in that case let's have both versions. Can you add that @RomanPodymov?

RomanPodymov:
> > Since Hashable is defined as a specialization of Equatable(`protocol Hashable: Equatable`) the compiler will always attempt the overloads and choose for the most specialized in that case.
> 
> @LucianoPAlmeida awesome, in that case let's have both versions. Can you add that @RomanPodymov?

OK, but to test these 2 new methods I need to add a new non-`Hashable` field to `Person`. That's why I created `Location`.

Tests/SwiftStdlibTests/ArrayExtensionsTests.swift
```diff
@@ -8,14 +8,25 @@
 import XCTest
 @testable import SwifterSwift
 
-private struct Person: Equatable {
+private struct Person: Equatable, Hashable {
```
guykogus: Remove `Hashable`, you don't need it and then you won't need to implement the `hash(into:)` function.

Tests/SwiftStdlibTests/ArrayExtensionsTests.swift
```diff
@@ -8,14 +8,25 @@
...
     var name: String
     var age: Int?
+    var location: Location
 
-    static func == (lhs: Person, rhs: Person) -> Bool {
-        return lhs.name == rhs.name && lhs.age == rhs.age
+    init(name: String, age: Int?, location: Location = Location(city: "New York")) {
```
guykogus:
```suggestion
    init(name: String, age: Int?, location: Location = nil) {
```
`location` should be optional. Is that possible?

Tests/SwiftStdlibTests/ArrayExtensionsTests.swift
```diff
@@ -8,14 +8,25 @@
...
     }
 
+    func hash(into hasher: inout Hasher) {
+        hasher.combine(name)
+        hasher.combine(age)
+    }
+}
+
+private struct Location: Equatable {
+    var city: String
```
guykogus:
```suggestion
    let city: String
```

Tests/SwiftStdlibTests/ArrayExtensionsTests.swift
RomanPodymov: Yes.

Tests/SwiftStdlibTests/ArrayExtensionsTests.swift
RomanPodymov: `Location` is not `Hashable`, that's why `func hash(into hasher: inout Hasher)` cannot be generated for `Person`.

guykogus: You don't need `Person` to be `Hashable`, so you can remove it.

RomanPodymov: Got you. Done.

Issue #644, opened 2019-03-10
Add Probot-Stale for Issues and PR


BennX:
It might make sense to add a Stale Bot to the Issues and PR's.

https://github.com/probot/stale
https://github.com/apps/stale

What do you think?

LucianoPAlmeida:
Sure @BennX 💯

LucianoPAlmeida:
Now that I just closed a lot of stale PR's manually, I'm pretty sure it will be a great addition. 
Labeling help wanted in case someone wants to tackle this :))

PR #705, opened 2019-08-17
Add a probot stale configuration

BennX:
See Issue #644 

Also see https://github.com/probot/stale

It's a default configuration with 30 Days until stale (label `won't fix`) and another 7 days until close.  It ignores issues with the label `help wanted`.

Additionally it has the following close comment:

> This issue has been automatically closed because it has not had recent activity. Thank you for your contributions.

When this is merged we have to setup the probot for this repository: https://github.com/apps/stale
(I'll request the add of the app as soon as this is merged. **The Org administrator has to verify this**.)

Since it doesn't add anything to the library itself, I haven't added a changelog entry.

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -12,7 +12,11 @@ onlyLabels: []
 
 # Issues or Pull Requests with these labels will never be considered stale. Set to `[]` to disable
 exemptLabels:
- - help wanted
+  - enhancement
+  - bug
+  - confirmed bug
+  - discussion
+  - help wanted
 
 # Set to true to ignore issues in a project (defaults to false)
 exemptProjects: false
@@ -24,7 +28,7 @@ exemptMilestones: false
 exemptAssignees: false
 
 # Label to use when marking as stale
-staleLabel: won't fix
+staleLabel: stale
 
 # Comment to post when marking as stale. Set to `false` to disable
 markComment: >
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,60 @@
+# Configuration for probot-stale - https://github.com/probot/stale
+
+# Number of days of inactivity before an Issue or Pull Request becomes stale
+daysUntilStale: 30
+
+# Number of days of inactivity before an Issue or Pull Request with the stale label is closed.
+# Set to false to disable. If disabled, issues still need to be closed manually, but will remain marked as stale.
+daysUntilClose: 7
+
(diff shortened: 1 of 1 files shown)


.github/stale.yml
```diff
@@ -0,0 +1,60 @@
...
+exemptProjects: false
+
+# Set to true to ignore issues in a milestone (defaults to false)
+exemptMilestones: false
+
+# Set to true to ignore issues with an assignee (defaults to false)
+exemptAssignees: false
+
+# Label to use when marking as stale
+staleLabel: won't fix
```
LucianoPAlmeida:
```suggestion
staleLabel: stale
```

.github/stale.yml
```diff
@@ -0,0 +1,60 @@
...
+closeComment: >
+  This issue has been automatically closed because it has not had
+  recent activity. Thank you for your contributions.
+
+# Limit the number of actions per hour, from 1-30. Default is 30
+limitPerRun: 30
+
+# Limit to only `issues` or `pulls`
+# only: issues
+
```
LucianoPAlmeida: Should add exclusion labels?
```
# Issues or Pull Requests with these labels will never be considered stale
exemptLabels:
  - "enhancement"
  - "bug?"
  - "confirmed bug"
  - "discussion"
```

BennX: sure 👍

.github/stale.yml
```diff
@@ -0,0 +1,64 @@
...
+# Set to false to disable. If disabled, issues still need to be closed manually, but will remain marked as stale.
+daysUntilClose: 7
+
+# Only issues or pull requests with all of these labels are check if stale. Defaults to `[]` (disabled)
+onlyLabels: []
+
+# Issues or Pull Requests with these labels will never be considered stale. Set to `[]` to disable
+exemptLabels:
+  - enhancement
+  - bug
```
LucianoPAlmeida: Is there a way to add a pattern here e.g. everything that contains 'bug'. We have many labels for bugs like 'possible bug', 'confirmed bug: source' ...

BennX: Not that I know of 😞  But I think it'll do for now 😁

BennX:
@LucianoPAlmeida Just in case, I'm sure letting the bot match on regex would be a straight forward Pull Request 😉 

Feel free to merge it. I'm sure we'll need @SD10 to confirm the bot adding in the end.

PR #706, opened 2019-08-24
♻️ Moved divided(by:) function from ArrayExtensions to SequenceExtensions

cHaLkdusT:
Categorize all items into two lists based on the where condition. 

`let (passed, failed) = [88, 65, 75, 99, 80].divided { $0 > 75 }`
`let (boys, girls) = students.divided { $0.gender == .boy }`

## Checklist

### DIFF
--- a/Sources/SwifterSwift/SwiftStdlib/SequenceExtensions.swift
+++ b/Sources/SwifterSwift/SwiftStdlib/SequenceExtensions.swift
@@ -180,7 +180,7 @@ public extension Sequence {
         //Inspired by: http://ruby-doc.org/core-2.5.0/Enumerable.html#method-i-partition
         var matching = ContiguousArray<Element>()
         var nonMatching = ContiguousArray<Element>()
-        
+
         var iterator = self.makeIterator()
         while let element = iterator.next() {
             try condition(element) ? matching.append(element) : nonMatching.append(element)
--- a/Sources/SwifterSwift/SwiftStdlib/SequenceExtensions.swift
+++ b/Sources/SwifterSwift/SwiftStdlib/SequenceExtensions.swift
@@ -255,7 +255,7 @@ public extension Sequence where Element: Numeric {
     ///
     /// - Returns: sum of the array's elements.
     func sum() -> Element {
-        return reduce(0, {$0 + $1})
+        return reduce(into: 0, +=)
     }
 
 }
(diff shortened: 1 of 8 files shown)


LucianoPAlmeida:
Hi @cHaLkdusT, thank's for extension. But we already have this implemented. [Here](https://github.com/SwifterSwift/SwifterSwift/blob/5c35781029ac1230bedd8a769ac5396884ae9731/Sources/SwifterSwift/SwiftStdlib/ArrayExtensions.swift#L43).
But it is actually on array extensions and it can for sure be moved to Sequence. So if you want to change this PR to update this extension instead, it will be great :))

cHaLkdusT:
Sorry! I miss that! 😅

cHaLkdusT:
Heya @LucianoPAlmeida, I have now moved `divided(by:)` from ArrayExtensions to SequenceExtensions. 😆

Sources/SwifterSwift/SwiftStdlib/SequenceExtensions.swift
```diff
@@ -167,6 +167,27 @@ public extension Sequence {
...
+    func divided(by condition: (Element) throws -> Bool) rethrows -> (matching: [Element], nonMatching: [Element]) {
+        //Inspired by: http://ruby-doc.org/core-2.5.0/Enumerable.html#method-i-partition
+        var matching = ContiguousArray<Element>()
+        var nonMatching = ContiguousArray<Element>()
+        
+        var iterator = self.makeIterator()
+        while let element = iterator.next() {
+            try condition(element) ? matching.append(element) : nonMatching.append(element)
+        }
+        return (Array(matching), Array(nonMatching))
```
LucianoPAlmeida: Instead of using a contiguous array and then converting here, we could just use an array :))

LucianoPAlmeida:
The linux build is failing. Can you run `swift test --generate-linuxmain`?

Sources/SwifterSwift/SwiftStdlib/SequenceExtensions.swift
cHaLkdusT: I just thought of using ContiguousArray to be more efficient and have more predictable performance than Array. I've just mimicked how Swift implements it. https://github.com/apple/swift/blob/master/stdlib/public/core/Sequence.swift#L646 Haha!

cHaLkdusT: Shall we use `[Element]` instead? :D

LucianoPAlmeida: Nice, I just thought that wouldn't be worth to pay the price for the creation of a new array. But makes sense to use Contiguous array for performance :)) So let's keep it that way

cHaLkdusT: Coolio 👍

PR #711, opened 2019-09-16
Mutating versions of removeDuplicates(keyPath:) and removeDuplicates() refactoring

### LINKED ISSUE #710: Mutating versions of removeDuplicates(keyPath:)
- [x] 🚀 I've read, understood, and done my best to follow the [Contributing guidelines](https://github.com/SwifterSwift/SwifterSwift/tree/master/CONTRIBUTING.md) before opening this issue.

Currently there are `nonmutating` `withoutDuplicates()` and `mutating` `removeDuplicates()` in `extension Array`. Let's add `mutating` versions of `func withoutDuplicates<E: Equatable>(keyPath path: KeyPath<Element, E>) -> [Element]` and `func withoutDuplicates<E: Hashable>(keyPath path: KeyPath<Element, E>) -> [Element]`.

### SwifterSwift Environment

- SwifterSwift version: 5.0.0
- Xcode version: 10.3
- macOS version running Xcode: 10.14.6
- Swift version: 5.0
- Platform(s) running SwifterSwift: macOS

RomanPodymov:
🚀 Fixed #710. Refactored `Array.removeDuplicates()`.

## Checklist

### DIFF
--- a/Sources/SwifterSwift/SwiftStdlib/ArrayExtensions.swift
+++ b/Sources/SwifterSwift/SwiftStdlib/ArrayExtensions.swift
@@ -122,7 +122,11 @@ public extension Array where Element: Equatable {
     @discardableResult
     mutating func removeDuplicates() -> [Element] {
         // Thanks to https://github.com/sairamkotha for improving the method
-        self = withoutDuplicates()
+        self = reduce(into: [Element]()) {
+            if !$0.contains($1) {
+                $0.append($1)
+            }
+        }
         return self
     }
 
--- a/Sources/SwifterSwift/SwiftStdlib/ArrayExtensions.swift
+++ b/Sources/SwifterSwift/SwiftStdlib/ArrayExtensions.swift
@@ -122,11 +122,7 @@ public extension Array where Element: Equatable {
     @discardableResult
     mutating func removeDuplicates() -> [Element] {
         // Thanks to https://github.com/sairamkotha for improving the method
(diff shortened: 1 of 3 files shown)


Sources/SwifterSwift/SwiftStdlib/ArrayExtensions.swift
```diff
@@ -122,11 +122,7 @@ public extension Array where Element: Equatable {
     @discardableResult
     mutating func removeDuplicates() -> [Element] {
         // Thanks to https://github.com/sairamkotha for improving the method
-        self = reduce(into: [Element]()) {
-            if !$0.contains($1) {
-                $0.append($1)
-            }
-        }
+        self = withoutDuplicates()
```
LucianoPAlmeida: I know that this makes sense for the code reuse point of view and there are other extensions on the project that does that, but since we are always learning and trying to improve, making a copy the returning array to `self` maybe is not the best option. Could we implement a more specialized version of this that doesn't make copy the array? Maybe some using `removeAll(where: { check for duplication })` I think this could work.
Same apply to the other method in this PR :))

RomanPodymov: Hello @LucianoPAlmeida 
I think we should leave `func removeDuplicates() -> [Element]` when `Element` is `Equatable` as it is. On the other hand I think we should add a version for `Element: Hashable`.

PR #712, opened 2019-09-21
Sorted key path improvements

LucianoPAlmeida:
## Checklist


Sources/SwifterSwift/SwiftStdlib/MutableCollectionExtensions.swift
```diff
@@ -0,0 +1,17 @@
+//
+//  MutableCollectionExtensions.swift
+//  SwifterSwift
+//
+//  Created by Luciano Almeida on 21/09/19.
+//  Copyright © 2019 SwifterSwift
+//
+
+public extension MutableCollection where Self: RandomAccessCollection {
+  /// SwifterSwift: Sort the collection  based on a keypath and a compare function.
```
guykogus: Only 2 spaces for tabs here, unlike the rest of the project

LucianoPAlmeida:
> You've deprecated functions that allow you to just call `sort(by: someKeyPath)` but the replacements don't allow for the same - you _must_ supply a comparison function.
> ```

Maybe we can, but the problem in allowing call it like that, is since we only deprecate the old one calling like that `sort(by: someKeyPath)`, the compiler would have two overloads to choose from the old `sort(by:ascending)` and `sort(by:with:)` which can lead to ambiguous error or it chooses the deprecated one.

guykogus:
> Maybe we can, but the problem in allowing call it like that, is since we only deprecate the old one calling like that `sort(by: someKeyPath)`, the compiler would have two overloads to choose from the old `sort(by:ascending)` and `sort(by:with:)` which can lead to ambiguous error or it chooses the deprecated one.

If there are overloads then delete the old ones that we don't need anymore, no?

LucianoPAlmeida:
> If there are overloads then delete the old ones that we don't need anymore, no?
>

Maybe, since it would not be a breaking change :)

LucianoPAlmeida:
> You've removed the ascending parameter, which means that you're introducing a breaking change.
>

I just removed the default value so that method can't be called like `sort(by: )` anymore, which is not a breaking change because those will be now calling the `sort(by:)` with the default comparator overload.

PR #717, opened 2019-10-02
Added element(:) extension to array

zmfrew:
:rocket: I added an extension on `Array`, which allows for safe accessing of elements in the array. It provides a clean, swifty way of accessing elements in the array without subscript syntax.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 ## Upcoming Release
 
 ### Added
+- **Array**:
+    - Added `element(at:)` for safely accessing elements without using subscript syntax. [#717](https://github.com/SwifterSwift/SwifterSwift/pull/717) by [Zach Frew](https://github.com/zmfrew).
+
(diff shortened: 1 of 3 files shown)


LucianoPAlmeida:
I think we have that already as `array[safe: 1]` :))

zmfrew:
@LucianoPAlmeida I saw that the `subscript(safe index: Index)` exists, and I see dot syntax as more _Swifty_ compared to subscript syntax.

LucianoPAlmeida:
Yeah, but it is so simple and provide the same functionality that I'm not sure if is worth adding both versions. I'd prefer to maintain only the one we have already. 
cc @guykogus

zmfrew:
Understandable as it's a stylistic choice rather than providing new functionality. Do you want me to go ahead and close out the PR?

LucianoPAlmeida:
> Understandable as it's a stylistic choice rather than providing new functionality. Do you want me to go ahead and close out the PR?

That's only my point of view, let's wait for the other guys opinion before close it :)) 
cc @guykogus @omaralbeik

guykogus:
I'm with @LucianoPAlmeida, we already have this functionality, and I don't see why using functions instead of subscripting is more "Swifty".

PR #718, opened 2019-10-02
Included isNotEmpty computed var on Collection

zmfrew:
🚀 This extension adds clarity and avoids double negatives when checking if a collection is not empty.

## Checklist

### DIFF
--- a/Sources/SwifterSwift/SwiftStdlib/CollectionExtensions.swift
+++ b/Sources/SwifterSwift/SwiftStdlib/CollectionExtensions.swift
@@ -10,18 +10,20 @@
 import Dispatch
 #endif
 
-// MARK: - Methods
+// MARK: - Properties
 public extension Collection {
-    
+
     /// SwifterSwift: Check if the given collection is not empty.
     ///
     ///     [1, 2, 3, 4, 5].isNotEmpty -> True
     ///     [].isNotEmpty -> False
     ///
     /// - Returns: true if the collection is not empty and false if the collection is empty
-    var isNotEmpty: Bool {
-        return !self.isEmpty
-    }
+    var isNotEmpty: Bool { !self.isEmpty }
+}
+
+// MARK: - Methods
+public extension Collection {
 
     #if canImport(Dispatch)
     /// SwifterSwift: Performs `each` closure for each element of collection in parallel.
--- a/Sources/SwifterSwift/SwiftStdlib/CollectionExtensions.swift
+++ b/Sources/SwifterSwift/SwiftStdlib/CollectionExtensions.swift
@@ -12,11 +12,17 @@ import Dispatch
 
 // MARK: - Methods
 public extension Collection {
-
+    
+    /// SwifterSwift: Check if the given collection is not empty.
+    ///
+    ///     [1, 2, 3, 4, 5].isNotEmpty -> True
+    ///     [].isNotEmpty -> False
+    ///
+    /// - Returns: true if the collection is not empty and false if the collection is empty
     var isNotEmpty: Bool {
         return !self.isEmpty
     }
-    
+
     #if canImport(Dispatch)
     /// SwifterSwift: Performs `each` closure for each element of collection in parallel.
     ///
--- a/Sources/SwifterSwift/SwiftStdlib/CollectionExtensions.swift
+++ b/Sources/SwifterSwift/SwiftStdlib/CollectionExtensions.swift
@@ -13,6 +13,10 @@ import Dispatch
 // MARK: - Methods
 public extension Collection {
 
+    var isNotEmpty: Bool {
+        return !self.isEmpty
+    }
+    
     #if canImport(Dispatch)
     /// SwifterSwift: Performs `each` closure for each element of collection in parallel.
     ///
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 ## Upcoming Release
 
 ### Added
+- **Collection**:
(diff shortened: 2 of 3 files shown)


LucianoPAlmeida:
I'm really not sure about that one,  to me is abstracting something that is too simple to add significant value. But seems that is an argument over readability... really don't know. I'm -1 for this but let the decision to you guys @guykogus @omaralbeik

omaralbeik:
Although this is a trivial extension, I think it adds more clarity and increases readability
especially when the line is long

```swift
guard !someClass.someProperty.anotherStruct.something.isEmpty else { return }
// vs.
guard someClass.someProperty.anotherStruct.something.isNotEmpty else { return }
```

Sources/SwifterSwift/SwiftStdlib/CollectionExtensions.swift
```diff
@@ -12,6 +12,16 @@ import Dispatch
...
 public extension Collection {
+    
+    /// SwifterSwift: Check if the given collection is not empty.
+    ///
+    ///     [1, 2, 3, 4, 5].isNotEmpty -> True
+    ///     [].isNotEmpty -> False
+    ///
+    /// - Returns: true if the collection is not empty and false if the collection is empty
+    var isNotEmpty: Bool {
+        return !self.isEmpty
```
omaralbeik: maybe refactor this to a more Swifty style 😉

```swift
var isNotEmpty: Bool { !isEmpty }
```

guykogus:
I'm also not a fan of this kind of abstraction. You could take this reasoning to extend to virtually every single boolean flag that exists.
I also think that this library should be more about extending functionality, rather than improving readability. I give it a -1 too.

LucianoPAlmeida:
> Although this is a trivial extension, I think it adds more clarity and increases readability
> especially when the line is long
> 
> ```swift
> guard !someClass.someProperty.anotherStruct.something.isEmpty else { return }
> // vs.
> guard someClass.someProperty.anotherStruct.something.isNotEmpty else { return }
> ```

@omaralbeik I've seen people disscussion about this with the argument of readability, but a downside would have to add this to all types that have isEmpty, because if I'm a user and in array I have isEmpty/isNotEmpty, I would expect it also in String, Set ...
So I'm not sure if it's worthed

guykogus:
@LucianoPAlmeida this extension is for the `Collection` protocol, so all those other classes would inherit the implementation automatically.

Sources/SwifterSwift/SwiftStdlib/CollectionExtensions.swift
zmfrew: Great suggestion! Updated in my most recent commit.

zmfrew:
I made a commit moving it into a separate extension for `Properties` rather than grouping it with the `Methods` to match the other files' structure. @omaralbeik example up above is exactly how I use it.

LucianoPAlmeida:
> @LucianoPAlmeida this extension is for the `Collection` protocol, so all those other classes would inherit the implementation automatically.

Yeah, was just an example, I meant that not all types that have `isEmpty` are collection, was thinking in other types like Set ... :))  The point is we would have to maintain consistence

guykogus:
@LucianoPAlmeida `Set` also conforms to `Collection` 😛 Regardless, I'm on your side and vote against this PR. Adding `isNot...` variations for properties opens the door to _every_ boolean property having the same, and it seems unnecessary.

LucianoPAlmeida:
Set is `Collection`?? That's good to know haha thought it was just `Sequence`. So we can access set by index? kinda weird since theres is no ordering garantee. But anyways, thank for the info :))

guykogus:
@LucianoPAlmeida the `Index` of a `Collection` doesn't necessarily have to be `Int`, so `Set` can easily still conform. The `Collection` documentation explains that if you want to conform to it you just need `startIndex`, `endIndex`, `index(after:)` and `subscript`. That gives the basic building block to any collection.

The really fun part is the definition of `Index`:
```Swift
associatedtype Index : Comparable where Self.Index == Self.Indices.Element, Self.Indices.Element == Self.Indices.Index, Self.Indices.Index == Self.SubSequence.Index, Self.SubSequence.Index == Self.Indices.Indices.Element, Self.Indices.Indices.Element == Self.Indices.Indices.Index, Self.Indices.Indices.Index == Self.SubSequence.Indices.Element, Self.SubSequence.Indices.Element == Self.SubSequence.Indices.Index, Self.SubSequence.Indices.Index == Self.Indices.Indices.Indices.Element, Self.Indices.Indices.Indices.Element == Self.Indices.Indices.Indices.Index, Self.Indices.Indices.Indices.Index == Self.SubSequence.Indices.Indices.Element, Self.SubSequence.Indices.Indices.Element == Self.SubSequence.Indices.Indices.Index, Self.SubSequence.Indices.Indices.Index == Self.SubSequence.Indices.Indices.Indices.Element, Self.SubSequence.Indices.Indices.Indices.Element == Self.SubSequence.Indices.Indices.Indices.Index
```
😳

omaralbeik:
I agree with @guykogus on this one, we don't want every bool property to have a `isNot...` extensions 😆, let's get more feedback on this PR

@SwifterSwift/contributors please vote with 👍 or 👎

Sources/SwifterSwift/SwiftStdlib/CollectionExtensions.swift
towry:
```swift
let someCollectionIsNotEmpty = !someClass.someProperty.anotherStruct.something.isEmpty;
guard someCollectionIsNotEmpty else { return }
// vs.
guard someClass.someProperty.anotherStruct.something.isNotEmpty else { return }
```

guykogus:
If what you're really looking to avoid is:
```Swift
guard !someClass.someProperty.anotherStruct.something.isEmpty else { return }
```
then you can always do
```Swift
guard someClass.someProperty.anotherStruct.something.isEmpty == false else { return }
```

It looks like the consensus is overwhelmingly against this PR, so I'm going to close it.

PR #719, opened 2019-10-02
Added replacingOccurrences extension to String

zmfrew:
🚀 This extension comes from Paul Hudson at Hacking with Swift. It allows for replacing a string within a string with specific text a specified amount of times.

https://www.hackingwithswift.com/articles/141/8-useful-swift-extensions

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 ## Upcoming Release
 
 ### Added
+- **String**:
+    - Added `replacingOccurrences(of search:, with replacement:, count maxReplacements:)`, which allows for replacing a string within a string with specific text a specified amount of times. [#719](https://github.com/SwifterSwift/SwifterSwift/pull/719) by [Zach Frew](https://github.com/zmfrew).
 
 ### Changed
 
--- a/Sources/SwifterSwift/SwiftStdlib/StringExtensions.swift
+++ b/Sources/SwifterSwift/SwiftStdlib/StringExtensions.swift
@@ -604,7 +604,7 @@ public extension String {
 
         while let range = returnValue.range(of: search), count < maxReplacements {
             returnValue.replaceSubrange(range, with: replacement)
-             count += 1
+            count += 1
         }
 
         return returnValue
--- a/Sources/SwifterSwift/SwiftStdlib/StringExtensions.swift
+++ b/Sources/SwifterSwift/SwiftStdlib/StringExtensions.swift
@@ -602,14 +602,9 @@ public extension String {
         var count = 0
         var returnValue = self
 
-        while let range = returnValue.range(of: search) {
-            returnValue = returnValue.replacingCharacters(in: range, with: replacement)
-            count += 1
-
-            // exit as soon as we've made all replacements
-            if count == maxReplacements {
-                return returnValue
-            }
+        while let range = returnValue.range(of: search), count < maxReplacements {
+             returnValue = returnValue.replacingCharacters(in: range, with: replacement)
+             count += 1
         }
 
         return returnValue
--- a/Sources/SwifterSwift/SwiftStdlib/StringExtensions.swift
+++ b/Sources/SwifterSwift/SwiftStdlib/StringExtensions.swift
@@ -586,6 +586,36 @@ public extension String {
     }
     #endif
 
+    #if canImport(Foundation)
+    // https://www.hackingwithswift.com/articles/141/8-useful-swift-extensions
+    /// SwifterSwift: Replaces a set number of occurrences of a string with replacement text.
+    ///
+    ///        "I swift swift".replacingOccurrences(of: "swift", with: "love", count: 1) -> "I love swift."
+    /// - Parameters:
+    ///     - search: the string to be replaced
+    ///     - replacement: the replacement string
+    ///     - maxReplacements: the maximum number of times to replace the string, which must be greater than 0
+    /// - Returns: The new string with the replacement text swapped the specified amount of times.
+    func replacingOccurrences(of search: String, with replacement: String, count maxReplacements: Int) -> String {
+        guard maxReplacements > 0 else { return self }
+
+        var count = 0
+        var returnValue = self
+
+        while let range = returnValue.range(of: search) {
+            returnValue = returnValue.replacingCharacters(in: range, with: replacement)
+            count += 1
+
+            // exit as soon as we've made all replacements
+            if count == maxReplacements {
+                return returnValue
+            }
+        }
+
+        return returnValue
+    }
+    #endif
+
     /// SwifterSwift: Safely subscript string with index.
     ///
     ///		"Hello World!"[safe: 3] -> "l"
--- a/Sources/SwifterSwift/SwiftStdlib/StringExtensions.swift
+++ b/Sources/SwifterSwift/SwiftStdlib/StringExtensions.swift
@@ -603,7 +603,7 @@ public extension String {
         var returnValue = self
 
         while let range = returnValue.range(of: search), count < maxReplacements {
-             returnValue = returnValue.replacingCharacters(in: range, with: replacement)
+            returnValue.replaceSubrange(range, with: replacement)
              count += 1
         }
 
--- a/Tests/SwiftStdlibTests/StringExtensionsTests.swift
+++ b/Tests/SwiftStdlibTests/StringExtensionsTests.swift
@@ -336,6 +336,7 @@ final class StringExtensionsTests: XCTestCase {
     }
 
     func testReplacingOccurrences() {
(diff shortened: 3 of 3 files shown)


Sources/SwifterSwift/SwiftStdlib/StringExtensions.swift
```diff
@@ -586,6 +586,36 @@ public extension String {
...
+    ///     - replacement: the replacement string
+    ///     - maxReplacements: the maximum number of times to replace the string, which must be greater than 0
+    /// - Returns: The new string with the replacement text swapped the specified amount of times.
+    func replacingOccurrences(of search: String, with replacement: String, count maxReplacements: Int) -> String {
+        guard maxReplacements > 0 else { return self }
+
+        var count = 0
+        var returnValue = self
+
+        while let range = returnValue.range(of: search) {
```
LucianoPAlmeida: Maybe 
```swift 
        while let range = returnValue.range(of: search), count < maxReplacements {
            returnValue = returnValue.replacingCharacters(in: range, with: replacement)
            count += 1
        }
```
What you think?

CHANGELOG.md
```diff
@@ -5,6 +5,10 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 ## Upcoming Release
 
 ### Added
+- **String**:
+    - Added `replacingOccurrences(of search:, with replacement:, count maxReplacements:)`, which allows for replacing a string within a string with specific text a specified amount of times. [#719](https://github.com/SwifterSwift/SwifterSwift/pull/719) by [Zach Frew](https://github.com/zmfrew).
+
+Added `replacingOccurrences(of search:, with replacement:, count maxReplacements:)`, which allows for replacing a string within a string with specific text a specified amount of times. #719 by Zach Frew.
```
LucianoPAlmeida: Duplicated addition.

Sources/SwifterSwift/SwiftStdlib/StringExtensions.swift
```diff
@@ -586,6 +586,36 @@ public extension String {
...
+            returnValue = returnValue.replacingCharacters(in: range, with: replacement)
```
LucianoPAlmeida: Also, we are copying the string multiple types here, maybe there is a in-place mutating version of `replacingCharacters` ?

CHANGELOG.md
zmfrew: Removed in most recent commit.

Sources/SwifterSwift/SwiftStdlib/StringExtensions.swift
zmfrew: I think your suggestion looks cleaner, and the tests pass with the update. I'll update in the next commit.

Sources/SwifterSwift/SwiftStdlib/StringExtensions.swift
zmfrew: I could not find an in-place mutating version in the standard library or Foundation. Do you have any other suggestions?

LucianoPAlmeida: What about [replaceSubrange](https://developer.apple.com/documentation/swift/string/1641462-replacesubrange)?

Sources/SwifterSwift/SwiftStdlib/StringExtensions.swift
```diff
@@ -586,6 +586,31 @@ public extension String {
...
+    /// - Returns: The new string with the replacement text swapped the specified amount of times.
+    func replacingOccurrences(of search: String, with replacement: String, count maxReplacements: Int) -> String {
+        guard maxReplacements > 0 else { return self }
+
+        var count = 0
+        var returnValue = self
+
+        while let range = returnValue.range(of: search), count < maxReplacements {
+            returnValue.replaceSubrange(range, with: replacement)
+             count += 1
```
LucianoPAlmeida:
```suggestion
            count += 1
```

zmfrew: Good catch, updated!

Sources/SwifterSwift/SwiftStdlib/StringExtensions.swift
```diff
@@ -586,6 +586,31 @@ public extension String {
     }
     #endif
 
+    #if canImport(Foundation)
+    // https://www.hackingwithswift.com/articles/141/8-useful-swift-extensions
```
omaralbeik: @zmfrew could you please move the source comment to be inside the function

guykogus: I think it should be in `See also:` in the method documentation

Sources/SwifterSwift/SwiftStdlib/StringExtensions.swift
```diff
@@ -586,6 +586,31 @@ public extension String {
...
+    ///     - replacement: the replacement string
+    ///     - maxReplacements: the maximum number of times to replace the string, which must be greater than 0
+    /// - Returns: The new string with the replacement text swapped the specified amount of times.
+    func replacingOccurrences(of search: String, with replacement: String, count maxReplacements: Int) -> String {
+        guard maxReplacements > 0 else { return self }
+
+        var count = 0
+        var returnValue = self
+
+        while let range = returnValue.range(of: search), count < maxReplacements {
```
guykogus: Reverse these to avoid performing an extra search on the string at the end of the count. It's a tiny optimisation but why not? 😄 
```suggestion
        while count < maxReplacements,
           let range = returnValue.range(of: search) {
```

Sources/SwifterSwift/SwiftStdlib/StringExtensions.swift
```diff
@@ -586,6 +586,31 @@ public extension String {
...
+    /// SwifterSwift: Replaces a set number of occurrences of a string with replacement text.
+    ///
+    ///        "I swift swift".replacingOccurrences(of: "swift", with: "love", count: 1) -> "I love swift."
+    /// - Parameters:
+    ///     - search: the string to be replaced
+    ///     - replacement: the replacement string
+    ///     - maxReplacements: the maximum number of times to replace the string, which must be greater than 0
+    /// - Returns: The new string with the replacement text swapped the specified amount of times.
+    func replacingOccurrences(of search: String, with replacement: String, count maxReplacements: Int) -> String {
+        guard maxReplacements > 0 else { return self }
```
guykogus: You don't need this `guard`, the function will perform almost exactly the same without it.

Sources/SwifterSwift/SwiftStdlib/StringExtensions.swift
```diff
@@ -586,6 +586,31 @@ public extension String {
     }
     #endif
 
+    #if canImport(Foundation)
```
guykogus: Why do you need this `canImport`?

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 ## Upcoming Release
 
 ### Added
+- **String**:
+    - Added `replacingOccurrences(of search:, with replacement:, count maxReplacements:)`, which allows for replacing a string within a string with specific text a specified amount of times. [#719](https://github.com/SwifterSwift/SwifterSwift/pull/719) by [Zach Frew](https://github.com/zmfrew).
```
```suggestion
    - Added `replacingOccurrences(of:with:count:)`, which allows for replacing a string within a string with specific text a specified amount of times. [#719](https://github.com/SwifterSwift/SwifterSwift/pull/719) by [Zach Frew](https://github.com/zmfrew).
```

Sources/SwifterSwift/SwiftStdlib/StringExtensions.swift
```diff
@@ -586,6 +586,31 @@ public extension String {
...
+    // https://www.hackingwithswift.com/articles/141/8-useful-swift-extensions
+    /// SwifterSwift: Replaces a set number of occurrences of a string with replacement text.
+    ///
+    ///        "I swift swift".replacingOccurrences(of: "swift", with: "love", count: 1) -> "I love swift."
+    /// - Parameters:
+    ///     - search: the string to be replaced
+    ///     - replacement: the replacement string
+    ///     - maxReplacements: the maximum number of times to replace the string, which must be greater than 0
+    /// - Returns: The new string with the replacement text swapped the specified amount of times.
+    func replacingOccurrences(of search: String, with replacement: String, count maxReplacements: Int) -> String {
```
guykogus: This is just a suggestion, but I think you can get rid of `count`.
```suggestion
    func replacingOccurrences(of search: String, with replacement: String, maxReplacements: Int) -> String {
```

chulbert: For maximum flexibility I would recommend adding a `String.CompareOptions` parameter and passing that along to the underlying implementation.

That said, this seems like an _awfully_ contrived use case.

PR #721, opened 2019-10-06
Add withBackgroundColor(_:) extensions to UIImage

MaxHaertwig:
This extension allows to specify a background color for partially transparent UIImages.

I also changed my old username in the changelog to my new one.

## Checklist

### DIFF
--- a/Sources/SwifterSwift/UIKit/UIImageExtensions.swift
+++ b/Sources/SwifterSwift/UIKit/UIImageExtensions.swift
@@ -201,6 +201,28 @@ public extension UIImage {
         return UIGraphicsGetImageFromCurrentImageContext()!
     }
 
+    /// SwifterSwift: UImage with background color
+    ///
+    /// - Parameters:
+    ///   - backgroundColor: Color to use as background color
+    /// - Returns: UIImage with a background color that is visible where alpha < 1
+    func withBackgroundColor(_ backgroundColor: UIColor) -> UIImage? {
+        let rect = CGRect(origin: .zero, size: size)
+        UIGraphicsBeginImageContextWithOptions(rect.size, false, scale)
+        guard let context = UIGraphicsGetCurrentContext() else {return nil}
+
+        backgroundColor.setFill()
+        UIRectFill(rect)
+
+        context.translateBy(x: 0, y: size.height)
+        context.scaleBy(x: 1, y: -1)
+        context.draw(cgImage!, in: rect)
+
+        defer { UIGraphicsEndImageContext() }
+
+        return UIGraphicsGetImageFromCurrentImageContext()
+    }
+
     /// SwifterSwift: UIImage with rounded corners
     ///
     /// - Parameters:
--- a/Sources/SwifterSwift/UIKit/UIImageExtensions.swift
+++ b/Sources/SwifterSwift/UIKit/UIImageExtensions.swift
@@ -207,7 +207,7 @@ public extension UIImage {
     ///   - backgroundColor: Color to use as background color
     /// - Returns: UIImage with a background color that is visible where alpha < 1
     func withBackgroundColor(_ backgroundColor: UIColor) -> UIImage {
-        if #available(iOS 10.0, *) {
+        if #available(iOS 10.0, tvOS 10.0, *) {
             let format = UIGraphicsImageRendererFormat()
             format.scale = scale
             return UIGraphicsImageRenderer(size: size, format: format).image { context in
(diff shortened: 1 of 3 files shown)


Sources/SwifterSwift/UIKit/UIImageExtensions.swift
```diff
@@ -201,6 +201,28 @@ public extension UIImage {
...
     }
 
+    /// SwifterSwift: UImage with background color
+    ///
+    /// - Parameters:
+    ///   - backgroundColor: Color to use as background color
+    /// - Returns: UIImage with a background color that is visible where alpha < 1
+    func withBackgroundColor(_ backgroundColor: UIColor) -> UIImage? {
+        let rect = CGRect(origin: .zero, size: size)
+        UIGraphicsBeginImageContextWithOptions(rect.size, false, scale)
```
LucianoPAlmeida: This is an old API, and although is not deprecated I think is not recommended to use it anymore. I know all the other extensions use that (we maybe change those in the future) is preferable for new extensions to use `UIGraphicsImageRenderer` API instead because unlike `UIGraphicsBeginImageContextWithOptions`, `UIGraphicsImageRenderer` is capable of rendering wide-color content correctly.

MaxHaertwig: I agree. However, UIGraphicsImageRenderer is only available in iOS 10 and later. I could provide a fallback, but the result would still be `UIImage?` instead of `UIImage` which would be possible with UIGraphicsImageRenderer.
What are your thoughts about this?

LucianoPAlmeida: I'm fine with providing this API only for iOS 10+, but feel free to provide a fallback if you think it adds value. I personally think that fewer people are supporting less than iOS 10 these days so it wouldn't affect too much :))

Tests/UIKitTests/UIImageExtensionsTests.swift
```diff
@@ -130,6 +130,15 @@ final class UIImageExtensionsTests: XCTestCase {
         XCTAssertEqual(testImage.bytesSize, tintedImage.bytesSize)
     }
 
+    func testWithBackgroundColor() {
+        let size = CGSize(width: 1, height: 1)
+        let clearImage = UIImage(color: .clear, size: size)
+        let imageWithBackgroundColor = clearImage.withBackgroundColor(.black)
+        XCTAssertNotNil(imageWithBackgroundColor)
+        let blackImage = UIImage(color: .black, size: size)
+        XCTAssertEqual(imageWithBackgroundColor!.bytesSize, blackImage.bytesSize)
```
guykogus: Perhaps instead you should be comparing the PNG or JPEG data representation of the images? Just comparing the number of bytes doesn't seem accurate.

Sources/SwifterSwift/UIKit/UIImageExtensions.swift
```diff
@@ -201,6 +201,28 @@ public extension UIImage {
...
+        guard let context = UIGraphicsGetCurrentContext() else {return nil}
+
+        backgroundColor.setFill()
+        UIRectFill(rect)
+
+        context.translateBy(x: 0, y: size.height)
+        context.scaleBy(x: 1, y: -1)
+        context.draw(cgImage!, in: rect)
+
+        defer { UIGraphicsEndImageContext() }
```
guykogus: It's just a style comment, but I think the `defer` should come straight after creating the context. If you're going to do it down here at the end, where a `defer` isn't actually necessary, I think it would be cleaner to store the image in a temp local, end the context and then return the image.

Sources/SwifterSwift/UIKit/UIImageExtensions.swift
guykogus: I'm all for using `UIGraphicsImageRenderer` with a backwards-compatible fallback, I think it would be the right way to go about this.

Sources/SwifterSwift/UIKit/UIImageExtensions.swift
MaxHaertwig: Makes sense, thanks!

Tests/UIKitTests/UIImageExtensionsTests.swift
MaxHaertwig: Done

guykogus:
@MaxHaertwig the Linux tests failed. Can you please fix it? @LucianoPAlmeida any suggestions on how to do that?

MaxHaertwig:
I fixed the Linux tests.

MaxHaertwig:
I guess we can't test it with two different versions of iOS and combine the results.

PR #722, opened 2019-10-06
Add init(light:dark:) to NSColor and UIColor

MaxHaertwig:
This extension allows to create an NSColor or UIColor with different colors for light mode and dark mode. I created separate `NSColorExtensions.swift` and `UIColorExtensions.swift` files, because NSColor and UIColor use different APIs for this problem and can't be unified.

I couldn't figure out how to create a test for macOS, because it uses NSAppearance instead of UITraitCollection. Any suggestions?

## Checklist

### DIFF
--- a/Sources/SwifterSwift/UIKit/UIColorExtensions.swift
+++ b/Sources/SwifterSwift/UIKit/UIColorExtensions.swift
@@ -6,6 +6,7 @@
 //  Copyright © 2019 SwifterSwift
 //
 
+#if canImport(UIKit)
 import UIKit
 
 public extension UIColor {
@@ -21,3 +22,5 @@ public extension UIColor {
     }
 
 }
+
+#endif
--- a/Sources/SwifterSwift/UIKit/UIColorExtensions.swift
+++ b/Sources/SwifterSwift/UIKit/UIColorExtensions.swift
@@ -1,0 +1,23 @@
+//
+//  UIColorExtensions.swift
+//  SwifterSwift
+//
+//  Created by Max Haertwig on 10/06/19.
+//  Copyright © 2019 SwifterSwift
+//
+
+import UIKit
+
+public extension UIColor {
+
+    /// SwifterSwift: Create a UIColor with different colors for light and dark mode.
+    ///
+    /// - Parameters:
+    ///     - light: Color to use in light/unspecified mode.
+    ///     - dark: Color to use in dark mode.
+    @available(iOS 13.0, tvOS 13.0, *)
+    convenience init(light: UIColor, dark: UIColor) {
+        self.init(dynamicProvider: { $0.userInterfaceStyle == .dark ? dark : light })
+    }
+
+}
--- a/Tests/AppKitTests/NSColorExtensionsTests.swift
+++ b/Tests/AppKitTests/NSColorExtensionsTests.swift
@@ -1,0 +1,33 @@
+//
+//  UIColorExtensionsTests.swift
+//  SwifterSwift
+//
+//  Created by Max Haertwig on 10/08/19.
+//  Copyright © 2019 SwifterSwift
+//
+
+import Foundation
+
+import XCTest
+@testable import SwifterSwift
+
+final class NSColorExtensionsTests: XCTestCase {
+    
+    @available(OSX 10.15, *)
+    func testInitLightDark() {
+        let lightModeColor = NSColor.red
+        let darkModeColor = NSColor.blue
+        let color = NSColor(light: lightModeColor, dark: darkModeColor)
+        
+        let view = NSView()
+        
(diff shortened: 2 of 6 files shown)


Sources/SwifterSwift/UIKit/UIColorExtensions.swift
```diff
@@ -0,0 +1,23 @@
+//
+//  UIColorExtensions.swift
+//  SwifterSwift
+//
+//  Created by Max Haertwig on 10/06/19.
+//  Copyright © 2019 SwifterSwift
+//
+
```
LucianoPAlmeida: Missing #if canImport(UIKit)

MaxHaertwig: Thanks

MaxHaertwig:
I'd like to add a test for `NSColor`, but I don't know how to set an NSAppearance without having a view hierarchy. Any suggestions?

guykogus:
@MaxHaertwig I don't think there's any way around the view hierarchy. I ran this in Playgrounds and it works:
```Swift
let color = NSColor(light: .red, dark: .blue)
let view = NSView()
view.wantsLayer = true

NSAppearance.current = NSAppearance(named: .aqua)
view.layer?.backgroundColor = color.cgColor
view.layer?.backgroundColor == NSColor.red.cgColor

NSAppearance.current = NSAppearance(named: .darkAqua)
view.layer?.backgroundColor = color.cgColor
view.layer?.backgroundColor == NSColor.blue.cgColor
```

MaxHaertwig:
Thanks @guykogus, I added a test.

Tests/AppKitTests/NSColorExtensionsTests.swift
```diff
@@ -0,0 +1,33 @@
+//
+//  UIColorExtensionsTests.swift
+//  SwifterSwift
+//
+//  Created by Max Haertwig on 10/08/19.
+//  Copyright © 2019 SwifterSwift
+//
+
+import Foundation
```
guykogus: Shouldn't it be the same as the source file?
```Swift
#if canImport(Cocoa)
import Cocoa
```

MaxHaertwig: Of course

MaxHaertwig:
@guykogus, @LucianoPAlmeida do you know why the travis build failes?

LucianoPAlmeida:
Seems something on Danger, I'll take a look  as soon as possible :))

guykogus:
@MaxHaertwig check that `NSColorExtensions.swift` was added to the `SwifterSwift-macOS` target?

MaxHaertwig:
It belongs to `SwifterSwift-macOS `.

MaxHaertwig:
I fixed the error. However, I don't know why codecov is complaining. Any ideas?

guykogus:
@MaxHaertwig codecov can be a bit strange sometimes. I wouldn't worry about it. It's complaining about a decrease in code coverage from a recent `UIImage` extension.

MaxHaertwig:
Is it possible to run the Travis build on a macOS 10.15 machine?

guykogus:
@MaxHaertwig not yet https://docs.travis-ci.com/user/reference/osx/

VincentSit:
@LucianoPAlmeida @guykogus 

I think this method should be changed to the following to be more compatible.
```swift
convenience init?(light: UIColor?, dark: UIColor?) {
    guard let light = light, let dark = dark else {
      return nil
    }
    
    if #available(iOS 13.0, *) {
      self.init(dynamicProvider: { $0.userInterfaceStyle == .dark ? dark : light })
    } else {
      self.init(cgColor: light.cgColor)
    }
  }
```

Otherwise you will encounter such errors in projects with a minimum deployment version lower than iOS 13.

image

The reason for using optional variables is to be compatible with the colors loaded from the bundle.

guykogus:
What does your code look like that the colours would be optional?
The main issue with changing this to be a nullable initialiser is that it's a breaking change. You could just make `dark` optional, which may help and still keeps things simple.
```Swift
convenience init(light: UIColor, dark: UIColor?) {
    if #available(iOS 13.0, tvOS 13.0, *),
        let dark = dark {
        self.init(dynamicProvider: { $0.userInterfaceStyle == .dark ? dark : light })
    } else {
        self.init(cgColor: light.cgColor)
    }
}
```

VincentSit:
@guykogus 

Because I use R.swift, loading the color from the bundle (via `init(named:)`) will return an optional value. Your suggestion is great, I did not expect it to cause a breaking change.

guykogus:
My suggestions are usually great 😏 (that's a lie, never listen to me)
Feel free to create a PR 🙂

VincentSit:
😆 Thank you, will do it shortly.

PR #725, opened 2019-10-08
Add custom SwiftLint rules

omaralbeik:
This PR adds some custom [SwiftLint](https://github.com/realm/SwiftLint) rules including checking for the `SwifterSwift: ` prefix in doc strings

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -27,11 +27,6 @@ custom_rules:
     regex: '(\t+| +)//\n'
     message: "Remove useless comment lines or use /* format */"
     severity: warning
-  invalid_mark_format:
-    name: "Invalid MARK Format"
-    regex: '(?m-s)(\/\/[\s]*?MARK(?!(\:[\s]{1}\-[\s]{1}){1}))'
-    message: "Use format: MARK: - Your Info"
-    severity: warning
   get_prefixed_function:
     name: "Get prefixed function"
     regex: '(?-smxi)(func get[A-Z0-9][\w]+?(\())'
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -24,7 +24,7 @@ custom_rules:
       - doccomment.field
   empty_commented_line:
     name: "Empty Commented Out Line"
-    regex: '(\t+| +)//\n'
+    regex: '(\t+| +)//$'
     message: "Remove useless comment lines or use /* format */"
     severity: warning
   get_prefixed_function:
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -1,3 +1,43 @@
+custom_rules:
+  missing_swifterswift_prefix:
+    included: ".*\\.swift"
+    excluded: ".*Tests\\.swift"
+    name: "Missing SwifterSwift: prefix"
+    regex: "///\\s(?!SwifterSwift:\\s)\\w+[^\\n]*$"
+    match_kinds:
+      - doccomment
+    message: "Doc string should always start with the SwifterSwift: prefix"
+    severity: warning
+  unnecessary_type:
+    name: "Unnecessary Type"
+    regex: '(?sm)[ \ta-zA-Z0-9]?(?:let|var){1} [ \ta-zA-Z0-9]+?:[ \t]+?([a-zA-Z0-9]+?)[\t ]+?=[\t ]?\1'
+    message: "Type Definition Not Needed"
+    severity: warning
+  comments_space:
+    name: "Space After Comment"
+    regex: '(//\w+|///\w+)'
+    message: "There should be a space between the start of a comment and the first character."
+    severity: warning
+    match_kinds:
+      - comment
+      - doccomment
+      - doccomment.field
+  empty_commented_line:
+    name: "Empty Commented Out Line"
+    regex: '(\t+| +)//\n'
+    message: "Remove useless comment lines or use /* format */"
+    severity: warning
+  invalid_mark_format:
+    name: "Invalid MARK Format"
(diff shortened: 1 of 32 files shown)


.swiftlint.yml
```diff
@@ -1,3 +1,43 @@
...
+    excluded: ".*Tests\\.swift"
+    name: "Missing SwifterSwift: prefix"
+    regex: "///\\s(?!SwifterSwift:\\s)\\w+[^\\n]*$"
+    match_kinds:
+      - doccomment
+    message: "Doc string should always start with the SwifterSwift: prefix"
+    severity: warning
+  unnecessary_type:
+    name: "Unnecessary Type"
+    regex: '(?sm)[ \ta-zA-Z0-9]?(?:let|var){1} [ \ta-zA-Z0-9]+?:[ \t]+?([a-zA-Z0-9]+?)[\t ]+?=[\t ]?\1'
```
LucianoPAlmeida: This would not allow things like 
```swift
let s: String = getString()
```
?
That might give false positive because I think sometimes this use of types is necessary to disambiguate expressions.

.swiftlint.yml
```diff
@@ -1,3 +1,43 @@
...
+    match_kinds:
+      - comment
+      - doccomment
+      - doccomment.field
+  empty_commented_line:
+    name: "Empty Commented Out Line"
+    regex: '(\t+| +)//\n'
+    message: "Remove useless comment lines or use /* format */"
+    severity: warning
+  invalid_mark_format:
```
LucianoPAlmeida: Does this one already exists I think - [Mark](https://github.com/realm/SwiftLint/blob/2dcaf3ee7803e25c8163e2fbbb9b8ac2be722a2d/Rules.md#L11971)

Tests/UIKitTests/UIViewExtensionsTests.swift
```diff
@@ -298,6 +298,22 @@ final class UIViewExtensionsTests: XCTestCase {
         waitForExpectations(timeout: 0.5, handler: nil)
     }
 
+    func testScale() {
```
LucianoPAlmeida: This one is related to the lint? 😆

LucianoPAlmeida:
Cool stuff @omaralbeik 💯 🎉

.swiftlint.yml
omaralbeik: Sure, I'm not saying this rule is perfect 😁, at least it helps in some scenarios like
```swift
let date: Date = Date()
```
In the cases where having the type is required to disambiguate expressions, we can always add an exception, or if we see this rule bothering us a lot in the future, we can remove it of course

.swiftlint.yml
omaralbeik: Yep, will remove it

Tests/UIKitTests/UIViewExtensionsTests.swift
omaralbeik: I saw Codecov was not happy with the build, I thought maybe I can add a test or two to increase the coverage, turned out it is still not happy ¯\_(ツ)_/¯

.swiftlint.yml
LucianoPAlmeida: Sure @omaralbeik :)) I just mentioned it because I see this being used a lot (not here in this project but overall) to disambiguate, specially literals like `let s: NString = ""`  since if we do like `let s = ""` would be infered as String. 
 Of course we can always use `let s = "" as NSString` :))

Tests/UIKitTests/UIViewExtensionsTests.swift
LucianoPAlmeida: Ohh sure :)

.swiftlint.yml
```diff
@@ -1,3 +1,38 @@
...
+    message: "Doc string should always start with the SwifterSwift: prefix"
+    severity: warning
+  unnecessary_type:
+    name: "Unnecessary Type"
+    regex: '(?sm)[ \ta-zA-Z0-9]?(?:let|var){1} [ \ta-zA-Z0-9]+?:[ \t]+?([a-zA-Z0-9]+?)[\t ]+?=[\t ]?\1'
+    message: "Type Definition Not Needed"
+    severity: warning
+  comments_space:
+    name: "Space After Comment"
+    regex: '(//\w+|///\w+)'
```
guykogus: Shouldn't the regex be for a non-space char?
```suggestion
    regex: '(//\S|///\S)'
```

.swiftlint.yml
```diff
@@ -1,3 +1,38 @@
...
+    message: "There should be a space between the start of a comment and the first character."
+    severity: warning
+    match_kinds:
+      - comment
+      - doccomment
+      - doccomment.field
+  empty_commented_line:
+    name: "Empty Commented Out Line"
+    regex: '(\t+| +)//\n'
```
guykogus: Again, can you check for whitespace chars? Or will this match newlines too?
```suggestion
    regex: '(\s+)//\n'
```
If `\s` matches newlines we can also do
```suggestion
    regex: '^(\s+)//$'
```

.swiftlint.yml
omaralbeik: No, because using a non-space char will also match the playground markdown
```
//: # Title
```

.swiftlint.yml
omaralbeik: 1. No, doing so will match lines like `//  SwifterSwift`
2. Yes, `$` looks better here

omaralbeik:
@guykogus I see this PR will be useful for PRs like [#727](https://github.com/SwifterSwift/SwifterSwift/pull/727) 😜

LucianoPAlmeida:
@omaralbeik @guykogus Any still discussion open in this one? Or good to go?

PR #727, opened 2019-10-09
Ns regular expression extensions

guykogus:
🚀
No more using `NSString`, `NSRange` and `UnsafeMutablePointer<ObjCBool>` with `NSRegularExpression` calls.

## Checklist

### DIFF
--- a/Sources/SwifterSwift/Foundation/NSRegularExpressionExtensions.swift
+++ b/Sources/SwifterSwift/Foundation/NSRegularExpressionExtensions.swift
@@ -1,0 +1,144 @@
+//
+//  NSRegularExpressionExtensions.swift
+//  SwifterSwift
+//
+//  Created by Guy Kogus on 09/10/2019.
+//  Copyright © 2019 SwifterSwift
+//
+
+#if canImport(Foundation)
+import Foundation
+
+public extension NSRegularExpression {
+    /// Enumerates the string allowing the Block to handle each regular expression match.
+    ///
+    /// - Parameters:
+    ///   - string: The string.
+    ///   - options: The matching options to report. See `NSRegularExpression.MatchingOptions` for the supported values.
+    ///   - range: The range of the string to test.
+    ///   - block: The Block enumerates the matches of the regular expression in the string.
+    ///     The block takes three arguments and returns `Void`:
+    ///   - result:
+    ///     An `NSTextCheckingResult` specifying the match. This result gives the overall matched range via its `range` property, and the range of each individual capture group via its `range(at:)` method. The range {NSNotFound, 0} is returned if one of the capture groups did not participate in this particular match.
+    ///   - flags:
+    ///     The current state of the matching progress. See `NSRegularExpression.MatchingFlags` for the possible values.
+    ///   - stop:
+    ///     A reference to a Boolean value. The Block can set the value to true to stop further processing of the array. The stop argument is an out-only argument. You should only ever set this Boolean to true within the Block.
+    func enumerateMatches(in string: String,
+                          options: MatchingOptions = [],
+                          range: Range<String.Index>,
+                          using block: (_ result: NSTextCheckingResult?, _ flags: MatchingFlags, _ stop: inout Bool) -> Void) {
+        enumerateMatches(in: string,
+                         options: options,
+                         range: NSRange(range, in: string)) { result, flags, stop in
+                            var shouldStop = false
+                            block(result, flags, &shouldStop)
+                            if shouldStop {
+                                stop.pointee = true
+                            }
+        }
+    }
+
+    /// Returns an array containing all the matches of the regular expression in the string.
+    ///
+    /// - Parameters:
+    ///   - string: The string to search.
+    ///   - options: The matching options to use. See NSRegularExpression.MatchingOptions for possible values.
+    ///   - range: The range of the string to search.
+    /// - Returns: An array of `NSTextCheckingResult` objects. Each result gives the overall matched range via its `range` property, and the range of each individual capture group via its `range(at:)` method. The range {NSNotFound, 0} is returned if one of the capture groups did not participate in this particular match.
+    func matches(in string: String,
+                 options: MatchingOptions = [],
+                 range: Range<String.Index>) -> [NSTextCheckingResult] {
+        return matches(in: string,
+                       options: options,
+                       range: NSRange(range, in: string))
+    }
+
+    /// Returns the number of matches of the regular expression within the specified range of the string.
+    ///
+    /// - Parameters:
+    ///   - string: The string to search.
+    ///   - options: The matching options to use. See NSRegularExpression.MatchingOptions for possible values.
+    ///   - range: The range of the string to search.
+    /// - Returns: The number of matches of the regular expression.
+    func numberOfMatches(in string: String,
+                         options: MatchingOptions = [],
+                         range: Range<String.Index>) -> Int {
+        return numberOfMatches(in: string,
+                               options: options,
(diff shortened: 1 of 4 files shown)


Sources/SwifterSwift/Foundation/NSRegularExpressionExtensions.swift
```diff
@@ -0,0 +1,146 @@
...
+    ///
+    /// - Parameters:
+    ///   - string: The string to search.
+    ///   - options: The matching options to use. See `NSRegularExpression.MatchingOptions` for possible values.
+    ///   - range: The range of the string to search.
+    /// - Returns: The range of the first match. Returns `nil` if no match is found.
+    func rangeOfFirstMatch(in string: String,
+                           options: MatchingOptions = [],
+                           range: Range<String.Index>) -> Range<String.Index>? {
+        return Range(rangeOfFirstMatch(in: string,
```
LucianoPAlmeida: If we are always creating a range and returning this will never be nil right?

Sources/SwifterSwift/Foundation/NSRegularExpressionExtensions.swift
```diff
@@ -0,0 +1,146 @@
...
+    ///   - stop:
+    ///     A reference to a Boolean value. The Block can set the value to true to stop further processing of the array. The stop argument is an out-only argument. You should only ever set this Boolean to true within the Block.
+    func enumerateMatches(in string: String,
+                          options: MatchingOptions = [],
+                          range: Range<String.Index>,
+                          using block: (_ result: NSTextCheckingResult?, _ flags: MatchingFlags, _ stop: inout Bool) -> Void) {
+        enumerateMatches(in: string,
+                         options: options,
+                         range: NSRange(range, in: string)) { result, flags, stop in
+                            var shouldStop = false
```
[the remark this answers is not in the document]
guykogus: I don't really love it because it forces a return value, which for the most part will be `false`, instead of just letting users set the `stop` param when necessary. You can see what I mean in my unit test, which would become:
```Swift
func testEnumerateMatches() {
    var count = 0
    regularExpression.enumerateMatches(in: string,
                                       options: [],
                                       range: string.startIndex..<string.endIndex) { (result, _) in
                                        XCTAssertEqual(String(string[Range(result!.range, in: string)!]), searchString)
                                        count += 1
                                        return false
    }
    XCTAssertEqual(count, expectedMatches)

    count = 0
    let max = 2 // must be less than expectedMatches
    regularExpression.enumerateMatches(in: string,
                                       options: [],
                                       range: string.startIndex..<string.endIndex) { (result, _) in
                                        XCTAssertEqual(String(string[Range(result!.range, in: string)!]), searchString)
                                        count += 1
                                        return count >= max
    }
    XCTAssertEqual(count, max)
}
```

Sources/SwifterSwift/Foundation/NSRegularExpressionExtensions.swift
```diff
@@ -0,0 +1,146 @@
...
+    /// Returns an array containing all the matches of the regular expression in the string.
+    ///
+    /// - Parameters:
+    ///   - string: The string to search.
+    ///   - options: The matching options to use. See NSRegularExpression.MatchingOptions for possible values.
+    ///   - range: The range of the string to search.
+    /// - Returns: An array of `NSTextCheckingResult` objects. Each result gives the overall matched range via its `range` property, and the range of each individual capture group via its `range(at:)` method. The range {NSNotFound, 0} is returned if one of the capture groups did not participate in this particular match.
+    func matches(in string: String,
+                 options: MatchingOptions = [],
+                 range: Range<String.Index>) -> [NSTextCheckingResult] {
```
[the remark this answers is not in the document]
guykogus: Unfortunately you can't reference a parameter to create the default value for another parameter. What I could do is create overloads that don't include the `range` parameter and use the entire string.

Sources/SwifterSwift/Foundation/NSRegularExpressionExtensions.swift
guykogus: This will be `nil` if the regex can't be found in the string. It's why I added [this test](https://github.com/SwifterSwift/SwifterSwift/pull/727/commits/f2750fd58c74bbbef864a255a56c2f90a70a9701).

Sources/SwifterSwift/Foundation/NSRegularExpressionExtensions.swift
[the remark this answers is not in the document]
LucianoPAlmeida: I'm fine with it, was just a question. I see the point, makes sense :))
It just that I personally prefer return, because from the caller propective is always more clear when return (at least to me).

Sources/SwifterSwift/Foundation/NSRegularExpressionExtensions.swift
LucianoPAlmeida: Nice :))

Sources/SwifterSwift/Foundation/NSRegularExpressionExtensions.swift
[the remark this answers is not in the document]
LucianoPAlmeida: That's sad :(( 
Not sure if it is worthed then

guykogus:
@LucianoPAlmeida looks like the Linux implementation treats the block as `@escaping`. Does this look ok? https://github.com/SwifterSwift/SwifterSwift/pull/727/commits/6fe2f92bc4ed29fd4ca91a108f69a003773e4ae9

Sources/SwifterSwift/Foundation/NSRegularExpressionExtensions.swift
```diff
@@ -0,0 +1,163 @@
...
+    ///     An `NSTextCheckingResult` specifying the match. This result gives the overall matched range via its `range` property, and the range of each individual capture group via its `range(at:)` method. The range {NSNotFound, 0} is returned if one of the capture groups did not participate in this particular match.
+    ///   - flags:
+    ///     The current state of the matching progress. See `NSRegularExpression.MatchingFlags` for the possible values.
+    ///   - stop:
+    ///     A reference to a Boolean value. The Block can set the value to true to stop further processing of the array. The stop argument is an out-only argument. You should only ever set this Boolean to true within the Block.
+    #if os(Linux)
+    func enumerateMatches(in string: String,
+                          options: MatchingOptions = [],
+                          range: Range<String.Index>,
+                          using block: @escaping (_ result: NSTextCheckingResult?, _ flags: MatchingFlags, _ stop: inout Bool) -> Void) {
```
LucianoPAlmeida: Just curiosity, do you know exactly why this is need for linux? CoreFoundation API requires escaping?

guykogus: The travis build was failing without it: https://travis-ci.org/SwifterSwift/SwifterSwift/jobs/595511815
In the Linux implementation of `enumerateMatches(in:options:range:using)`, `block` is `@escaping`, so if I then try to call a non-escaping block inside it the compiler throws an error.

LucianoPAlmeida: Seems like it 🙃

LucianoPAlmeida: Looks like an implementation detail requirement where the have to store the closure in a [object](https://github.com/apple/swift-corelibs-foundation/blob/432d15abe7aecb6698ab4458c709e2ce97412075/Foundation/NSRegularExpression.swift#L179), making it being required to be escaping

Issue #552, opened 2018-08-25
Add Date.numberOfDaysInYear


nuomi1:
> Please fill out this template when filing an issue.
>
> All comment lines beginning with an **>** instruct you with what info we expect. You can delete those lines once you've filled in the info.
>

[x] I've read, understood, and done my best to follow the [Contributing guidelines](https://github.com/SwifterSwift/SwifterSwift/tree/master/CONTRIBUTING.md) before opening this issue.

### What did you do?
> Provide a compute property.

### What did you expect to happen?
> Get days in year for `Date`.

### What happened instead?
> Please replace this with of what happened instead.

### SwifterSwift Environment

- SwifterSwift version: 4.4.0
- Xcode version: 9.4.1
- macOS version running Xcode: 10.13.6
- Swift version: 4.1.2
- Platform(s) running SwifterSwift: macOS

### Demo Project
``` Swift
extension Date {
    var numberOfDaysInYear: Int {
        return Calendar.current.ordinality(of: .day, in: .year, for: self)!
    }
}
```

``` Swift
let now = Date()

dump(now)
print(now.numberOfDaysInYear)
```

``` Text
▿ 2018-08-25 08:43:25 +0000
  - timeIntervalSinceReferenceDate: 556879405.01351404
237
```

pvtmert:
name should be `numberOfDaysLeftInYear` or something like it.
because number-of-days-in-year is either 365 or 366 depending on leap-year

guykogus:
Actually the code example is `dayInYear`. To get `numberOfDaysLeftInYear` it would be calculated as such:
```Swift
extension Date {
    var dayInYear: Int {
        return Calendar.current.ordinality(of: .day, in: .year, for: self)!
    }

    var numberOfDaysInYear: Int {
        return Calendar.current.range(of: .day, in: .year, for: self)!.upperBound - 1
    }

    var numberOfDaysLeftInYear: Int {
        return numberOfDaysInYear - dayInYear
    }
}

Date().numberOfDaysLeftInYear // 81 (for Oct 11, 2018)
```

PR #730, opened 2019-10-12
Added "numberOfDaysLeftInYear" to "Date" extension

Hero93:
I provided the requested computed property for the Date Extension in
https://github.com/SwifterSwift/SwifterSwift/issues/552.
This is my first pull-request, I hope I made everything correctly.
Thanks for reading :).

#552 Added "numberOfDaysLeftInYear" to "Date" extension

- [ x] I checked the [**Contributing Guidelines**](https://github.com/SwifterSwift/SwifterSwift/blob/master/CONTRIBUTING.md) before creating this request.
- [x ] New extensions are written in Swift 5.0.
- [x ] New extensions support iOS 8.0+ / tvOS 9.0+ / macOS 10.10+ / watchOS 2.0+, or use `@available` if not.
- [x ] I have added tests for new extensions, and they passed.


LucianoPAlmeida:
Hey @Hero93. Welcome to the project \o/ 

Some points to address before the review:
* Make sure to run swiftlint autocorrect to fix the trailing_whitespace issues.
* We have a guideline that enforces a one-to-one extension/test relation meaning if you have 3 extensions, you may add 3 test methods one for testing each.
* Can you add a changelog entry for those changes?

Hero93:
> Hey @Hero93. Welcome to the project \o/
> 
> Some points to address before the review:
> 
> * Make sure to run swiftlint autocorrect to fix the trailing_whitespace issues.
> * We have a guideline that enforces a one-to-one extension/test relation meaning if you have 3 extensions, you may add 3 test methods one for testing each.
> * Can you add a changelog entry for those changes?

Thanks very much @LucianoPAlmeida for the feedbacks!
I fixed the points you wrote and I'll add the changelog entry as soon as possible :).
Thanks,
Luca.

Hero93:
> > Hey @Hero93. Welcome to the project \o/
> > Some points to address before the review:
> > 
> > * Make sure to run swiftlint autocorrect to fix the trailing_whitespace issues.
> > * We have a guideline that enforces a one-to-one extension/test relation meaning if you have 3 extensions, you may add 3 test methods one for testing each.
> > * Can you add a changelog entry for those changes?
> 
> Thanks very much @LucianoPAlmeida for the feedbacks!
> I fixed the points you wrote and I'll add the changelog entry as soon as possible :).
> Thanks,
> Luca.

Done with the changelog :) :)
Luca.

Sources/SwifterSwift/Foundation/DateExtensions.swift
```diff
@@ -487,6 +487,20 @@ public extension Date {
         return timeIntervalSince1970
     }
 
+    /// SwifterSwift: Current day in a year
+    var dayInYear: Int {
+        return calendar.ordinality(of: .day, in: .year, for: self)!
```
LucianoPAlmeida: This could be nil by any chance?

Sources/SwifterSwift/Foundation/DateExtensions.swift
```diff
@@ -487,6 +487,20 @@ public extension Date {
...
+    }
+
+    /// SwifterSwift: How many days in a year
+    var numberOfDaysInYear: Int {
+        return (calendar.range(of: .day, in: .year, for: self)!.upperBound) - 1
```
LucianoPAlmeida: Same question about nil here too :)

Sources/SwifterSwift/Foundation/DateExtensions.swift
```diff
@@ -487,6 +487,20 @@ public extension Date {
...
+    }
+
+    /// SwifterSwift: How many days are left in a year
+    var numberOfDaysLeftInYear: Int {
```
LucianoPAlmeida: What about `remainingDays(in: .year)`, that would be a more general extension I think :))

Sources/SwifterSwift/Foundation/DateExtensions.swift
Hero93: I don't thinks so. The documentation states that :
Returns
The ordinal number of smaller within larger at the time specified by date. Returns nil if larger is not logically bigger than smaller in the calendar, or the given combination of components does not make sense (or is a computation which is undefined).
Day is always smaller than year, so nil is not thrown. 
What do you think ? Did I understand correctly ?

Sources/SwifterSwift/Foundation/DateExtensions.swift
Hero93: Same response as the previous.
smaller calendar component, day, is always smaller than larger calendar component year.
If I understood correctly it won't be nil by any chance :)

Sources/SwifterSwift/Foundation/DateExtensions.swift
Hero93: That's a great idea! Thanks so much!
So this method will be useful for both month and year, right ?

Sources/SwifterSwift/Foundation/DateExtensions.swift
LucianoPAlmeida: I don't have deep knowledge about calendars, so my question is more about that depending on the calendar used if this could by any chance crash. I assume our tests use always gregorian, so for this, I think it is never nil. But honestly no idea about others 😅

Sources/SwifterSwift/Foundation/DateExtensions.swift
LucianoPAlmeida: Yeap :)

Sources/SwifterSwift/Foundation/DateExtensions.swift
guykogus: In my branch for upgrading to Xcode 11.1 I found that this crashed with `.nanoseconds` -> `.seconds` but only on Linux builds. You never know when some implementation can go horribly wrong.
That being said, I think it's safe to assume that you can get which number of day in the year a specific date is, regardless of the calendar.

Sources/SwifterSwift/Foundation/DateExtensions.swift
guykogus: But then you'd need it to return an optional, because if the caller was `remainingDays(in: .seconds)` you would get back `nil`.

Tests/FoundationTests/DateExtensionsTests.swift
```diff
@@ -970,6 +970,31 @@ final class DateExtensionsTests: XCTestCase {
...
+    func testnNmberOfDaysLeftInYear() {
+        let date = Date()
+        let dayLeftInAYear = date.numberOfDaysLeftInYear
+
+        let dayInYear = Calendar.current.ordinality(of: .day, in: .year, for: date)!
+        let numberOfDaysInYear = (Calendar.current.range(of: .day, in: .year, for: date)!.upperBound) - 1
+        let dayLeftInAYearCheck = numberOfDaysInYear - dayInYear
+
+        XCTAssertEqual(dayLeftInAYear, dayLeftInAYearCheck)
+    }
```
guykogus: This is not a valid way to do unit tests. You're performing the function and what you know is its implementation and checking that they're equal, which logically they will be (unless someone changes either the test or the implementation).
What you should be doing instead is using fixed values, i.e. a specific date where you know how many days are in that year, which day of the year it is and how many days left in that year.

LucianoPAlmeida: +1 Here

Hero93: @guykogus Many thanks for the comment!
I'll re-write the tests to make it correct :)

Sources/SwifterSwift/Foundation/DateExtensions.swift
chulbert: This is picky but by Apple's naming conventions this should _probably_ be `daysRemaining(in:)`

LucianoPAlmeida: @chulbert That makes sense :)

pbeyer08:
Is it safe to say this PR just needs a little unit testing TLC and should be good to go?

Issue #732, opened 2019-10-13
UIImage.filled(withColor:) improvement


LucianoPAlmeida:
### Hacktoberfest 

Here we can do a minor tweak where this extension is implemented using an old `UIGraphicsBeginImageContextWithOptions` API. That is not recommended. 
We can provide an improvement for it and make this using `UIGraphicsImageRenderer` API.

Note: `UIGraphicsImageRenderer` is only available from iOS 10. So we should maintain the current implementation for lower iOS versions like:
```swift
if #available(iOS 10.0, tvOS 10.0, *) {
        // UIGraphicsImageRenderer new implementation
} else {
       // Current implementation with UIGraphicsBeginImageContextWithOptions
}
```
See #721 for an example.

FraDeliro:
Hi @LucianoPAlmeida I could pick up this one :)

raysarebest:
@FraDeliro Sorry for not commenting, before I started, but I was literally just about to click the button to make a PR for this 😅

FraDeliro:
@raysarebest No worries :) go for it 👍

LucianoPAlmeida:
@FraDeliro @raysarebest Thanks for taking this one, @FraDeliro there are another issues like this one you could take, just make sure to comment on that so people know there’s someone already working :))

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

PR #733, opened 2019-10-13
🚀 UIGraphicsImageRenderer in filled(withColor:)

raysarebest:
🚀 This provides a fix for Issue #732. In iOS/tvOS 10, Apple introduced the [UIGraphicsImageRenderer API](https://developer.apple.com/documentation/uikit/uigraphicsimagerenderer), which effectively replaces the older `UIGraphicsImageContext` method this method was using before

## Checklist

I didn't add any new extensions in this PR. I changed the internals of one to use a newer API, but everything looks exactly the same to a consumer of the method, and all the pre-existing tests pass. Do I need to check any of the other boxes?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,9 +13,10 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
   - Added `enumerateMatches(in:options:range:using:)`, `matches(in:options:range:)`, `numberOfMatches(in:options:range:)`, `firstMatch(in:options:range:)`, `rangeOfFirstMatch(in:options:range:)`, `stringByReplacingMatches(in:options:range:withTemplate:)`, `replaceMatches(in:options:range:withTemplate:)`, which use `String` and `String.Range` in place of `NSString` and `NSRange` to make the calls Swifter. [#727](https://github.com/SwifterSwift/SwifterSwift/pull/727) by [guykogus](https://github.com/guykogus).
 - **UIImage**:
   - Added `withBackgroundColor(_:)` to specify a background color for a partially transparent image. [#721](https://github.com/SwifterSwift/SwifterSwift/pull/721) by [MaxHaertwig](https://github.com/maxhaertwig).
-  - Implemented `filled(withColor:)` using `UIGraphicsImageRenderer` when available. [#733](https://github.com/SwifterSwift/SwifterSwift/pull/733)
 
 ### Changed
+- **UIImage**:
+  - Implemented `filled(withColor:)` using `UIGraphicsImageRenderer` when available. [#733](https://github.com/SwifterSwift/SwifterSwift/pull/733)
 
 ### Deprecated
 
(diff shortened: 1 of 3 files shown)


Sources/SwifterSwift/UIKit/UIImageExtensions.swift
```diff
@@ -159,28 +159,40 @@ public extension UIImage {
         return newImage
     }
 
+    #if !os(watchOS)
```
LucianoPAlmeida: Do we really need that addition?

raysarebest:
> It's a shame watchOS support has to be dropped. Could you could copy the function and surround it with `@available(watchOS, *)` with the old implementation?

I bet I can actually configure the OS macro in such a way that it'll only compile the new way on platforms that're not watchOS. That way, we don't have to maintain basically the same code twice. I'll see if I can get that patched when I get home from work later

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
   - Added `enumerateMatches(in:options:range:using:)`, `matches(in:options:range:)`, `numberOfMatches(in:options:range:)`, `firstMatch(in:options:range:)`, `rangeOfFirstMatch(in:options:range:)`, `stringByReplacingMatches(in:options:range:withTemplate:)`, `replaceMatches(in:options:range:withTemplate:)`, which use `String` and `String.Range` in place of `NSString` and `NSRange` to make the calls Swifter. [#727](https://github.com/SwifterSwift/SwifterSwift/pull/727) by [guykogus](https://github.com/guykogus).
 - **UIImage**:
   - Added `withBackgroundColor(_:)` to specify a background color for a partially transparent image. [#721](https://github.com/SwifterSwift/SwifterSwift/pull/721) by [MaxHaertwig](https://github.com/maxhaertwig).
+  - Implemented `filled(withColor:)` using `UIGraphicsImageRenderer` when available. [#733](https://github.com/SwifterSwift/SwifterSwift/pull/733)
```
guykogus: This should be in the `Changed` section, not `Added`.

SwifterSwift.xcodeproj/project.pbxproj
```diff
@@ -241,11 +241,7 @@
 		07C50D491F5EB04700F46E5A /* UINavigationBarExtensionTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 07C50D161F5EB03200F46E5A /* UINavigationBarExtensionTests.swift */; };
 		07C50D4A1F5EB04700F46E5A /* UINavigationControllerExtensionsTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 07C50D171F5EB03200F46E5A /* UINavigationControllerExtensionsTests.swift */; };
 		07C50D4B1F5EB04700F46E5A /* UINavigationItemExtensionsTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 07C50D181F5EB03200F46E5A /* UINavigationItemExtensionsTests.swift */; };
-		07C50D4C1F5EB04700F46E5A /* UISearchBarExtensionsTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 07C50D191F5EB03200F46E5A /* UISearchBarExtensionsTests.swift */; };
```
guykogus: Please revert all of these changes.

Sources/SwifterSwift/UIKit/UIImageExtensions.swift
```diff
@@ -164,6 +164,19 @@ public extension UIImage {
     /// - Parameter color: color to fill image with.
     /// - Returns: UIImage filled with given color.
     func filled(withColor color: UIColor) -> UIImage {
+
+        #if !os(watchOS)
+        if #available(iOS 10, tvOS 10, *) {
+            let format = UIGraphicsImageRendererFormat()
+            format.scale = scale
+            let renderer = UIGraphicsImageRenderer(size: size, format: format)
+            return renderer.image { (context) in
```
guykogus:
```suggestion
            return renderer.image { context in
```

SwifterSwift.xcodeproj/project.pbxproj
raysarebest: Will do, but I feel like this should be PRed separately. There were a bunch of tests wrapped in `#if os(iOS)` added to the tvOS target. They'd never run because there isn't anything to compile for that platform, but because they exist and didn't pass anything, it'd make Xcode think the tests for tvOS weren't entirely successful

raysarebest:
I've addressed the changes @guykogus requested, but it seems like the tvOS build failed, despite all targets compiling and tests passing on my machine. It looks like it's because its call to `xcodebuild` to compile the target exited with code `132`, and from googling around a bit, it seems like that's a SourceKit crash that happens sometimes, and not actually anything wrong with the code/project. Is there any way to force a rerun of Travis?

SD10:

Issue #731, opened 2019-10-13
UIImage.tint(color:blendMode:) improvement


LucianoPAlmeida:
#Hacktoberfest 

Here we can do a minor tweak where this extension is implemented using an old `UIGraphicsBeginImageContextWithOptions` API. That is not recommended. 
We can provide an improvement for it and make this using `UIGraphicsImageRenderer` API.

Note: `UIGraphicsImageRenderer` is only available from iOS 10. So we should maintain the current implementation for lower iOS versions like:
```swift
if #available(iOS 10.0, tvOS 10.0, *) {
        // UIGraphicsImageRenderer new implementation
} else {
       // Current implementation with UIGraphicsBeginImageContextWithOptions
}
```
See #721 for an example.

FraDeliro:
Thanks @LucianoPAlmeida, I will pick up this one :)

PR #734, opened 2019-10-13
Improve UIImage.tint(color:blendMode:) using UIGraphicsImageRenderer

FraDeliro:
- This PR improve the UIImage.tint(color:blendMode:) function using UIGraphicsImageRenderer as specified in the #731 issue.

## Checklist

### DIFF
--- a/Sources/SwifterSwift/UIKit/UIImageExtensions.swift
+++ b/Sources/SwifterSwift/UIKit/UIImageExtensions.swift
@@ -195,7 +195,6 @@ public extension UIImage {
         return newImage
     }
 
-    #if !os(watchOS)
     /// SwifterSwift: UIImage tinted with color
     ///
     /// - Parameters:
@@ -204,6 +203,8 @@ public extension UIImage {
     /// - Returns: UIImage tinted with given color.
     func tint(_ color: UIColor, blendMode: CGBlendMode, alpha: CGFloat = 1.0) -> UIImage {
         let drawRect = CGRect(origin: .zero, size: size)
+
+        #if !os(watchOS)
         if #available(iOS 10.0, tvOS 10.0, *) {
             let format = UIGraphicsImageRendererFormat()
             format.scale = scale
@@ -213,6 +214,7 @@ public extension UIImage {
                 draw(in: drawRect, blendMode: blendMode, alpha: alpha)
             }
         }
+        #endif
 
         UIGraphicsBeginImageContextWithOptions(size, false, scale)
         defer {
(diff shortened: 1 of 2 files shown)


Sources/SwifterSwift/UIKit/UIImageExtensions.swift
```diff
@@ -182,26 +182,40 @@ public extension UIImage {
...
     }
 
+    #if !os(watchOS)
     /// SwifterSwift: UIImage tinted with color
     ///
     /// - Parameters:
     ///   - color: color to tint image with.
     ///   - blendMode: how to blend the tint
     /// - Returns: UIImage tinted with given color.
     func tint(_ color: UIColor, blendMode: CGBlendMode) -> UIImage {
```
LucianoPAlmeida: Not related to the issue, but since we are here: Note that we work with an alpha 1.0 always, so would be a good improvement add it as parameter :)
So the function looks like `func tint(_ color: UIColor, blendMode: CGBlendMode, alpha: CGFloat = 1.0) -> UIImage`
It is important to let the alpha with default value 1.0 so we make it optional and don't break existing code ;)

Sources/SwifterSwift/UIKit/UIImageExtensions.swift
```diff
@@ -182,26 +182,40 @@ public extension UIImage {
...
-        defer {
-            UIGraphicsEndImageContext()
+        let drawRect = CGRect(origin: .zero, size: size)
+        if #available(iOS 10.0, tvOS 10.0, *) {
+            let format = UIGraphicsImageRendererFormat()
+            format.scale = scale
+            let renderer = UIGraphicsImageRenderer(
+                size: size,
+                format: format
+            )
```
guykogus: This can all go on 1 line. In fact, if you look at the next function you can even avoid keeping this temporary reference.

Sources/SwifterSwift/UIKit/UIImageExtensions.swift
FraDeliro: It makes sense to me, thanks for the suggestion ;)
I will update it.

Sources/SwifterSwift/UIKit/UIImageExtensions.swift
FraDeliro: It makes sense to me 👍

Sources/SwifterSwift/UIKit/UIImageExtensions.swift
```diff
@@ -195,26 +195,36 @@ public extension UIImage {
         return newImage
     }
 
+    #if !os(watchOS)
```
LucianoPAlmeida: I think the same concept form #733 should apply here, where the watch OS check should be inside the function so we don't simply drop support of it for watchOS.

FraDeliro: Good point @LucianoPAlmeida, so maybe I could update the next function too (`withBackgroundColor(_: UIColor) -> UIImage`) to be consistent, what do you think?

SD10:

PR #736, opened 2019-10-15
Add sum(for keyPath:) to Sequence

### LINKED ISSUE #735: Extend Sequence to get sum by keypath
Extend The `Sequence` type with a `sum(for keyPath:)` function to return the sum of nested number for all objects within a sequence

Examples from [**Swift by Sundell**](https://swiftbysundell.com/articles/reducers-in-swift/)'s article on reducers
```swift
let unreadCount = mailboxes.sum(for: \.unreadMessages.count)
let totalScore = levels.sum(for: \.playerScore)
let meetingDuration = today.meetings.sum(for: \.duration)
```

moritzsternemann:
🚀 Sum up a Numeric property, referenced by KeyPath, of all elements in a sequence. Originally from https://swiftbysundell.com/articles/reducers-in-swift/.

Resolves #735 

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 - **CAGradientLayer**:
   - Added `init(colors:locations:startPoint:endPoint:type:)` convenience initializer. [#726](https://github.com/SwifterSwift/SwifterSwift/pull/726) by [JayMehta97](https://github.com/JayMehta97).
 - **Sequence**:
-  - Added `sum(for:)` to sum up a Numeric property, referenced by KeyPath, of all elements in a sequence. [#736](https://github.com/SwifterSwift/SwifterSwift/pull/736) by [Moritz Sternemann](https://github.com/moritzsternemann).
+  - Added `sum(for:)` to sum up an `AdditiveArithmetic` property, referenced by `KeyPath`, of all elements in a sequence. [#736](https://github.com/SwifterSwift/SwifterSwift/pull/736) by [Moritz Sternemann](https://github.com/moritzsternemann).
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
   - Added `withBackgroundColor(_:)` to specify a background color for a partially transparent image. [#721](https://github.com/SwifterSwift/SwifterSwift/pull/721) by [MaxHaertwig](https://github.com/maxhaertwig).
 - **CAGradientLayer**:
   - Added `init(colors:locations:startPoint:endPoint:type:)` convenience initializer. [#726](https://github.com/SwifterSwift/SwifterSwift/pull/726) by [JayMehta97](https://github.com/JayMehta97).
+- **Sequence**:
+  - Added `sum(for:)` to sum up a Numeric property, referenced by KeyPath, of all elements in a sequence. [#736](https://github.com/SwifterSwift/SwifterSwift/pull/736) by [Moritz Sternemann](https://github.com/moritzsternemann).
 
 ### Changed
 
--- a/Sources/SwifterSwift/SwiftStdlib/SequenceExtensions.swift
+++ b/Sources/SwifterSwift/SwiftStdlib/SequenceExtensions.swift
@@ -203,17 +203,15 @@ public extension Sequence {
         return sorted { $0[keyPath: keyPath] < $1[keyPath: keyPath] }
     }
 
-    /// SwifterSwift: Sum of a `Numeric` property of each `Element` in a `Sequence`.
+    /// SwifterSwift: Sum of a `AdditiveArithmetic` property of each `Element` in a `Sequence`.
     ///
     ///     ["James", "Wade", "Bryant"].sum(for: \.count) -> 15
     ///
-    /// - Parameter keyPath: Key path of the `Numeric` property.
-    /// - Returns: The sum of the `Numeric` propertys at `keyPath`.
-    func sum<T: Numeric>(for keyPath: KeyPath<Element, T>) -> T {
-        // Implementation from: https://swiftbysundell.com/articles/reducers-in-swift/
-        return reduce(0) { sum, element in
-            sum + element[keyPath: keyPath]
-        }
+    /// - Parameter keyPath: Key path of the `AdditiveArithmetic` property.
+    /// - Returns: The sum of the `AdditiveArithmetic` propertys at `keyPath`.
+    func sum<T: AdditiveArithmetic>(for keyPath: KeyPath<Element, T>) -> T {
+        // Inspired by: https://swiftbysundell.com/articles/reducers-in-swift/
+        return reduce(.zero) { $0 + $1[keyPath: keyPath] }
     }
 }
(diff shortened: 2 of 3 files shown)


Sources/SwifterSwift/SwiftStdlib/SequenceExtensions.swift
```diff
@@ -202,6 +202,19 @@ public extension Sequence {
...
         return sorted { $0[keyPath: keyPath] < $1[keyPath: keyPath] }
     }
+
+    /// SwifterSwift: Sum of a `Numeric` property of each `Element` in a `Sequence`.
+    ///
+    ///     ["James", "Wade", "Bryant"].sum(for: \.count) -> 15
+    ///
+    /// - Parameter keyPath: Key path of the `Numeric` property.
+    /// - Returns: The sum of the `Numeric` propertys at `keyPath`.
+    func sum<T: Numeric>(for keyPath: KeyPath<Element, T>) -> T {
```
guykogus: If you're already going for a generic implementation, why not go up a step to `AdditiveArithmetic`?
```suggestion
    func sum<T>(for keyPath: KeyPath<Element, T>) -> T where T: AdditiveArithmetic {
```

Sources/SwifterSwift/SwiftStdlib/SequenceExtensions.swift
```diff
@@ -202,6 +202,19 @@ public extension Sequence {
...
+        // Implementation from: https://swiftbysundell.com/articles/reducers-in-swift/
+        return reduce(0) { sum, element in
```
guykogus: You can 1-line this :)
```suggestion
        return reduce(.zero) { $0 + $1[keyPath: keyPath] }
```
Also notice using `.zero`, which is part of the protocol, as opposed to `0`, which is an `Int`.

CHANGELOG.md
```diff
@@ -13,6 +13,8 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
   - Added `enumerateMatches(in:options:range:using:)`, `matches(in:options:range:)`, `numberOfMatches(in:options:range:)`, `firstMatch(in:options:range:)`, `rangeOfFirstMatch(in:options:range:)`, `stringByReplacingMatches(in:options:range:withTemplate:)`, `replaceMatches(in:options:range:withTemplate:)`, which use `String` and `String.Range` in place of `NSString` and `NSRange` to make the calls Swifter. [#727](https://github.com/SwifterSwift/SwifterSwift/pull/727) by [guykogus](https://github.com/guykogus).
 - **UIImage**:
   - Added `withBackgroundColor(_:)` to specify a background color for a partially transparent image. [#721](https://github.com/SwifterSwift/SwifterSwift/pull/721) by [MaxHaertwig](https://github.com/maxhaertwig).
+- **Sequence**:
+  - Added `sum(for:)` to sum up a Numeric property, referenced by KeyPath, of all elements in a sequence. [#736](https://github.com/SwifterSwift/SwifterSwift/pull/736) by [Moritz Sternemann](https://github.com/moritzsternemann).
```
```suggestion
  - Added `sum(for:)` to sum up an `AdditiveArithmetic` property, referenced by `KeyPath`, of all elements in a sequence. [#736](https://github.com/SwifterSwift/SwifterSwift/pull/736) by [Moritz Sternemann](https://github.com/moritzsternemann).
```

Sources/SwifterSwift/SwiftStdlib/SequenceExtensions.swift
chulbert: I'm generally a proponent of building on top of existing solutions.  SwifterSwift already implements `sum()` so I would utilize that:

`return map { $0[keyPath: keyPath] }.sum()`

guykogus: SwifterSwift is designed to be copyable piecemeal, so we can't reference our own functions between themselves. Otherwise, anybody wanting to copy a function manually may find themselves traversing one function after the next. It's a big drawback, but it makes the library more flexible.

moritzsternemann: I found the following rule in the contributing guidelines:
> - Do not use an existing SwifterSwift extension inside another SwifterSwift extension. All extensions should be able to be copied and pasted and work immediately without having to copy another extension.

Let me know if I should change the implementation to your proposed solution regardless :)

moritzsternemann:
Thanks for your feedback @guykogus, will update the PR in a minute!

Sources/SwifterSwift/SwiftStdlib/SequenceExtensions.swift
moritzsternemann: I think a type constraint is enough here, no need for a generic where clause. What do you think?

guykogus: They're actually the exact same, I just prefer always using `where` (up to you, personally). You can do it whichever way you like 😄

moritzsternemann: Okay, got it. All the other generic functions in the Sequence extension use the type constraint syntax, so I'd like to keep it for this one as well 😇

Sources/SwifterSwift/SwiftStdlib/SequenceExtensions.swift
chulbert: If the intention is each _function_ stand on its own then I agree.  Perhaps the guidelines could make that a little clearer (function vs extension).

That said, I might argue anyone copying implementations out of your library isn't _using your library_ so it's not really "more flexible."  ;)

guykogus: Yeah, I realised straight away that "flexible" was the wrong word, but you get me ;)

Sources/SwifterSwift/SwiftStdlib/SequenceExtensions.swift
```diff
@@ -202,6 +202,17 @@ public extension Sequence {
...
+
+    /// SwifterSwift: Sum of a `AdditiveArithmetic` property of each `Element` in a `Sequence`.
+    ///
+    ///     ["James", "Wade", "Bryant"].sum(for: \.count) -> 15
+    ///
+    /// - Parameter keyPath: Key path of the `AdditiveArithmetic` property.
+    /// - Returns: The sum of the `AdditiveArithmetic` propertys at `keyPath`.
+    func sum<T: AdditiveArithmetic>(for keyPath: KeyPath<Element, T>) -> T {
+        // Inspired by: https://swiftbysundell.com/articles/reducers-in-swift/
+        return reduce(.zero) { $0 + $1[keyPath: keyPath] }
```
LucianoPAlmeida:
```suggestion
        return reduce(into: .zero) { $0 += $1[keyPath: keyPath] }
```
Just a suggestion, what do you think?

guykogus: Same-same, no? What benefit is there to using the mutating version?

LucianoPAlmeida: In this case not much, is more useful when we are reducing sequeces, collections where avoid the copy is a performance boost. But here, copy or not doesn't really make a difference... anyway, just a suggestion

SD10:

PR #738, opened 2019-10-18
Add Thread Extensions

FraDeliro:
## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
   - Added `init(colors:locations:startPoint:endPoint:type:)` convenience initializer. [#726](https://github.com/SwifterSwift/SwifterSwift/pull/726) by [JayMehta97](https://github.com/JayMehta97).
 - **Sequence**:
   - Added `sum(for:)` to sum up an `AdditiveArithmetic` property, referenced by `KeyPath`, of all elements in a sequence. [#736](https://github.com/SwifterSwift/SwifterSwift/pull/736) by [Moritz Sternemann](https://github.com/moritzsternemann).
+- **Thread**:
+  - Added `mainThreadCompletion<T>(_: result:)`, which provides a method to execute a completion block on main thread. [#738](https://github.com/SwifterSwift/SwifterSwift/pull/738) by [Francesco Deliro](https://github.com/FraDeliro).
 
 ### Changed
 - **UIImage**:
--- a/Sources/SwifterSwift/Foundation/ThreadExtensions.swift
+++ b/Sources/SwifterSwift/Foundation/ThreadExtensions.swift
@@ -20,7 +20,7 @@ public extension Thread {
         _ completion: @escaping (Result<T, Error>) -> Void,
         result: Result<T, Error>
     ) {
-        if current == main {
+        if isMainThread {
             completion(result)
         } else {
             DispatchQueue.main.async {
--- a/Sources/SwifterSwift/Foundation/ThreadExtensions.swift
+++ b/Sources/SwifterSwift/Foundation/ThreadExtensions.swift
@@ -1,0 +1,34 @@
+//
+//  ThreadExtensions.swift
+//  SwifterSwift
+//
+//  Created by Francesco Deliro on 18/10/2019.
+//  Copyright © 2019 SwifterSwift
+//
+
+#if canImport(Foundation)
+import Foundation
+
+// MARK: - Methods
+public extension Thread {
+
+    /// SwifterSwift: Execute a completion block on the main thread.
+    ///
+    /// - Parameter completion: The completion to execute.
+    /// - Parameter result: The result of the completion.
+    static func mainThreadCompletion<T>(
+        _ completion: @escaping (Result<T, Error>) -> Void,
+        result: Result<T, Error>
+    ) {
+        if current == main {
+            completion(result)
+        } else {
+            DispatchQueue.main.async {
+                completion(result)
+            }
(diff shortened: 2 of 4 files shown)


omaralbeik:
Thank you for putting the time in this extension @FraDeliro, to be honest, I have some mixed feelings about it, why don't we just use `DispatchQueue.main.async { ... }` directly 🤔

Could you please share a use case where you think this extension is useful

chulbert:
I cannot recommend this approach.  A given callback should _always_ be synchronous or it should _always_ be asynchronous.  Relying on this construct is a bug waiting to happen.  Consider:

```
firstThing()
Thread.mainThreadCompletion { 
    secondThing()
}
thirdThing()
```

What is the order of operations?  It depends on the _calling thread_ and that'll drive you to madness.

FraDeliro:
> Thank you for putting the time in this extension @FraDeliro, to be honest, I have some mixed feelings about it, why don't we just use `DispatchQueue.main.async { ... }` directly 🤔
> 
> Could you please share a use case where you think this extension is useful

Hi @omaralbeik, thank you for your reply. I found this extension useful when I had to test asynchronous code. For example if we fetch data in the background we need to update the UI on main thread but this is not needed during tests if we use mock data.

guykogus:
@FraDeliro if your protocol marks the callbacks as `@escaping` then you shouldn't create a hack like this to make them non-escaping just for tests. It can also cause issues in other situations, as mentioned by @chulbert.

What you may instead want is something like this:

```Swift
extension XCTestCase {
    func perform<T>(timeout: TimeInterval = 1,
                    action: (@escaping (Result<T, Error>) -> Void) -> Void,
                    completion: @escaping (Result<T, Error>) -> Void) {
        let expectation = XCTestExpectation()
        action { result in
            completion(result)
            expectation.fulfill()
        }
        wait(for: [expectation], timeout: timeout)
    }
}
```

Then you can use it in your test, such as in this simple example:

```Swift
final class MyTests: XCTestCase {

    private struct StringLoader {
        let string: String

        func load(completion: @escaping (Result<String, Error>) -> Void) {
            DispatchQueue.main.asyncAfter(deadline: .now() + 0.5) {
                completion(.success(self.string))
            }
        }
    }

    func testStringResult() {
        let string = "Hello world"
        perform(action: StringLoader(string: string).load) { result in
            switch result {
            case .success(let text):
                XCTAssertEqual(text, string)
            case .failure(let error):
                XCTFail(error.localizedDescription)
            }
        }
    }

}
```

FraDeliro:
Thanks @guykogus, @chulbert. I understand what you mean.
I will try to show you how I used it with an axample, but if you think that this is something that can cause confusion I can close the PR or update it with the solution @guykogus proposed.

```Swift
protocol ApiClientType {
  func fetchImage(from url: String, completion: @escaping (Result<UIImage, Error>) -> Void)
}

protocol ViewModelType {
  func fetchImage(completion: @escaping (Result<UIImage, Error>) -> Void)
}

struct ViewModel: ViewModelType {
  
  let apiClient: ApiClientType
  
  init(apiClient: ApiClientType) {
    self.apiClient = apiClient
  }
  
  func fetchImage(completion: @escaping (Result<UIImage, Error>) -> Void) {
    apiClient.fetchImage(from: "viewModel.imageURL") { result in
      switch result {
      case .success(let image):
        Thread.mainThreadCompletion(completion, result: .success(image))
      case .failure(let error):
        Thread.mainThreadCompletion(completion, result: .failure(error))
      }
    }
  }

}
 
final class ExampleViewController: UIViewController {
  
  @IBOutlet var imageView: UIImageView!
  
  private let viewModel: ViewModelType
  
  init(viewModel: ViewModelType) {
    self.viewModel = viewModel
  }
  
  func fetchImage() {
    viewModel.fetchImage { [weak self] result in
      switch result {
      case .success(let image):
        self?.imageView.image = image
      case .failure(let error):
        print(error.localizedDescription)
      }
    }
  }
  
}

import XCTest

final class MockApiClient: ApiClientType {
  
  let mockImage = UIImage()
  
  func fetchImage(from url: String, completion: @escaping (Result<UIImage, Error>) -> Void) {
    completion(.success(mockImage))
  }
  
}

final class ExampleViewControllerTests: XCTextCase {
  
  func testFetchImage() {
    let mockApiClient = MockApiClient()
    let viewModel = ViewModel(apiClient: mockApiClient)
    let sut = ExampleViewController(viewModel: viewModel)
    sut.fetchImage()
    XCTAssertEqual(sut.imageView.image, mockApiClient.mockImage)
  }
  
}
```

guykogus:
This is the first I've seen of someone running unit tests on a view controller...
Anyway, why run the unit test on the view controller when you can do it on the view model? Then, using my suggestion you can test `fetchImage` in your test.
```Swift
perform(action: viewModel.fetchImage) { result in
   ...
}
```
Alternatively, you could ensure that you `ApiClientType` conformer always calls its completion block on the main thread. There's nothing I hate more than clients that don't do this, especially when they're returning a UI object like an image!

FraDeliro:
@guykogus This was just a quick example (my fault I haven't gone through all the steps) and it was focused to show how to use this extension in tests. I agree with you about testing the viewModel and that the completion coming from the ApiClient class should be executed on the main thread, this function called from the viewModel in a real project looks like:

```Swift
func fetchImage(completion: @escaping (Result<UIImage, Error>) -> Void) {
    apiClient.fetchImage(from: "viewModel.imageURL", completion: completion)
 }
```
Anyway this was out of the scope, and yes I worked in big projects where devs add unit test to view controllers :)
But in conclusion this extension seems not to respond to a need, I will close the PR ;)

PR #739, opened 2019-10-20
Add missing import and fix podspec

guykogus:
🚀
When I try to open the app in Xcode 11.1 it opens it as a Swift Package. There seems to be a few problems there as opposed to opening it via the `.xcworkspace` file. Not really sure if they're worth fixing, e.g. missing resources during tests, but I started fixing a few things and realised that the podspec was broken. I think it's time to re-instate `pod lib lint` with our tests.


guykogus:
Anybody know where the error is in travis? 😕

LucianoPAlmeida:
> Anybody know where the error is in travis? 😕


Kinda ... that's the reason we disable it back then :( 
Pod lib lint is taking too long, and since it doesn't give Travis an output while executing, it ends up timing out in 10 min. At least it is what I think is happening ...

guykogus:
@LucianoPAlmeida there were 2 problems:
1. With no output Travis times out after 10 mins
2. With output the log becomes too big and Travis aborts

I solved it by splitting the pod lib lint task into 4 separate Travis jobs, one for each platform 🎉

LucianoPAlmeida:
Nice @guykogus \o/

LucianoPAlmeida:
cc @omaralbeik

PR #749, opened 2019-10-30
adding performance docs for existing ArrayExtensions, and creating more efficient functions

jpsoultanis:
Added big-O documentation for existing ArrayExtensions, and created more efficient implementations for some functions by leveraging hashable elements and sets.

## Checklist

### DIFF
--- a/Sources/SwifterSwift/SwiftStdlib/ArrayExtensions.swift
+++ b/Sources/SwifterSwift/SwiftStdlib/ArrayExtensions.swift
@@ -9,7 +9,7 @@
 // MARK: - Methods
 public extension Array {
 
-    /// SwifterSwift: Insert an element at the beginning of array.
+    /// SwifterSwift: Insert an element at the beginning of array. Complexity: O(n)
     ///
     ///        [2, 3, 4, 5].prepend(1) -> [1, 2, 3, 4, 5]
     ///        ["e", "l", "l", "o"].prepend("h") -> ["h", "e", "l", "l", "o"]
@@ -19,7 +19,7 @@ public extension Array {
         insert(newElement, at: 0)
     }
 
-    /// SwifterSwift: Safely swap values at given index positions.
+    /// SwifterSwift: Safely swap values at given index positions. Complexity: O(1)
     ///
     ///        [1, 2, 3, 4, 5].safeSwap(from: 3, to: 0) -> [4, 2, 3, 1, 5]
     ///        ["h", "e", "l", "l", "o"].safeSwap(from: 1, to: 0) -> ["e", "h", "l", "l", "o"]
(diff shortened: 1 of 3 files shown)


Sources/SwifterSwift/SwiftStdlib/ArrayExtensions.swift
```diff
@@ -34,11 +34,62 @@ public extension Array {
...
+    /// SwifterSwift: Remove all duplicate elements from Array, maintaining order. Complexity: O(n)
+    ///
+    ///        [1, 2, 2, 3, 4, 5].removeDuplicates() -> [1, 2, 3, 4, 5]
+    ///        ["h", "e", "l", "l", "o"]. removeDuplicates() -> ["h", "e", "l", "o"]
+    ///
+    /// - Returns: Return array with all duplicate elements removed.
+    @discardableResult
+    mutating func removeDuplicates() -> [Element] {
+        var foundElements = Set<Element>()
+        self = reduce(into: [Element]()) {
```
LucianoPAlmeida: I think that could be better if we could use `removeAll(where: )` ... that way we avoid assign to self and creating a new array... taking advantage on the already performant partion based in-place mutation. 
 what do you think?

Sources/SwifterSwift/SwiftStdlib/ArrayExtensions.swift
```diff
@@ -34,11 +34,62 @@ public extension Array {
         swapAt(index, otherIndex)
     }
 }
+// MARK: - Methods (Hashable)
+public extension Array where Element: Hashable {
+    /// SwifterSwift: Remove all instances contained in items parameter from array. Complexity: O(n)
```
LucianoPAlmeida: Nit: Instead of adding complexity here, we could add it as the key so it would be highlighted on the docs generated.
Example: 
```
/// - Parameter items: unique items to remove.
/// - Returns: self after removing all instances of all items in given array.  
/// - Complexity: O(1)

```

Sources/SwifterSwift/SwiftStdlib/ArrayExtensions.swift
```diff
@@ -34,11 +34,62 @@ public extension Array {
...
+    ///
+    ///        [1, 2, 2, 3, 4, 5].removeDuplicates() -> [1, 2, 3, 4, 5]
+    ///        ["h", "e", "l", "l", "o"]. removeDuplicates() -> ["h", "e", "l", "o"]
+    ///
+    /// - Returns: Return array with all duplicate elements removed.
+    @discardableResult
+    mutating func removeDuplicates() -> [Element] {
+        var foundElements = Set<Element>()
+        self = reduce(into: [Element]()) {
+            if !foundElements.contains($1) {
```
guykogus: You can combine these 2 lines as:
```Swift
if !foundElements.insert($1).inserted {
```

Sources/SwifterSwift/SwiftStdlib/ArrayExtensions.swift
```diff
@@ -34,11 +34,62 @@ public extension Array {
...
+    ///
+    ///     [1, 1, 2, 2, 3, 3, 3, 4, 5].withoutDuplicates() -> [1, 2, 3, 4, 5])
+    ///     ["h", "e", "l", "l", "o"].withoutDuplicates() -> ["h", "e", "l", "o"])
+    ///
+    /// - Returns: an array of unique elements.
+    ///
+    func withoutDuplicates() -> [Element] {
+        var foundElements = Set<Element>()
+        return reduce(into: [Element]()) {
+            if !foundElements.contains($1) {
```
guykogus: Same comment as above regarding making it 1 line

Sources/SwifterSwift/SwiftStdlib/ArrayExtensions.swift
chulbert: How would you remove "all but the first"?

guykogus: We already have a similar implementation in `RangeReplaceableCollectionExtensions.swift`.
```Swift
mutating func removeDuplicates() {
    var set = Set<Element>()
    removeAll { !set.insert($0).inserted }
}
```

Issue #743, opened 2019-10-21
Add KeyedDecodingContainer extensions


omaralbeik:
Add the following extensions to `KeyedDecodingContainer`

```swift
public extension KeyedDecodingContainer where Key: CodingKey {
    /// SwifterSwift: Try to decode a Bool as Int then String before decoding as Bool.
    ///
    /// - Parameter key: Key.
    /// - Returns: Decoded Bool value.
    /// - Throws: Decoding error.
    public func decodeBoolAsIntOrString(forKey key: K) throws -> Bool { ... }

    /// SwifterSwift: Try to decode a Bool as Int then String before decoding as Bool if present.
    ///
    /// - Parameter key: Key.
    /// - Returns: Decoded Bool value.
    /// - Throws: Decoding error.
    public func decodeBoolAsIntOrStringIfPresent(forKey key: K) throws -> Bool? { ... }
}
```

FraDeliro:
I can take this one ;)

omaralbeik:
@FraDeliro Go for it, I assigned it to you 🚀

FraDeliro:
Thanks @omaralbeik  👍

PR #750, opened 2019-10-30
Add KeyedDecodingContainer extensions

FraDeliro:
This PR adds KeyedDecodingContainer extensions as required by this issue #743 

## Checklist


Sources/SwifterSwift/SwiftStdlib/KeyedDecodingContainer.swift
```diff
@@ -0,0 +1,47 @@
...
+    /// - Returns: Decoded Bool value.
+    /// - Throws: Decoding error.
+    func decodeBoolAsIntOrString(forKey key: K) throws -> Bool {
+        if let intValue = try? decode(Int.self, forKey: key) {
+            return (intValue as NSNumber).boolValue
+        } else if let stringValue = try? decode(String.self, forKey: key) {
+            return (stringValue as NSString).boolValue
+        } else {
+            do {
+                return try decode(Bool.self, forKey: key)
```
guykogus: You can just call this without the `do-catch`.

Sources/SwifterSwift/SwiftStdlib/KeyedDecodingContainer.swift
```diff
@@ -0,0 +1,47 @@
...
+    /// - Returns: Decoded Bool value.
+    /// - Throws: Decoding error.
+    func decodeBoolAsIntOrStringIfPresent(forKey key: K) throws -> Bool? {
+        if let intValue = try? decodeIfPresent(Int.self, forKey: key) {
+            return (intValue as NSNumber).boolValue
+        } else if let stringValue = try? decodeIfPresent(String.self, forKey: key) {
+            return (stringValue as NSString).boolValue
+        } else {
+            do {
+                return try decodeIfPresent(Bool.self, forKey: key)
```
guykogus: As above, no need for `do-catch`.

Tests/SwiftStdlibTests/KeyedDecodingContainerTests.swift
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        isPlaying = try container.decodeBoolAsIntOrString(forKey: .isPlaying)
+        isFullScreen = try container.decodeBoolAsIntOrStringIfPresent(forKey: .isFullScreen)
+    }
+}
+
+private enum ElementToTest: Decodable {
```
guykogus: You can remove this and simplify all the test strings you have by getting rid of the `data` root object, instead just use the internal value.

Tests/SwiftStdlibTests/KeyedDecodingContainerTests.swift
```diff
@@ -0,0 +1,139 @@
...
+    {
+      "data" : {
+        "isPlaying" : [0],
+        "isFullScreen" : ["false"]
+      }
+    }
+    """.utf8)
+
+    func testDecodeBoolAsIntOrStringDataAsIntSuccessful() {
+        let result = try? JSONDecoder().decode(ElementToTest.self, from: isPlayingAndIsFullScreenAsInt)
```
guykogus: Extract this to a private function and change all the test `Data` objects above to string literals so that you're not initialising each one of them individually using their `utf8` representations.

guykogus:
I just noticed a major issue with your implementation: It requires `Foundation` because you're casting the values to `NSNumber` and `NSString`. You either need an alternative or cover the whole thing with `#if canImport(Foundation)`.

Sources/SwifterSwift/SwiftStdlib/KeyedDecodingContainer.swift
```diff
@@ -0,0 +1,43 @@
+//
+//  KeyedDecodingContainer.swift
+//  SwifterSwift
+//
+//  Created by Francesco Deliro on 23/10/2019.
+//  Copyright © 2019 SwifterSwift
+//
+
+#if canImport(Foundation)
```
guykogus: This needs to be placed around the functions themselves, since they're what will actually give you compile errors when `Foundation` isn't available.

Sources/SwifterSwift/SwiftStdlib/KeyedDecodingContainer.swift
```diff
@@ -0,0 +1,43 @@
...
+import Foundation
+#endif
+
+public extension KeyedDecodingContainer where Key: CodingKey {
```
guykogus: Add an empty line below

Sources/SwifterSwift/SwiftStdlib/KeyedDecodingContainer.swift
```diff
@@ -0,0 +1,43 @@
...
+    func decodeBoolAsIntOrStringIfPresent(forKey key: K) throws -> Bool? {
+        if let intValue = try? decodeIfPresent(Int.self, forKey: key) {
+            return (intValue as NSNumber).boolValue
+        } else if let stringValue = try? decodeIfPresent(String.self, forKey: key) {
+            return (stringValue as NSString).boolValue
+        } else {
+            return try decodeIfPresent(Bool.self, forKey: key)
+        }
+    }
+}
```
guykogus: Add empty line above

Tests/SwiftStdlibTests/KeyedDecodingContainerTests.swift
```diff
@@ -0,0 +1,97 @@
...
+
+import XCTest
+
+@testable import SwifterSwift
+
+private struct Video: Decodable {
+    var isPlaying: Bool
+    var isFullScreen: Bool?
+
+    enum CodingKeys: String, CodingKey {
```
guykogus: Does it still work if you delete this definition? I think it's auto-generated by the compiler.

Tests/SwiftStdlibTests/KeyedDecodingContainerTests.swift
```diff
@@ -0,0 +1,97 @@
...
+//  Created by Francesco Deliro on 23/10/2019.
+//  Copyright © 2019 SwifterSwift
+//
+
+import XCTest
+
+@testable import SwifterSwift
+
+private struct Video: Decodable {
+    var isPlaying: Bool
```
guykogus:
```suggestion
    let isPlaying: Bool
```

Tests/SwiftStdlibTests/KeyedDecodingContainerTests.swift
```diff
@@ -0,0 +1,97 @@
...
+    var isFullScreen: Bool?
```
guykogus:
```suggestion
    let isFullScreen: Bool?
```

Tests/SwiftStdlibTests/KeyedDecodingContainerTests.swift
```diff
@@ -0,0 +1,97 @@
...
+    init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        isPlaying = try container.decodeBoolAsIntOrString(forKey: .isPlaying)
+        isFullScreen = try container.decodeBoolAsIntOrStringIfPresent(forKey: .isFullScreen)
+    }
+}
+
+final class KeyedDecodingContainerTests: XCTestCase {
+
+    private let isPlayingAndIsFullScreenAsInt = #"{"isPlaying": 1, "isFullScreen": 0}"#
```
guykogus: Put these in their respective functions

Tests/SwiftStdlibTests/KeyedDecodingContainerTests.swift
```diff
@@ -0,0 +1,97 @@
...
+    private let isPlayingAndIsFullScreenAsString = #"{"isPlaying": "true", "isFullScreen": "false"}"#
+    private let isPlayingAndIsFullScreenAsBool = #"{"isPlaying": true, "isFullScreen": false}"#
+    private let isPlayingAsIntAndIsFullScreenNotPresent = #"{"isPlaying": 0}"#
+    private let isPlayingAndIsFullScreenTypeMismatch = #"{"isPlaying": [1], "isFullScreen": ["true"]}"#
+
+    func testDecodeBoolAsIntOrStringDataAsIntSuccessful() {
+        let data = mockJsonData(from: isPlayingAndIsFullScreenAsInt)
+        let result = try? JSONDecoder().decode(Video.self, from: data)
```
guykogus:
```suggestion
        let result = try! JSONDecoder().decode(Video.self, from: data) // swiftlint:disable:this force_try
```
Not sure if the `swiftlint` call is necessary (check without), but you know that this is going to work, so save yourself the extra code for all these functions.

Tests/SwiftStdlibTests/KeyedDecodingContainerTests.swift
```diff
@@ -0,0 +1,97 @@
...
+
+        if let video = result {
+            XCTAssertTrue(video.isPlaying)
+            XCTAssertNotNil(video.isFullScreen)
+            XCTAssertFalse(video.isFullScreen!)
```
guykogus: You don't need to assert that it's not `nil` if you just use:
```suggestion
            XCTAssertEqual(video.isFullScreen, false)
```
This way you can simplify all your tests.

Tests/SwiftStdlibTests/KeyedDecodingContainerTests.swift
```diff
@@ -0,0 +1,73 @@
...
+
+import XCTest
+
+@testable import SwifterSwift
+
+private struct Video: Decodable {
+    let isPlaying: Bool
+    let isFullScreen: Bool?
+
+    enum CodingKeys: CodingKey {
```
guykogus: So can you get rid of this `enum`? Does it still work without it?

FraDeliro: Hi @guykogus , no I cannot remove the enum.

guykogus: Make it `private` at least, then :)

FraDeliro: Updated :)

Sources/SwifterSwift/SwiftStdlib/KeyedDecodingContainer.swift
```diff
@@ -0,0 +1,49 @@
...
+//
+//  Created by Francesco Deliro on 23/10/2019.
+//  Copyright © 2019 SwifterSwift
+//
+
+#if canImport(Foundation)
+import Foundation
+#endif
+
+public extension KeyedDecodingContainer where Key: CodingKey {
```
guykogus: `Key` already conforms to `CodingKey`.
```suggestion
public extension KeyedDecodingContainer {
```

Sources/SwifterSwift/SwiftStdlib/KeyedDecodingContainer.swift
```diff
@@ -0,0 +1,49 @@
...
+
+    #if canImport(Foundation)
+    /// SwifterSwift: Try to decode a Bool as Int then String before decoding as Bool.
+    ///
+    /// - Parameter key: Key.
+    /// - Returns: Decoded Bool value.
+    /// - Throws: Decoding error.
+    func decodeBoolAsIntOrString(forKey key: K) throws -> Bool {
```
guykogus:
```suggestion
    func decodeBoolAsIntOrString(forKey key: Key) throws -> Bool {
```

Sources/SwifterSwift/SwiftStdlib/KeyedDecodingContainer.swift
```diff
@@ -0,0 +1,49 @@
...
+    }
+    #endif
+
+    #if canImport(Foundation)
+    /// SwifterSwift: Try to decode a Bool as Int then String before decoding as Bool if present.
+    ///
+    /// - Parameter key: Key.
+    /// - Returns: Decoded Bool value.
+    /// - Throws: Decoding error.
+    func decodeBoolAsIntOrStringIfPresent(forKey key: K) throws -> Bool? {
```
guykogus:
```suggestion
    func decodeBoolAsIntOrStringIfPresent(forKey key: Key) throws -> Bool? {
```

LucianoPAlmeida:
@guykogus I'll take a detailed look as soon as possible, but for this one specifically, I would like to have the review from @omaralbeik :))

Sources/SwifterSwift/SwiftStdlib/KeyedDecodingContainer.swift
```diff
@@ -0,0 +1,49 @@
+//
```
omaralbeik: Please rename the file to `KeyedDecodingContainerExtensions.swift `

Sources/SwifterSwift/SwiftStdlib/KeyedDecodingContainer.swift
```diff
@@ -0,0 +1,49 @@
...
+
+    #if canImport(Foundation)
+    /// SwifterSwift: Try to decode a Bool as Int then String before decoding as Bool if present.
+    ///
+    /// - Parameter key: Key.
+    /// - Returns: Decoded Bool value.
+    /// - Throws: Decoding error.
+    func decodeBoolAsIntOrStringIfPresent(forKey key: Key) throws -> Bool? {
+        if let intValue = try? decodeIfPresent(Int.self, forKey: key) {
+            return (intValue as NSNumber).boolValue
```
omaralbeik: Please replace this with return `return intValue == 1`

Sources/SwifterSwift/SwiftStdlib/KeyedDecodingContainer.swift
```diff
@@ -0,0 +1,49 @@
...
+        } else if let stringValue = try? decodeIfPresent(String.self, forKey: key) {
+            return (stringValue as NSString).boolValue
```
omaralbeik: Please replace this with `return stringValue == "1"`

Tests/SwiftStdlibTests/KeyedDecodingContainerTests.swift
```diff
@@ -0,0 +1,73 @@
+//
```
omaralbeik: Please rename the file to `KeyedDecodingContainerExtensionsTests.swift`

Sources/SwifterSwift/SwiftStdlib/KeyedDecodingContainer.swift
```diff
@@ -0,0 +1,49 @@
...
+//  KeyedDecodingContainer.swift
+//  SwifterSwift
+//
+//  Created by Francesco Deliro on 23/10/2019.
+//  Copyright © 2019 SwifterSwift
+//
+
+#if canImport(Foundation)
+import Foundation
+#endif
```
omaralbeik: `KeyedDecodingContainer` is defined in `Foundation`, please wrap the entire file with `#if canImport(Foundation)`

> If you do so, no need to wrap each function separately 

```swift
#if canImport(Foundation)
import Foundation

public extension KeyedDecodingContainer {
   ...
}
#endif
```

guykogus: `KeyedDecodingContainer` is part of the Swift stdlib. The reason that `Foundation` is being used in these new functions is in order to use `NSString` and `NSInteger`'s `.boolValue` properties, which look for one of several predefined values meaning `true`.

omaralbeik: Ohh, sorry my bad 😅

Sources/SwifterSwift/SwiftStdlib/KeyedDecodingContainer.swift
FraDeliro: As commented by @guykogus I preferred to use NSString and NSInteger's .boolValue properties, which look for one of several predefined values meaning true.

In this case quoting Apple Doc we have :

**NSNumber boolValue**
> A 0 value always means false, and any nonzero value is interpreted as true.

Here we can easily move to a swifty solution, but for the string I think it would be preferable to use boolValue, what do you think about? :)

Sources/SwifterSwift/SwiftStdlib/KeyedDecodingContainer.swift
FraDeliro: Quoting Apple doc:

**NSString boolValue**
> This property is true on encountering one of "Y", "y", "T", "t", or a digit 1-9—the method ignores any trailing characters. This property is false if the receiver doesn’t begin with a valid decimal text representation of a number.

> The property assumes a decimal representation and skips whitespace at the beginning of the string. It also skips initial whitespace characters, or optional -/+ sign followed by zeroes.

Sources/SwifterSwift/SwiftStdlib/KeyedDecodingContainer.swift
FraDeliro: Updated 👍

Tests/SwiftStdlibTests/KeyedDecodingContainerTests.swift
FraDeliro: Updated 👍

PR #751, opened 2019-10-30
Add `init(grouping:by:)` for Dictionary that uses a keypath. 🚀

mmdock:
## Checklist

- [X] New extensions are written in Swift 5.0.
- [X] New extensions support iOS 8.0+ / tvOS 9.0+ / macOS 10.10+ / watchOS 2.0+, or use `@available` if not.

### DIFF
--- a/Sources/SwifterSwift/SwiftStdlib/DictionaryExtensions.swift
+++ b/Sources/SwifterSwift/SwiftStdlib/DictionaryExtensions.swift
@@ -13,6 +13,15 @@ import Foundation
 // MARK: - Methods
 public extension Dictionary {
 
+    /// SwifterSwift: Creates a Dictionary from a given sequence grouped by a given Keypath
+    ///
+    /// - Parameters:
+    ///   - grouping: Sequence being grouped
+    ///   - by: The Keypath to group by.
+    init<S: Sequence>(grouping sequence: S, by keyPath: KeyPath<S.Element, Key>) where Value == [S.Element] {
+      self.init(grouping: sequence, by: { $0[keyPath: keyPath] })
+    }
+
     /// SwifterSwift: Check if key exists in dictionary.
     ///
     ///        let dict: [String: Any] = ["testKey": "testValue", "testArrayKey": [1, 2, 3, 4, 5]]
--- a/Sources/SwifterSwift/SwiftStdlib/DictionaryExtensions.swift
+++ b/Sources/SwifterSwift/SwiftStdlib/DictionaryExtensions.swift
@@ -16,10 +16,10 @@ public extension Dictionary {
     /// SwifterSwift: Creates a Dictionary from a given sequence grouped by a given Keypath
     ///
     /// - Parameters:
-    ///   - grouping: Sequence being grouped
-    ///   - by: The Keypath to group by.
+    ///   - sequence: Sequence being grouped
+    ///   - keypath: The Keypath to group by.
     init<S: Sequence>(grouping sequence: S, by keyPath: KeyPath<S.Element, Key>) where Value == [S.Element] {
-      self.init(grouping: sequence, by: { $0[keyPath: keyPath] })
+       self.init(grouping: sequence, by: { $0[keyPath: keyPath] })
     }
 
     /// SwifterSwift: Check if key exists in dictionary.
--- a/Sources/SwifterSwift/SwiftStdlib/DictionaryExtensions.swift
+++ b/Sources/SwifterSwift/SwiftStdlib/DictionaryExtensions.swift
@@ -13,11 +13,11 @@ import Foundation
 // MARK: - Methods
 public extension Dictionary {
 
-    /// SwifterSwift: Creates a Dictionary from a given sequence grouped by a given Keypath
+    /// SwifterSwift: Creates a Dictionary from a given sequence grouped by a given key path.
     ///
     /// - Parameters:
     ///   - sequence: Sequence being grouped
-    ///   - keypath: The Keypath to group by.
+    ///   - keypath: The key path to group by.
     init<S: Sequence>(grouping sequence: S, by keyPath: KeyPath<S.Element, Key>) where Value == [S.Element] {
        self.init(grouping: sequence, by: { $0[keyPath: keyPath] })
     }
--- a/Sources/SwifterSwift/SwiftStdlib/SequenceExtensions.swift
+++ b/Sources/SwifterSwift/SwiftStdlib/SequenceExtensions.swift
@@ -186,15 +186,6 @@ public extension Sequence {
         return (Array(matching), Array(nonMatching))
     }
 
-    /// SwifterSwift: Return a map grouped on a keypath value.
-    ///
-    /// - Parameter by: Key path to group by.  Must be Hashable.
-    /// - Returns: Dictionary of arrays using the keypath value as the key.
-    func grouped<T: Hashable>(by keyPath: KeyPath<Element, T>) -> [T: [Element]] {
-        // https://twitter.com/johnsundell/status/1106886417756704768?lang=en
-        return Dictionary(grouping: self, by: { $0[keyPath: keyPath] })
-    }
-
     /// SwifterSwift: Return a sorted array  based on a keypath and a compare function.
     ///
     /// - Parameter path: Key path to sort.
--- a/Sources/SwifterSwift/SwiftStdlib/SequenceExtensions.swift
+++ b/Sources/SwifterSwift/SwiftStdlib/SequenceExtensions.swift
@@ -185,17 +185,26 @@ public extension Sequence {
         }
         return (Array(matching), Array(nonMatching))
     }
+    
+    /// SwifterSwift: Return a map grouped on a keypath value.
+    ///
+    /// - Parameter by: Key path to group by.  Must be Hashable.
+    /// - Returns: Dictionary of arrays using the keypath value as the key.
+    func grouped<T: Hashable>(by keyPath: KeyPath<Element, T>) -> [T: [Element]] {
+        // https://twitter.com/johnsundell/status/1106886417756704768?lang=en
+        return .init(grouping: self, by: { $0[keyPath: keyPath] })
+    }
 
     /// SwifterSwift: Return a sorted array  based on a keypath and a compare function.
     ///
-    /// - Parameter path: Key path to sort. The key path type must be Comparable.
+    /// - Parameter path: Key path to sort.
     /// - Parameter compare: Comparation function that will determine the ordering.
     /// - Returns: The sorted array.
     func sorted<T>(by keyPath: KeyPath<Element, T>, with compare: (T, T) -> Bool) -> [Element] {
         return sorted { compare($0[keyPath: keyPath], $1[keyPath: keyPath]) }
     }
 
-    /// SwifterSwift: Return a sorted array  based on a keypath and a compare function.
+    /// SwifterSwift: Return a sorted array  based on a keypath.
     ///
     /// - Parameter path: Key path to sort. The key path type must be Comparable.
     /// - Returns: The sorted array.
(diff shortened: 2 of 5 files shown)


Sources/SwifterSwift/SwiftStdlib/SequenceExtensions.swift
```diff
@@ -185,17 +185,26 @@ public extension Sequence {
         }
         return (Array(matching), Array(nonMatching))
     }
+    
+    /// SwifterSwift: Return a map grouped on a keypath value.
+    ///
+    /// - Parameter by: Key path to group by.  Must be Hashable.
+    /// - Returns: Dictionary of arrays using the keypath value as the key.
+    func grouped<T: Hashable>(by keyPath: KeyPath<Element, T>) -> [T: [Element]] {
```
LucianoPAlmeida: @mmdock I think here we could instead of having `Sequence`, maintain the pattern of initialization and have an dictionary extension like `init<S: Sequence, T: Hashable>(grouping sequence: S, by keyPath: KeyPath<Element, T>)`... 
What you think?

Sources/SwifterSwift/SwiftStdlib/SequenceExtensions.swift
```diff
@@ -185,17 +185,26 @@ public extension Sequence {
...
+        // https://twitter.com/johnsundell/status/1106886417756704768?lang=en
+        return .init(grouping: self, by: { $0[keyPath: keyPath] })
+    }
 
     /// SwifterSwift: Return a sorted array  based on a keypath and a compare function.
     ///
-    /// - Parameter path: Key path to sort. The key path type must be Comparable.
+    /// - Parameter path: Key path to sort.
```
LucianoPAlmeida: Nice, Thank's :)

Tests/SwiftStdlibTests/SequenceExtensionsTests.swift
```diff
@@ -151,5 +151,10 @@ final class SequenceExtensionsTests: XCTestCase {
         let array2 = ["James", "Wade", "Bryant", ""]
         XCTAssertEqual(array2.sorted(by: \String.first, with: optionalCompare), ["Bryant", "James", "Wade", ""])
     }
+    
+    func testKeyPathGrouped() {
+        let array = ["James", "Wade", "Bryant", "John"]
```
LucianoPAlmeida: Could you add an empty sequence case?

mmdock: @LucianoPAlmeida added a few extra cases actually, please take a look.

I am a little unhappy with how optional keypaths get handled, but can't think of a clean solution of the top of my head.  I will probably mess with it later, unless you got something obvious / easy i am missing or forgetting.

Sources/SwifterSwift/SwiftStdlib/SequenceExtensions.swift
mmdock: @LucianoPAlmeida I think both are valid independent extensions. 

I think `Dictionary(grouping: someSequence, by:  \String.count)` can be not as nice to read as `someSequence.grouped(by: \String.count)`, but that is my preferred way of writing swift code, which I am sure many would disagree with, so i will defer to your judgement on this.

As a note: I just tried (briefly, over my lunch break) 
`    init<S: Sequence, T: Hashable>(grouping sequence: S, by keyPath: KeyPath<Element, T>){
        self.init(grouping: sequence, by: { $0[keyPath: keyPath] })
    }`

but am receiving a `Type of expression is ambiguous without more context` on the `by: { $0[keyPath: keyPath] }`  segment.  I am not sure why this would work in the Sequence extension but not in this init 🤔.  Will think about this more when i have time this evening.

LucianoPAlmeida: > @LucianoPAlmeida I think both are valid independent extensions.
> 
> I think `Dictionary(grouping: someSequence, by: \String.count)` can be not as nice to read as `someSequence.grouped(by: \String.count)`, but that is my preferred way of writing swift code, which I am sure many would disagree with, so I will defer to your judgement on this.
> 

yeah, I do prefer this too ... but I think the dictionary one is preferable because is closer to the native one(the one used on the implementation). I mean if I'm a user who knows the swift stdlib has a dictionary init to do this I'll probably look for something key path based that follows the same structure. To have a native way to group the value as a Dictionary init and an extension that do the same but keyPath based as a Sequence extension seems a little inconsistent to me. So my vote is for the dictionary one 
let's see what the others think cc @omaralbeik @guykogus @chulbert 

> As a note: I just tried (briefly, over my lunch break)
> ` init<S: Sequence, T: Hashable>(grouping sequence: S, by keyPath: KeyPath<Element, T>){ self.init(grouping: sequence, by: { $0[keyPath: keyPath] }) }`
> 
> but am receiving a `Type of expression is ambiguous without more context` on the `by: { $0[keyPath: keyPath] }` segment. I am not sure why this would work in the Sequence extension but not in this init 🤔. Will think about this more when i have time this evening. If you have any obvious things to mention i can 🤦‍♂ over later, please comment it

Ohh sorry, I just typed it as a sketch... didn't know if it would be compilable...
You would have some constraints on generic types and T is actually not needed
```swift
  init<S: Sequence>(grouping sequence: S,
          by keyPath: KeyPath<S.Element, Key>) where Value == [S.Element] {
    self.init(grouping: sequence, by: { $0[keyPath: keyPath] }) 
  }
```

Tests/SwiftStdlibTests/SequenceExtensionsTests.swift
LucianoPAlmeida: I don't see a problem with optionals here... to me seems the expected behavior :))

Sources/SwifterSwift/SwiftStdlib/SequenceExtensions.swift
mmdock: @LucianoPAlmeida it took me way longer than I care to admit to understand what you were doing with that init.  I, for some reason, did not know `where` clauses can be used as part of an `init`.  I feel that that should be basic knowledge, but nope! Not me.

Implemented that and duplicated the tests.   Left in the sequence extension until we know which we prefer (or want to keep both?)

LucianoPAlmeida: Sure, let's see what the others think :) 
cc @omaralbeik @guykogus @chulbert

Sources/SwifterSwift/SwiftStdlib/DictionaryExtensions.swift
```diff
@@ -13,6 +13,15 @@ import Foundation
 // MARK: - Methods
 public extension Dictionary {
 
+    /// SwifterSwift: Creates a Dictionary from a given sequence grouped by a given Keypath
+    ///
+    /// - Parameters:
+    ///   - grouping: Sequence being grouped
+    ///   - by: The Keypath to group by.
+    init<S: Sequence>(grouping sequence: S, by keyPath: KeyPath<S.Element, Key>) where Value == [S.Element] {
+      self.init(grouping: sequence, by: { $0[keyPath: keyPath] })
```
guykogus:
```suggestion
       self.init(grouping: sequence, by: { $0[keyPath: keyPath] })
```

Sources/SwifterSwift/SwiftStdlib/DictionaryExtensions.swift
```diff
@@ -13,6 +13,15 @@ import Foundation
 // MARK: - Methods
 public extension Dictionary {
 
+    /// SwifterSwift: Creates a Dictionary from a given sequence grouped by a given Keypath
+    ///
+    /// - Parameters:
+    ///   - grouping: Sequence being grouped
+    ///   - by: The Keypath to group by.
```
guykogus:
```suggestion
    ///   - keypath: The Keypath to group by.
```

Sources/SwifterSwift/SwiftStdlib/DictionaryExtensions.swift
```diff
@@ -13,6 +13,15 @@ import Foundation
 // MARK: - Methods
 public extension Dictionary {
 
+    /// SwifterSwift: Creates a Dictionary from a given sequence grouped by a given Keypath
+    ///
+    /// - Parameters:
+    ///   - grouping: Sequence being grouped
```
guykogus:
```suggestion
    ///   - sequence: Sequence being grouped
```

Sources/SwifterSwift/SwiftStdlib/SequenceExtensions.swift
```diff
@@ -186,16 +186,25 @@ public extension Sequence {
         return (Array(matching), Array(nonMatching))
     }
 
+    /// SwifterSwift: Return a map grouped on a keypath value.
+    ///
+    /// - Parameter by: Key path to group by.  Must be Hashable.
+    /// - Returns: Dictionary of arrays using the keypath value as the key.
+    func grouped<T: Hashable>(by keyPath: KeyPath<Element, T>) -> [T: [Element]] {
+        // https://twitter.com/johnsundell/status/1106886417756704768?lang=en
+        return .init(grouping: self, by: { $0[keyPath: keyPath] })
```
guykogus: I used to like `.init(...)`, but not anymore...
```suggestion
        return Dictionary(grouping: self, by: { $0[keyPath: keyPath] })
```

Sources/SwifterSwift/SwiftStdlib/SequenceExtensions.swift
guykogus: I like them both, actually. If I had to choose I'd go for the `Dictionary` extension, which as @LucianoPAlmeida says is closer to the native initialiser.

mmdock: @guykogus @LucianoPAlmeida how about we keep both, so developers can use whichever fits their development style?  Rather than deciding for anyone wanting to use this functionality in the future.

LucianoPAlmeida: I'm not sure, about that ... because 
* We are implementing the same functionality twice
* Will be more work to maintain two than one, and we have to make sure the behavior is consistent
* I prefer to have only one and let the user make a wrapper in his project if it likes other style better. 

Those are the reasons I don't like the idea, but the first one is the deal breaker :) 
But ... let's see cc @omaralbeik @guykogus

guykogus: Agreed, let's not have this twice.

mmdock:
Removed the sequence extension `grouped(by:)` in latest commit.  Removed change log entry for that, and removed tests.  This commit is now in favor of the new init for Dictionary.  PR name will be edited to reflect that

Sources/SwifterSwift/SwiftStdlib/DictionaryExtensions.swift
```diff
@@ -13,6 +13,15 @@ import Foundation
 // MARK: - Methods
 public extension Dictionary {
 
+    /// SwifterSwift: Creates a Dictionary from a given sequence grouped by a given Keypath
```
LucianoPAlmeida:
```suggestion
    /// SwifterSwift: Creates a Dictionary from a given sequence grouped by a given key path.
```

Sources/SwifterSwift/SwiftStdlib/DictionaryExtensions.swift
```diff
@@ -13,6 +13,15 @@ import Foundation
...
+    ///
+    /// - Parameters:
+    ///   - sequence: Sequence being grouped
+    ///   - keypath: The Keypath to group by.
```
LucianoPAlmeida:
```suggestion
    ///   - keypath: The key path to group by.
```

mmdock:
@LucianoPAlmeida Not exactly sure how to make send of that TravisCI log tbh

LucianoPAlmeida:
Just trying to rebuild, let's see :)

LucianoPAlmeida:
Thank's @mmdock :))

PR #752, opened 2019-11-01
Xcode 11.2

guykogus:
🚀
Fix imports and tests for Xcode 11.2 and macCatalyst. I fixed some of the tests for macCatalyst but there are a few with incredibly strange results. At the very least this can now be used and tested with Xcode 11.2.

## Checklist

### DIFF
--- a/Sources/SwifterSwift/Foundation/DateExtensions.swift
+++ b/Sources/SwifterSwift/Foundation/DateExtensions.swift
@@ -257,7 +257,11 @@ public extension Date {
             return calendar.component(.nanosecond, from: self)
         }
         set {
+            #if targetEnvironment(macCatalyst)
+            let allowedRange = 0..<1000000000
+            #else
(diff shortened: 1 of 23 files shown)


Sources/SwifterSwift/Foundation/DateExtensions.swift
```diff
@@ -257,7 +257,11 @@ public extension Date {
             return calendar.component(.nanosecond, from: self)
         }
         set {
+            #if targetEnvironment(macCatalyst)
+            let allowedRange = 0..<1000000000
```
LucianoPAlmeida: We should probably add a comment here for why this is required. And also why is that value ... is there a way to get this max range dynamically?

guykogus: Comments added. The proper way would be to use the `Calendar` method, but for some reason it returns `nil` in `macCatalyst` as opposed to the proper range. That a nanosecond is 1,000,000,000th of a second I think is safe to hard-code, no?

PR #756, opened 2019-11-11
Remove the need for UIView Enums 🚀

mmdock:
## Checklist

Further Details;

Removed enums `ShakeDirection`, `AngleUnit`, `ShakeAnimationType` and their dependent functions (or more correctly, re-wrote them)

One of the goals of SwifterSwift is to allow for anyone to just copy what they need without needing to copy anything else.  While these enums do allow for a cleaner code read, I feel they take away from the spirit of the project.

Note:  I did, however, remove those enums and dependent functions, instead of maybe more gently deprecating them for now for anyone who may be using the SwifterSwift pod directly in their project.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,14 +25,16 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
   - Added `init(colors:locations:startPoint:endPoint:type:)` convenience initializer. [#726](https://github.com/SwifterSwift/SwifterSwift/pull/726) by [JayMehta97](https://github.com/JayMehta97).
 - **Sequence**:
   - Added `sum(for:)` to sum up an `AdditiveArithmetic` property, referenced by `KeyPath`, of all elements in a sequence. [#736](https://github.com/SwifterSwift/SwifterSwift/pull/736) by [Moritz Sternemann](https://github.com/moritzsternemann).
+- **UIView**
+  - Added `verticalShake(duration:animationType:completion:)` and `horizontalShake(duration:animationType:completion:)` functions to UIView extensions [#756](https://github.com/SwifterSwift/SwifterSwift/pull/756) by [mmdock](https://github.com/mmdock)
+  - Added `rotate(toRadian:animated:duration:completion)`, and `rotate(toDegree:animated:duration:completion)` functions to UIView extensions [#756](https://github.com/SwifterSwift/SwifterSwift/pull/756) by [mmdock](https://github.com/mmdock)
+  - Added `rotate(byRadian:animated:duration:completion)`, and `rotate(byDegree:animated:duration:completion)` functions to UIView extensions [#756](https://github.com/SwifterSwift/SwifterSwift/pull/756) by [mmdock](https://github.com/mmdock)
 
 ### Changed
 - **UIImage**:
   - Implemented `filled(withColor:)` using `UIGraphicsImageRenderer` when available. [#733](https://github.com/SwifterSwift/SwifterSwift/pull/733)
   - Updated `kilobytesSize` to be computed independently from `bytesSize` [#753](https://github.com/SwifterSwift/SwifterSwift/pull/753) by [mmdock](https://github.com/mmdock)
   - Updated `init?(base64String:)` to take in a `scale` factor paramater. [#753](https://github.com/SwifterSwift/SwifterSwift/pull/753) by [mmdock](https://github.com/mmdock)
-  
-- **UIImage**:
   - Refactored `tint(_:blendMode:)` using UIGraphicsImageRenderer if available. [#731](https://github.com/SwifterSwift/SwifterSwift/pull/731) by [FraDeliro](https://github.com/FraDeliro).
 
 - **Sequence**:
@@ -41,6 +43,8 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 ### Deprecated
 
 ### Removed
+- **UIView**
+  - Removed enums `ShakeDirection`, `AngleUnit`, `ShakeAnimationType` and their dependent functions.  [#756](https://github.com/SwifterSwift/SwifterSwift/pull/756) by [mmdock](https://github.com/mmdock)
 
 ### Fixed
 
--- a/Sources/SwifterSwift/UIKit/UIViewExtensions.swift
+++ b/Sources/SwifterSwift/UIKit/UIViewExtensions.swift
@@ -9,55 +9,6 @@
 #if canImport(UIKit) && !os(watchOS)
 import UIKit
 
-// MARK: - enums
-public extension UIView {
-
-    /// SwifterSwift: Shake directions of a view.
-    ///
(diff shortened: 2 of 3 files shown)


Sources/SwifterSwift/UIKit/UIViewExtensions.swift
```diff
@@ -351,35 +302,61 @@ public extension UIView {
...
-    ///   - angle: angle to rotate view by.
-    ///   - type: type of the rotation angle.
-    ///   - animated: set true to animate rotation (default is true).
+    ///   - radian: radians to rotate view by.
+    ///   - animated: set true to animate rotation (default is false).
     ///   - duration: animation duration in seconds (default is 1 second).
     ///   - completion: optional completion handler to run with animation finishes (default is nil).
-    func rotate(byAngle angle: CGFloat, ofType type: AngleUnit, animated: Bool = false, duration: TimeInterval = 1, completion: ((Bool) -> Void)? = nil) {
-        let angleWithType = (type == .degrees) ? .pi * angle / 180.0 : angle
+    func rotate(byRadian radian: CGFloat, animated: Bool = false, duration: TimeInterval = 1, completion: ((Bool) -> Void)? = nil) {
```
guykogus: How do we feel about `Measurement<UnitAngle>`? Unfortunately the conversion between radians and degrees isn't exactly correct (don't ask me why, it's the simplest freaking calculation in the world), but there you have it...
```suggestion
    func rotate(byAngle angle: Measurement<UnitAngle>, animated: Bool = false, duration: TimeInterval = 1, completion: ((Bool) -> Void)? = nil) {
```
I guess the only issue is that `Measurement` is only available from iOS 10+ and we're supporting back to iOS 8, so we'd need 2 implementations.

If you don't want to go for this option, can you please change the parameter name to `radians`? It should be pluralised.

Sources/SwifterSwift/UIKit/UIViewExtensions.swift
```diff
@@ -351,35 +302,61 @@ public extension UIView {
...
     /// - Parameters:
-    ///   - angle: angle to rotate view to.
-    ///   - type: type of the rotation angle.
+    ///   - degree: degrees to rotate view by.
     ///   - animated: set true to animate rotation (default is false).
     ///   - duration: animation duration in seconds (default is 1 second).
     ///   - completion: optional completion handler to run with animation finishes (default is nil).
-    func rotate(toAngle angle: CGFloat, ofType type: AngleUnit, animated: Bool = false, duration: TimeInterval = 1, completion: ((Bool) -> Void)? = nil) {
-        let angleWithType = (type == .degrees) ? .pi * angle / 180.0 : angle
+    func rotate(byDegree degree: CGFloat, animated: Bool = false, duration: TimeInterval = 1, completion: ((Bool) -> Void)? = nil) {
```
guykogus: Same thing here about making it pluralised
```suggestion
    func rotate(byDegrees degrees: CGFloat, animated: Bool = false, duration: TimeInterval = 1, completion: ((Bool) -> Void)? = nil) {
```

guykogus: Same for the other functions...

Sources/SwifterSwift/UIKit/UIViewExtensions.swift
```diff
@@ -401,32 +378,33 @@ public extension UIView {
...
-            animation.timingFunction = CAMediaTimingFunction(name: CAMediaTimingFunctionName.easeOut)
-        case .easeInOut:
-            animation.timingFunction = CAMediaTimingFunction(name: CAMediaTimingFunctionName.easeInEaseOut)
-        }
+        let animation = CAKeyframeAnimation(keyPath: "transform.translation.y")
+        animation.timingFunction = CAMediaTimingFunction(name: animationType)
+        CATransaction.setCompletionBlock(completion)
+        animation.duration = duration
+        animation.values = [-20.0, 20.0, -20.0, 20.0, -10.0, 10.0, -5.0, 5.0, 0.0 ]
+        layer.add(animation, forKey: "shake")
```
guykogus: Maybe the names should be different? Like `"vertical shake"` and `"horizontal shake"`?

Tests/UIKitTests/UIViewExtensionsTests.swift
```diff
@@ -314,6 +360,22 @@ final class UIViewExtensionsTests: XCTestCase {
         XCTAssertEqual(view1.transform, view3.transform)
     }
 
+    func testHorizontalShake() {
+        let view = UIView()
+        view.horizontalShake(animationType: .easeIn)
+        let currentAnimation = view.layer.animation(forKey: view.layer.animationKeys()?.first ?? "")
+
+        XCTAssertEqual(currentAnimation?.timingFunction, CAMediaTimingFunction(name: .easeIn))
```
guykogus: You don't want to test the fulfilment like the others? Make sure they're back in the original position, maybe?

CHANGELOG.md
```diff
@@ -41,6 +43,8 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 ### Deprecated
 
 ### Removed
+- **UIView**
+  - Removed enums `ShakeDirection`, `AngleUnit`, `ShakeAnimationType` and their dependent functions.  [#756](https://github.com/SwifterSwift/SwifterSwift/pull/756) by [mmdock](https://github.com/mmdock)
```
guykogus: This means we have a breaking change, technically. Shouldn't be including this until the next major version.

mmdock: @guykogus yeah, i was thinking maybe i just  add those back and mark deprecated

Sources/SwifterSwift/UIKit/UIViewExtensions.swift
mmdock: @guykogus haven't written out the unit tests yet, but wrote the function that uses Measurement<UnitAngle>.  Totally didn't know about `Measurement` at all.

Before I proceeded anymore, i was wondering the following:

What is the procedure on deprecation / removal? 

I ask because I was wondering if I should just re-add the original functions (that used enum) and deprecate them while also keeping the work i already did and have these new `Measurement` based functions as well.  

I don't want to create unnecessary amount of clutter in the functions, and wasn't sure how SwifterSwift normally would go about doing this.  

I added an additional commit that deprecated my functions i was trying to add to move away from enums, but it seems weird to add deprecated functions in with the PR.  But at the same time, in terms of transitional process I was thinking:  We'd want to mark the enum functions deprecated and get them using the newer deprecated functions if < iOS 10 or the `Measurement` based functions >= iOS 10.  So, enum functions removed first in some build, then when, eventually, SwifterSwift min is iOS 10, we drop the others.

Just my thoughts. It may be too much work this way.

CHANGELOG.md
guykogus: I think that would be better. We actually have a `Deprecated` folder for each section, although currently there's only `StdlibDeprecated.swift`. You'd need to move these to `UIViewDeprecated.swift`.

Issue #714, opened 2019-09-22
Use XcodeGen


omaralbeik:
It would be great to use [XcodeGen](https://github.com/yonaskolb/XcodeGen) to auto-generate the `.xcodeproj` and `xcworkspace` files from a `yml` or `json` file

This will make the code cleaner and help reduce merge conflicts :)

Saik0s:
Hi @omaralbeik, I would like to work on this improvement.

LucianoPAlmeida:
@Saik0s Go for it ;)

moritzsternemann:
Hey @Saik0s 👋 what's your progress? Maybe I can help out 😃

Saik0s:
Hi @moritzsternemann! I am half way through, it just takes some time to write the instruction file for all the targets, so going to finish it in upcoming couple of days. Thanks!

Saik0s:
Hi guys, unfortunately, I am not able to focus on this ticket right now and can get back to it at the end of the next week. @moritzsternemann If you want to pick it up - feel free. In another case, I will work on it next week.

PR #761, opened 2019-11-15
Add XcodeGen support #714

Saik0s:
This PR introduces XcodeGen support #714. The project file is now in `.gitignore` and requires [XcodeGen](https://github.com/yonaskolb/XcodeGen) to generate it from the `project.yml` file. There is also a new `Configs` folder, which contains all the build settings from the old project file. I decided to leave it like that and include `xcconfig` files in `project.yml` to make it cleaner.

The workspace file is still in the repository, but I think it can be completely removed.
Also, I think that instruction how to generate project file should be added somewhere, but I don't know if it should be in `readme.md` or `contributing.md`

## Checklist


project.yml
```diff
@@ -0,0 +1,157 @@
...
+          - "SwifterSwift/UIKit/UIRefreshControlExtensions.swift"
+          - "SwifterSwift/UIKit/UIActivityExtensions.swift"
+          - "SwifterSwift/CoreAnimation/CAGradientLayerExtensions.swift"
+          - "SwifterSwift/AppKit/*"
+  SwifterSwift-watchOS:
+    type: framework
+    platform: [watchOS]
+    platformSuffix: ""
+    configFiles:
+      Debug: Configs/SwifterSwift-watchOS-Debug.xcconfig
```
omaralbeik: As much as I love seperating configs into their own files, I don't think we need them for this project
Must of the configs we have are Xcode's defaults

LucianoPAlmeida:
Nice work here, just raising a point that if the users who use Carthage will continue to be able to install, I remember it used to rely on `xcodeproj` to be able to work. If we drop the xcodeproj that may be a problem ... So we probably will have to keep the xcodeproj versioned.

omaralbeik:
@LucianoPAlmeida I remember Xcodegen supports Carthage, users should not face any problems installing the project

Saik0s:
@omaralbeik Unfortunately, I think that @LucianoPAlmeida is right. XcodeGen supports Carthage in terms of knowing the correct path for dependencies installed using Carthage. But in the SwifterSwift case, the repository itself should be supported by Carthage, and to do that, it should have a project file with shared schemes in the root folder of a repository.

omaralbeik:
@Saik0s I thought of adding XcodeGen to ignore project files, in this case, I'm sorry I don't think we will be able to use it 😢

omaralbeik:
I'm very sorry I missed this while opening the issue 😬

Saik0s:
It looks like Carthage is not going even to support SPM in the nearest future because of some [edge cases](https://github.com/Carthage/Carthage/pull/1945). I should have checked it before starting this PR. At least learned something new :)

PR #762, opened 2019-11-17
Exclude AppKit extensions to be compiled for macCatalyst

MaxHaertwig:
## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,8 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 
 ### Fixed
 
+- Fixed build error occurring when building AppKit extensions for macCatalyst. [#762](https://github.com/SwifterSwift/SwifterSwift/pull/762) by [MaxHaertwig](https://github.com/maxhaertwig).
+
 ### Security
 
 ---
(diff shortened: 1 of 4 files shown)


Sources/SwifterSwift/AppKit/NSColorExtensions.swift
```diff
@@ -6,7 +6,7 @@
 //  Copyright © 2019 SwifterSwift
 //
 
-#if canImport(AppKit)
+#if canImport(AppKit) && !targetEnvironment(macCatalyst)
```
omaralbeik: Why don't we use `#if os(macOS)` instead here?

guykogus: I think both are correct, technically. We use this `#if` elsewhere in the project, so this usage is in line with the others, and we should change it everywhere else if we use `#if os(macOS)` here.
If we're thinking about future-proofing, unfortunately it's impossible to know what will happen. Will `AppKit` suddenly be available on another platform that doesn't use `NSColor`? If so, we'll need to update all of these `#if`s. If the new platform does use `NSColor` but we've used `#if os(macOS)` then none of these extensions will be available on the new platform (and we won't know about it, most likely).

CHANGELOG.md
```diff
@@ -47,6 +47,8 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 
 ### Fixed
```
guykogus: Remove this empty line

Sources/SwifterSwift/AppKit/NSColorExtensions.swift
omaralbeik: Apple ¯\\_(ツ)_/¯

MaxHaertwig: We also use `#if canImport(AppKit) && !targetEnvironment(macCatalyst)` in `ColorExtensions.swift` and other files.

guykogus: @MaxHaertwig we already have :)

PR #764, opened 2019-11-21
Fixed build errors in KeyedDecodingContainerTests

RomanPodymov:
🚀 Fixed build errors in `KeyedDecodingContainerTests` caused by `try!`.

## Checklist

### DIFF
--- a/Tests/SwiftStdlibTests/KeyedDecodingContainerExtensionsTests.swift
+++ b/Tests/SwiftStdlibTests/KeyedDecodingContainerExtensionsTests.swift
@@ -1,5 +1,5 @@
 //
-//  KeyedDecodingContainerTests.swift
+//  KeyedDecodingContainerExtensionsTests.swift
 //  SwifterSwift
 //
 //  Created by Francesco Deliro on 23/10/2019.
--- a/Tests/SwiftStdlibTests/KeyedDecodingContainerExtensionsTests.swift
+++ b/Tests/SwiftStdlibTests/KeyedDecodingContainerExtensionsTests.swift
@@ -31,7 +31,7 @@ final class KeyedDecodingContainerTests: XCTestCase {
     func testDecodeBoolAsIntOrStringDataAsIntSuccessful() {
         let isPlayingAndIsFullScreenAsInt = #"{"isPlaying": 1, "isFullScreen": 0}"#
         let data = mockJsonData(from: isPlayingAndIsFullScreenAsInt)
-        let video = try? JSONDecoder().decode(Video.self, from: data)
+        let video = try! JSONDecoder().decode(Video.self, from: data) // swiftlint:disable:this force_try
         XCTAssertNotNil(video)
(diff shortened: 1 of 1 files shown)


Tests/SwiftStdlibTests/KeyedDecodingContainerExtensionsTests.swift
```diff
@@ -31,41 +31,45 @@ final class KeyedDecodingContainerTests: XCTestCase {
     func testDecodeBoolAsIntOrStringDataAsIntSuccessful() {
         let isPlayingAndIsFullScreenAsInt = #"{"isPlaying": 1, "isFullScreen": 0}"#
         let data = mockJsonData(from: isPlayingAndIsFullScreenAsInt)
-        let video = try! JSONDecoder().decode(Video.self, from: data)
-        XCTAssertTrue(video.isPlaying)
-        XCTAssertEqual(video.isFullScreen, false)
+        let video = try? JSONDecoder().decode(Video.self, from: data)
```
guykogus:
```suggestion
        let video = try! JSONDecoder().decode(Video.self, from: data) // swiftlint:disable:this force_try
```

LucianoPAlmeida:
There is a more important issue to be solved here is the CI should've caught this on the PR that got merged ...
Currently, the Swiftlint script on the project is under a `if [[ ${TRAVIS} == "" ]]; then` so it will not run on Travis ... I remember I did that 😅 but can't remember why ... probably because we are already running Swiftlint on Danger which by the way only output the issues on the report and doesn't actually break the build :(

LucianoPAlmeida:
Just checked and remembered that we were having some issues with swiftlint, let's just remove all the `if [[ ${TRAVIS} == "" ]]; then` and see ... 
Can you do that @RomanPodymov? Really appreciate it :))

Tests/SwiftStdlibTests/KeyedDecodingContainerExtensionsTests.swift
RomanPodymov: Hello @guykogus 
Yes, the problem is that we cannot use `try!` with swiftlint.

RomanPodymov:
> Just checked and remembered that we were having some issues with swiftlint, let's just remove all the `if [[ ${TRAVIS} == "" ]]; then` and see ...
> Can you do that @RomanPodymov? Really appreciate it :))

Done.

LucianoPAlmeida:
Seems like we have some issues there, let's do this: Can you revert this last commit, to merge this already and I'll take a deeper look on this later.
At the moment is important that the project is building when someone checks it out :)

RomanPodymov:
> Seems like we have some issues there, let's do this: Can you revert this last commit, to merge this already and I'll take a deeper look on this later.
> At the moment is important that the project is building when someone checks it out :)
> cc @RomanPodymov

Done.

LucianoPAlmeida:
Thank you @RomanPodymov

PR #769, opened 2019-11-27
Add the manual layout properties for UIView.

tospery:
Add the manual layout properties for UIView.

### DIFF
--- a/Sources/SwifterSwift/UIKit/UIViewExtensions.swift
+++ b/Sources/SwifterSwift/UIKit/UIViewExtensions.swift
@@ -227,6 +227,66 @@ public extension UIView {
     }
     // swiftlint:enable identifier_name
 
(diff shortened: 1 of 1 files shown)


LucianoPAlmeida:
Hey @tospery :))
A couple of points:
* Can you add a changelog entry?
* Also add unit tests for those extensions.
* Run `swiftlint autocorrect` to fix the lint issues pointed by danger.

guykogus:
To expand on my previous point, I'll give you an example:
```Swift
let view = UIView()
view.left = 50
view.right = 100
```
From reading that I would expect the view's frame to be `CGRect(x: 50, y: 0, width: 50, height: 0)`. But of course it won't, it will be `CGRect(x: 100, y: 0, width: 0, height: 0)`.

PR #772, opened 2019-12-04
Add sorted(like:keyPath:) to Array

MaxHaertwig:
Add sorted(like:keyPath:) to Array to sort an array like another array based on a keyPath.

## Checklist

- [/] New extensions are written in Swift 5.0.
- [/] New extensions support iOS 8.0+ / tvOS 9.0+ / macOS 10.10+ / watchOS 2.0+, or use `@available` if not.
- [/] I have added tests for new extensions, and they passed.
- [/] All extensions have a **clear** comments explaining their functionality, all parameters and return type in English.
- [/] All extensions are declared as **public**.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 ## Upcoming Release
 
 ### Added
+- **Array**:
+  - Added `sorted(like:keyPath:)` to sort an array like another array based on a keyPath. [#772](https://github.com/SwifterSwift/SwifterSwift/pull/772) by [MaxHaertwig](https://github.com/maxhaertwig).
 - **Dictionary**:
   - Added `init(grouping:by:)` to initialize a dictionary by grouping sequence from a hashable `KeyPath`. [#751](https://github.com/SwifterSwift/SwifterSwift/pull/751) by [mmdock](https://github.com/mmdock)
 - **RangeReplaceableCollection**:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 
 ### Added
 - **Array**:
-  - Added `sorted(like:keyPath:)` to sort an array like another array based on a keyPath. [#772](https://github.com/SwifterSwift/SwifterSwift/pull/772) by [MaxHaertwig](https://github.com/maxhaertwig).
+  - Added `sorted(like:keyPath:)` to sort an array like another array based on a key path. [#772](https://github.com/SwifterSwift/SwifterSwift/pull/772) by [MaxHaertwig](https://github.com/maxhaertwig).
 - **Dictionary**:
   - Added `init(grouping:by:)` to initialize a dictionary by grouping sequence from a hashable `KeyPath`. [#751](https://github.com/SwifterSwift/SwifterSwift/pull/751) by [mmdock](https://github.com/mmdock)
 - **RangeReplaceableCollection**:
--- a/Sources/SwifterSwift/SwiftStdlib/ArrayExtensions.swift
+++ b/Sources/SwifterSwift/SwiftStdlib/ArrayExtensions.swift
@@ -34,7 +34,7 @@ public extension Array {
         swapAt(index, otherIndex)
     }
 
-    /// SwifterSwift: Sort an array like another array based on a keyPath. If the other array doesn't contain a certain value, it will be sorted last.
+    /// SwifterSwift: Sort an array like another array based on a key path. If the other array doesn't contain a certain value, it will be sorted last.
     ///
     ///        [MyStruct(x: 3), MyStruct(x: 1), MyStruct(x: 2)].sorted(like: [1, 2, 3], keyPath: \.x)
     ///            -> [MyStruct(x: 1), MyStruct(x: 2), MyStruct(x: 3)]
--- a/Sources/SwifterSwift/SwiftStdlib/ArrayExtensions.swift
+++ b/Sources/SwifterSwift/SwiftStdlib/ArrayExtensions.swift
@@ -33,6 +33,24 @@ public extension Array {
         guard startIndex..<endIndex ~= otherIndex else { return }
         swapAt(index, otherIndex)
     }
+
+    /// SwifterSwift: Sort an array like another array based on a keyPath. If the other array doesn't contain a certain value, it will be sorted last.
+    ///
+    ///        [MyStruct(x: 3), MyStruct(x: 1), MyStruct(x: 2)].sorted(like: [1, 2, 3], keyPath: \.x)
+    ///            -> [MyStruct(x: 1), MyStruct(x: 2), MyStruct(x: 3)]
+    ///
+    /// - Parameters:
+    ///   - otherArray: array containing elements in the desired order.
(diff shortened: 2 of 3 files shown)


Sources/SwifterSwift/SwiftStdlib/ArrayExtensions.swift
```diff
@@ -33,6 +33,24 @@ public extension Array {
         guard startIndex..<endIndex ~= otherIndex else { return }
         swapAt(index, otherIndex)
     }
+
+    /// SwifterSwift: Sort an array like another array based on a keyPath. If the other array doesn't contain a certain value, it will be sorted last.
```
LucianoPAlmeida:
```suggestion
    /// SwifterSwift: Sort an array like another array based on a key path. If the other array doesn't contain a certain value, it will be sorted last.
```

Sources/SwifterSwift/SwiftStdlib/ArrayExtensions.swift
```diff
@@ -33,6 +33,24 @@ public extension Array {
...
+    ///
+    ///        [MyStruct(x: 3), MyStruct(x: 1), MyStruct(x: 2)].sorted(like: [1, 2, 3], keyPath: \.x)
+    ///            -> [MyStruct(x: 1), MyStruct(x: 2), MyStruct(x: 3)]
+    ///
+    /// - Parameters:
+    ///   - otherArray: array containing elements in the desired order.
+    ///   - keyPath: keyPath indiciating the property that the array should be sorted by
+    /// - Returns: sorted array.
+    func sorted<T: Hashable>(like otherArray: [T], keyPath: KeyPath<Element, T>) -> [Element] {
```
LucianoPAlmeida: This is an interesting method, seems very specific tho... could you provide an practical use case for this?

Tests/SwiftStdlibTests/ArrayExtensionsTests.swift
```diff
@@ -36,6 +36,21 @@ final class ArrayExtensionsTests: XCTestCase {
         XCTAssertEqual(swappedEmptyArray, emptyArray)
     }
 
+    func testSortedLike() {
+        let expected1 = [1, 2, 3, 4, 5]
+        let candidate1 = [2, 5, 3, 1, 4]
+        XCTAssertEqual(candidate1.sorted(like: expected1, keyPath: \.self), expected1)
+
+        let candidate2 = [2, 5, 3, 6, 1, 4]
+        XCTAssertEqual(candidate2.sorted(like: expected1, keyPath: \.self), [1, 2, 3, 4, 5, 6])
```
LucianoPAlmeida: If we remove the `5` value from the expected(btw seems not the correct name), can we guarantee that the order is stable? I mean the result is always  `[1, 2, 3, 4, 5, 6]` or it could be `[1, 2, 3, 4, 6, 5])` ?

Tests/SwiftStdlibTests/ArrayExtensionsTests.swift
```diff
@@ -36,6 +36,21 @@ final class ArrayExtensionsTests: XCTestCase {
         XCTAssertEqual(swappedEmptyArray, emptyArray)
     }
 
+    func testSortedLike() {
+        let expected1 = [1, 2, 3, 4, 5]
```
LucianoPAlmeida: Expected seems not the correct name for this, in tests it is normally the name used in the variable used in the assertion expressing the expected result, and since here it is the `like` parameter ... the expected name might be confusing.

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 ## Upcoming Release
 
 ### Added
+- **Array**:
+  - Added `sorted(like:keyPath:)` to sort an array like another array based on a keyPath. [#772](https://github.com/SwifterSwift/SwifterSwift/pull/772) by [MaxHaertwig](https://github.com/maxhaertwig).
```
LucianoPAlmeida:
```suggestion
  - Added `sorted(like:keyPath:)` to sort an array like another array based on a key path. [#772](https://github.com/SwifterSwift/SwifterSwift/pull/772) by [MaxHaertwig](https://github.com/maxhaertwig).
```

Tests/SwiftStdlibTests/ArrayExtensionsTests.swift
```diff
@@ -36,6 +36,21 @@ final class ArrayExtensionsTests: XCTestCase {
...
+        let candidate1 = [2, 5, 3, 1, 4]
+        XCTAssertEqual(candidate1.sorted(like: expected1, keyPath: \.self), expected1)
+
+        let candidate2 = [2, 5, 3, 6, 1, 4]
+        XCTAssertEqual(candidate2.sorted(like: expected1, keyPath: \.self), [1, 2, 3, 4, 5, 6])
+
+        // swiftlint:disable:next nesting
+        struct TestStruct { let prop: String }
```
LucianoPAlmeida: Also, what happens if the candidate size is smaller than the `like` array? Could you add a test case for that?

Sources/SwifterSwift/SwiftStdlib/ArrayExtensions.swift
MaxHaertwig: I have a list of CloudKit record names (Strings). I query CloudKit for the records, but they are returned in random order. This method can be used to sort the retrieved records based on the order of my record names. This principle can also be applied to database queries.

Tests/SwiftStdlibTests/ArrayExtensionsTests.swift
MaxHaertwig: You are correct, values not present in the ordered array are sorted last and are not stable.

Sources/SwifterSwift/SwiftStdlib/ArrayExtensions.swift
LucianoPAlmeida: Hummm... thanks for the response, but why just sort by name would not be sufficient?

MaxHaertwig: Because I want to achieve an order like in the other array that can't be expressed by a sort descriptor. E.g the order [2, 3, 1] can't be expressed using sort descriptors.

Tests/SwiftStdlibTests/ArrayExtensionsTests.swift
guykogus: Try `order` instead

Tests/SwiftStdlibTests/ArrayExtensionsTests.swift
guykogus: It's not completely unstable, it just orders them as they are after the sorted elements, no?

MaxHaertwig: I think it depends on the implementation of the sorting algorithm, `sorted(by:)` in this case.

MaxHaertwig:
@guykogus Applying your suggestion, it would be used like `a1.sorted(by: a2, keyPath: \.id`, but this doesn't sound very intuitive to me either. Maybe `sorted(accordingTo:keyPath:)` or `sorted(order:keyPath:)`? Could also be desiredOrder or targetOrder...

MaxHaertwig:
Thanks. Can I merge?

LucianoPAlmeida:
>Thanks. Can I merge?

@MaxHaertwig sure :))

PR #781, opened 2019-12-26
Added SKProduct extensions and its tests.

strawb3rryx7:
Added localized price property to SKProduct extensions. Also, it's tests written and they passed.

## Checklist

### DIFF
--- a/Sources/SwifterSwift/StoreKit/SKProductExtensions.swift
+++ b/Sources/SwifterSwift/StoreKit/SKProductExtensions.swift
@@ -1,0 +1,23 @@
+//
+//  SKProductExtensions.swift
+//  SwifterSwift
+//
+//  Created by Batuhan Saka on 26.12.2019.
+//  Copyright © 2019 SwifterSwift
+//
+
+#if canImport(StoreKit)
+import StoreKit
+
+public extension SKProduct {
+
+    /// SwifterSwift: Localized price of SKProduct
+    var localizedPrice: String? {
+        let formatter = NumberFormatter()
+        formatter.numberStyle = .currency
+        formatter.locale = self.priceLocale
+        return formatter.string(from: self.price)
+    }
+
+}
+#endif
--- a/Sources/SwifterSwift/StoreKit/SKProductExtensions.swift
+++ b/Sources/SwifterSwift/StoreKit/SKProductExtensions.swift
@@ -11,12 +11,17 @@ import StoreKit
 
 public extension SKProduct {
 
+    private static let priceFormatter: NumberFormatter = {
+        let priceFormatter = NumberFormatter()
+        priceFormatter.numberStyle = .currency
+        return priceFormatter
+    }()
+
     /// SwifterSwift: Localized price of SKProduct
     var localizedPrice: String? {
-        let formatter = NumberFormatter()
-        formatter.numberStyle = .currency
-        formatter.locale = self.priceLocale
-        return formatter.string(from: self.price)
+        let formatter = SKProduct.priceFormatter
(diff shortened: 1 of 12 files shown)


Tests/StoreKitTests/SKProductTests.swift
```diff
@@ -0,0 +1,52 @@
...
+//
+
+import XCTest
+@testable import SwifterSwift
+
+#if canImport(StoreKit)
+import StoreKit
+
+public extension SKProduct {
+
```
omaralbeik: Please make this extension `private`

Tests/StoreKitTests/SKProductTests.swift
```diff
@@ -0,0 +1,52 @@
...
+    /// Creates a mocked `SKProduct`.
+    /// - Parameters:
+    ///   - identifier: Product identifier.
+    ///   - price: Product price.
+    ///   - priceLocale: Product price locale.
+    fileprivate convenience init(identifier: String, price: String, priceLocale: Locale) {
```
omaralbeik: You don't need the modifier `fileprivate` if you make the extension `private`

Tests/StoreKitTests/SKProductTests.swift
strawb3rryx7: It was already fileprivate in the beginning. I changed because of Error #1. But, there had a problem while the tests running via TravisCI. But, I did not face any problem during the running tests in Xcode.

I will give it a try.

Thanks for the review.

strawb3rryx7: Requested changes have been applied. @omaralbeik

Sources/SwifterSwift/StoreKit/SKProductExtensions.swift
```diff
@@ -0,0 +1,23 @@
...
+//
+
+#if canImport(StoreKit)
+import StoreKit
+
+public extension SKProduct {
+
+    /// SwifterSwift: Localized price of SKProduct
+    var localizedPrice: String? {
+        let formatter = NumberFormatter()
```
guykogus: Formatters are notoriously heavy performance-wise. If somebody is using this on an array of products you'll get a huge performance hit. What might be a good idea here is to create a reusable price formatter:
```Swift
private static let priceFormatter: NumberFormatter = {
    let priceFormatter = NumberFormatter()
    priceFormatter.numberStyle = .currency
    return priceFormatter
}()
```
You'll still need to update the locale in your function, but you can save recreating the formatter. The only problem is that it breaks our guideline of each extension being self-contained. What do you think @LucianoPAlmeida @omaralbeik and others?

strawb3rryx7: Of course. I will take care of that. Your suggestion will be better in case over use scenario.

strawb3rryx7: Performance improvement has been applied. @guykogus

omaralbeik:
@strawb3rryx7 I think Travis is not happy because you didn't add `SKProductExtensions.swift` to all source targets

Screen Shot 2019-12-29 at 13 02 19

Sources/SwifterSwift/StoreKit/SKProductExtensions.swift
LucianoPAlmeida: @guykogus To me this is fine... I think this rule could be discussed if we really have a good reason, and in this case, seems like a good one :) @omaralbeik what you think?

SD10:

PR #784, opened 2020-01-01
Danger API key expose to forks 

LucianoPAlmeida:
I really wanted we could find a better way to do this :)) 

## Checklist


LucianoPAlmeida:
This is a problem deeper than I expected ... as we can't put the key hardcoded on the file and commit it because github security destroys the token as soon as we commit. The only other way is to add it as a secret, but secrets can't be exposed for a PR coming from a fork so we would fail danger on PR's coming from forks on missing key. I still couldn't find a way to add it other than hardcoded or secret... maybe someone has an idea :)) 
cc @guykogus @omaralbeik

LucianoPAlmeida:
Seems like, the only option we have right now is to set it up hard coded 😭 
https://github.community/t5/GitHub-Actions/Make-secrets-available-to-builds-of-forks/m-p/30685#M511

guykogus:
@LucianoPAlmeida what's the danger (excuse the pun) of exposing the key?

The only other workaround I can think of right now is to give public write access to the repo, lock the `master` branch to non-admins, and get people to create PRs directly from branches.

LucianoPAlmeida:
> @LucianoPAlmeida what's the danger (excuse the pun) of exposing the key?
> 
None, when we generate the key, we configure the rights that it has. The one we are using has only, public access and Github action workflow rights, so there is no danger exposing it. The problem is that Github security invalidates it anyways as soon as it sees the hardcoded key in a commit :(( 

> The only other workaround I can think of right now is to give public write access to the repo, lock the `master` branch to non-admins, and get people to create PRs directly from branches.
> 

That would be a possibility, but I think it would create a more complex flow for new contributors and I think create PRs via fork is an important part of the open source flow. Not sure, maybe we should consider that if it is the only option.

 For now, I'm thinking in two other "hacky" options ... 
* Store the hardcoded key reversed(so it bypass the security check), and then manually via sh revert it back to the original.
* Store the key in a public place other than Github, and fetch it via curl or something on danger step.

LucianoPAlmeida:
Super mega "hack" but it works :) Let me know what you think, or if there are any other ideas 👍

LucianoPAlmeida:
> Haha I mean if this is the best option we have for now, and there's no danger in exposing the key like this, then I guess we're stuck until GitHub enables secret envs for forks... 🤷‍♂

Seems like it, they said it is on a queue to look at https://github.community/t5/GitHub-Actions/Make-secrets-available-to-builds-of-forks/m-p/30685#M511
Maybe I should add this link as a comment on the YAML configuration.

LucianoPAlmeida:
cc @omaralbeik

LucianoPAlmeida:
Thank you!

PR #796, opened 2020-01-15
Enable sequence and collection to be sorted by multiple key paths

MaxHaertwig:
…and:and:) to sort collections/sequences by multiple key paths

## Checklist

### DIFF
--- a/Sources/SwifterSwift/SwiftStdlib/MutableCollectionExtensions.swift
+++ b/Sources/SwifterSwift/SwiftStdlib/MutableCollectionExtensions.swift
@@ -7,19 +7,54 @@
 //
 
 public extension MutableCollection where Self: RandomAccessCollection {
-    /// SwifterSwift: Sort the collection  based on a keypath and a compare function.
+    /// SwifterSwift: Sort the collection based on a keypath and a compare function.
     ///
-    /// - Parameter path: Key path to sort. The key path type must be Comparable.
+    /// - Parameter keyPath: Key path to sort by. The key path type must be Comparable.
     /// - Parameter compare: Comparation function that will determine the ordering.
     mutating func sort<T>(by keyPath: KeyPath<Element, T>, with compare: (T, T) -> Bool) {
         sort { compare($0[keyPath: keyPath], $1[keyPath: keyPath]) }
     }
 
-    /// SwifterSwift: Sort the collection  based on a keypath and a compare function.
+    /// SwifterSwift: Sort the collection based on a keypath.
     ///
-    /// - Parameter path: Key path to sort. The key path type must be Comparable.
+    /// - Parameter keyPath: Key path to sort by. The key path type must be Comparable.
     mutating func sort<T: Comparable>(by keyPath: KeyPath<Element, T>) {
         sort { $0[keyPath: keyPath] < $1[keyPath: keyPath] }
+    }
+
+    /// SwifterSwift: Sort the collection based on two key paths. The second one will be used in case the values of the first one match.
+    ///
+    /// - Parameters:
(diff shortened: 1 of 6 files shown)


Sources/SwifterSwift/SwiftStdlib/MutableCollectionExtensions.swift
```diff
@@ -7,20 +7,55 @@
...
+        }
+    }
+
+    /// SwifterSwift: Sort the collection based on three key paths. Whenever the values of one key path match, the next one will be used.
+    ///
+    /// - Parameters:
+    ///     - keyPath1: Key path to sort by. Must be Comparable.
+    ///     - keyPath2: Key path to sort by in case the values of `keyPath1` match. Must be Comparable.
+    ///     - keyPath3: Key path to sort by in case the values of `keyPath1` and `keyPath2` match. Must be Comparable.
+    mutating func sort<T: Comparable, U: Comparable, V: Comparable>(by keyPath1: KeyPath<Element, T>,
```
LucianoPAlmeida: Humm ... that interesting, but I wonder if do this manually is the right choice because we have to much manual work and only few options. But I don't know if we should consider using gyb or any other strategy for code generation for this.
cc @guykogus

MaxHaertwig: [Sourcery](https://github.com/krzysztofzablocki/Sourcery) could be a viable option here. But even without code generation, sorting by three different properties already covers the vast majority of cases.

LucianoPAlmeida: Makes sense but even only 3, I think generates a lot of manual work and code "duplication", lets hear the opinion of the others on that :))

MaxHaertwig: To reduce code duplication, I could factor out the comparison functions and re-use them in `sort` and `sorted`.

LucianoPAlmeida: Sure, that would be nice... but let's wait for the other thoughts before doing this, to avoid re-work :)

omaralbeik: Hmm, I'm not sure about the extension, and I don't think we should have extensions that rely on other extensions, this will defeat the purpose of this library, keeping in mind that many people use it as a reference where they can pick only what they need and expect it to work without any external dependencies

MaxHaertwig: Are you opposed to the extensions being added with this pull request, the idea of code generation, or having the comparison happen in a separate function?

omaralbeik: @MaxHaertwig two things:

I have no problem with the extensions added in this PR 👍 

but I don't like having the comparison happen in a separate function, while this is nice in other projects, in this project we try to have methods not rely on any other dependencies (as much as possible)

Also me personally, I've never used the project as a framework, instead, I use it a reference where I can copy what I need, I'm not sure how code generation will affect this

MaxHaertwig: Okay, so can we get this merged?

omaralbeik: @MaxHaertwig I don't see why we can't, the branch is out of date, please update it and ping me

MaxHaertwig: @omaralbeik Done

PR #797, opened 2020-01-16
Added model parsing extension in codable type

mustafagunes:
## Checklist


Sources/SwifterSwift/SwiftStdlib/DecodableExtensions.swift
```diff
@@ -0,0 +1,31 @@
...
+    static func decode(_ data: Data) throws -> Self? {
+        let decoder = JSONDecoder()
+        decoder.dataDecodingStrategy = .deferredToData
+        do {
+            let parsed = try decoder.decode(self, from: data)
+            return parsed
+        } catch {
+            return nil
+        }
+    }
```
omaralbeik: @mustafagunes Thanks for your extension, here is my two cents on it:

1. The method is marked as `throws`, but you're ignoring the error, but not throwing the decoding error anyway!
2. What happens if implementers have a custom `JSONDecoder`, or different than the one used in this method?
3. Why are you using `dataDecodingStrategy` = `.deferredToData`?
4. I think this makes sense to be a failable initializer instead of a static function

```swift
init?(from data: Data, using decoder: JSONDecoder = .init()) {
    guard let parsed = try? decoder.decode(Self.self, from: data) else { return nil }
    self = parsed
}
```

Tests/SwiftStdlibTests/CodableExtensionsTests.swift
```diff
@@ -0,0 +1,40 @@
+//
+//  CodableExtensionsTests.swift
+//  SwifterSwift
+//
+//  Created by Mustafa GUNES on 16.01.2020.
+//  Copyright © 2020 SwifterSwift
+//
+
+import XCTest
+
```
omaralbeik: Please remove the extra empty line here

Tests/SwiftStdlibTests/CodableExtensionsTests.swift
```diff
@@ -0,0 +1,40 @@
...
+@testable import SwifterSwift
+
+private struct City: Codable {
+    var cityId: Int
+    var name: String
+    var photoUrl: String
+
+    private enum CodingKeys: String, CodingKey {
+        case cityId = "id"
```
omaralbeik: why `cityId = "id"`, just make it `id`

Tests/SwiftStdlibTests/CodableExtensionsTests.swift
```diff
@@ -0,0 +1,40 @@
...
+        case name
+        case photoUrl = "photo_url"
```
omaralbeik: Same here, just make it `url`

Tests/SwiftStdlibTests/CodableExtensionsTests.swift
```diff
@@ -0,0 +1,40 @@
+//
+//  CodableExtensionsTests.swift
```
omaralbeik: You should be testing for `Decodable` only sense this is what you're extending in your extension

Tests/SwiftStdlibTests/CodableExtensionsTests.swift
```diff
@@ -0,0 +1,40 @@
...
+
+    private var mockJsonData: Data {
+        return #"{"id": 1, "name": "Şanlıurfa", "photo_url": "https://cdn.pixabay.com/photo/2017/09/27/20/55/sanliurfa-2793424_1280.jpg"}"#.data(using: .utf8)!
+    }
+
+    func testDecodeModel() {
+        do {
+            let city = try City.decode(mockJsonData)
+            XCTAssertEqual(city?.cityId, 1)
+            XCTAssertEqual(city?.name, "Şanlıurfa")
```
omaralbeik: Why aren't you testing for `id`, `name`, but not `url`?

Tests/SwiftStdlibTests/CodableExtensionsTests.swift
```diff
@@ -0,0 +1,40 @@
...
+//
+//  Created by Mustafa GUNES on 16.01.2020.
+//  Copyright © 2020 SwifterSwift
+//
+
+import XCTest
+
+@testable import SwifterSwift
+
+private struct City: Codable {
```
omaralbeik: No need for `Codable` here, `Decodable` is enough

Tests/SwiftStdlibTests/CodableExtensionsTests.swift
mustafagunes: @omaralbeik I named it that way because of the swiftlint alert. But right now I would fix the problems :)

mustafagunes:
@omaralbeik I've completed the arrangements. Thank you for your support :)

Tests/SwiftStdlibTests/CodableExtensionsTests.swift
omaralbeik: > @omaralbeik I named it that way because of the swiftlint alert. But right now I would fix the problems :)

In this case, you can tell swiftlint to ignore the rule for this line

```swiftlint
// swiftlint:disable:next identifier_name
case id
```

Tests/SwiftStdlibTests/CodableExtensionsTests.swift
```diff
@@ -0,0 +1,41 @@
+//
+//  DecodableExtensionsTests.swift
```
omaralbeik: Please also rename the file to `DecodableExtensionsTests.swift` instead of `CodableExtensionsTests.swift`

Tests/SwiftStdlibTests/CodableExtensionsTests.swift
```diff
@@ -0,0 +1,40 @@
...
+
+    private enum CodingKeys: String, CodingKey {
+        case cityId = "id"
+        case name
+        case photoUrl = "photo_url"
+    }
+}
+
+final class CodableExtensionsTests: XCTestCase {
+
```
omaralbeik: Consider adding a test case for invalid data as well, an example: a city with an invalid URL 😉

Tests/SwiftStdlibTests/CodableExtensionsTests.swift
omaralbeik: @mustafagunes, please replace this with:

```swift
private struct City: Decodable {
    ...
}
```

Tests/SwiftStdlibTests/CodableExtensionsTests.swift
mustafagunes: @omaralbeik Thanks :)

mustafagunes: @omaralbeik  Should Method be wrong?

example:
```
func testDecodeModelInvalidData() {
    guard let city = City.init(from: mockJsonData) else {
        XCTAssert(false, "Could not parse model")
        return
    }

    let invalidURL = "https://cdn.pixabay.com/photo/2017/09/27/20/55/sanliurfa-2793424_1280"

    XCTAssertEqual(city.url, invalidURL)
}
```

Tests/SwiftStdlibTests/CodableExtensionsTests.swift
```diff
@@ -0,0 +1,41 @@
...
+private struct City: Decodable {
+    var id: Int
+    var name: String
+    var url: String
+
+    private enum CodingKeys: String, CodingKey {
+        case id
+        case name
+        case url = "photo_url"
+    }
```
omaralbeik: Please remove the `= "photo_url"`, and replace it in the mock string with "url"

Tests/SwiftStdlibTests/CodableExtensionsTests.swift
```diff
@@ -0,0 +1,41 @@
...
+}
+// swiftlint:enable identifier_name
+
+final class DecodableExtensionsTests: XCTestCase {
+
+    private var mockJsonData: Data {
+        return #"{"id": 1, "name": "Şanlıurfa", "photo_url": "https://cdn.pixabay.com/photo/2017/09/27/20/55/sanliurfa-2793424_1280.jpg"}"#.data(using: .utf8)!
+    }
```
omaralbeik: Please replace `photo_url` with `url`

Tests/SwiftStdlibTests/CodableExtensionsTests.swift
omaralbeik: No, just create an invalid data mock:
```swift
private var invalidMockJsonData: Data {
  return #"{"id": 1, "name": "Istanbul", "url": "invalid url"}"#.data(using: .utf8)!
}
```
and make sure that the initializer fails to create a `City` since the URL is invalid

```swift
XCTAssertNil(City.init(from: invalidMockJsonData))
```

mustafagunes: @omaralbeik I need to verify the URL for this. The task of the plug-in is to parse the JSON data. Why should the model return empty because of the invalid URL?

omaralbeik: Sorry my bad, please make the url property of type `URL` instead of `String`

mustafagunes:
@omaralbeik  I've completed the arrangements :)

Tests/SwiftStdlibTests/DecodableExtensionsTests.swift
```diff
@@ -0,0 +1,52 @@
...
+
+import XCTest
+@testable import SwifterSwift
+
+// swiftlint:disable identifier_name
+private struct City: Decodable {
+    var id: Int
+    var name: String
+    var location: CityLocation
+    var url: String
```
omaralbeik: This should be of type `URL`

Tests/SwiftStdlibTests/DecodableExtensionsTests.swift
```diff
@@ -0,0 +1,52 @@
...
+}
+
+private struct CityLocation: Decodable {
+    var latitude: Double
+    var longitude: Double
+}
+// swiftlint:enable identifier_name
```
omaralbeik: Please remove this type completely as it is not adding any value to the test

Tests/SwiftStdlibTests/DecodableExtensionsTests.swift
```diff
@@ -0,0 +1,44 @@
...
+            return
+        }
+
+        XCTAssertEqual(city.id, 1)
+        XCTAssertEqual(city.name, "Şanlıurfa")
+        XCTAssertEqual(city.url, URL(string: "https://cdn.pixabay.com/photo/2017/09/27/20/55/sanliurfa-2793424_1280.jpg"))
+    }
+
+    func testDecodeModelInvalidData() {
+        XCTAssertNil(City.init(from: invalidMockJsonData))
```
LucianoPAlmeida:
```suggestion
        XCTAssertNil(City(from: invalidMockJsonData))
```

Tests/SwiftStdlibTests/DecodableExtensionsTests.swift
```diff
@@ -0,0 +1,44 @@
...
+    private var mockJsonData: Data {
+        return #"{"id": 1, "name": "Şanlıurfa", "url": "https://cdn.pixabay.com/photo/2017/09/27/20/55/sanliurfa-2793424_1280.jpg"}"#.data(using: .utf8)!
+    }
+
+    private var invalidMockJsonData: Data {
+         #"{"id": "1", "name": "Şanlıurfa", "url": "https://cdn.pixabay.com/photo/2017/09/27/20/55/sanliurfa-2793424_1280.jpg"}"#.data(using: .utf8)!
+    }
+
+    func testDecodeModel() {
+        guard let city = City.init(from: mockJsonData) else {
```
LucianoPAlmeida:
```suggestion
        guard let city = City(from: mockJsonData) else {
```

Tests/SwiftStdlibTests/DecodableExtensionsTests.swift
mustafagunes: @LucianoPAlmeida thanks :)

LucianoPAlmeida:
Can you update the linux tests? Just `swift test --generate-linuxmain` should do it :)

mustafagunes:
> Can you update the linux tests? Just `swift test --generate-linuxmain` should do it :)

@LucianoPAlmeida Why are we doing this?

Tests/SwiftStdlibTests/DecodableExtensionsTests.swift
```diff
@@ -0,0 +1,44 @@
...
+// swiftlint:enable identifier_name
+
+final class DecodableExtensionsTests: XCTestCase {
+
+    private var mockJsonData: Data {
+        return #"{"id": 1, "name": "Şanlıurfa", "url": "https://cdn.pixabay.com/photo/2017/09/27/20/55/sanliurfa-2793424_1280.jpg"}"#.data(using: .utf8)!
+    }
+
+    private var invalidMockJsonData: Data {
+         #"{"id": "1", "name": "Şanlıurfa", "url": "https://cdn.pixabay.com/photo/2017/09/27/20/55/sanliurfa-2793424_1280.jpg"}"#.data(using: .utf8)!
```
LucianoPAlmeida:
```suggestion
       return  #"{"id": "1", "name": "Şanlıurfa", "url": "https://cdn.pixabay.com/photo/2017/09/27/20/55/sanliurfa-2793424_1280.jpg"}"#.data(using: .utf8)!
```

LucianoPAlmeida:
> @LucianoPAlmeida Why are we doing this?

I honestly can't remember why we are doing this manually, but we need to update the Linux tests anyways :)

PR #799, opened 2020-01-21
[CoreLocation] Array extension when CLLocation element to calculate all elements distance

trevorphillips:
## Checklist


Sources/SwifterSwift/SwiftStdlib/ArrayExtensions.swift
```diff
@@ -6,6 +6,40 @@
 //  Copyright © 2016 SwifterSwift
 //
 
+#if canImport(CoreLocation)
```
LucianoPAlmeida: I think those extensions should be under the CoreLocation folder. Eventhough they are arrays they involve CoreLocation :)

trevorphillips: ah ok. I couldn't decide if it was array or corelocation.

LucianoPAlmeida: Is still an array extension, but it should be under the CoreLocation extensions module :)

Sources/SwifterSwift/SwiftStdlib/ArrayExtensions.swift
```diff
@@ -6,6 +6,40 @@
...
+public extension Array where Element: CLLocation {
+
+    /// SwifterSwift: Distance unit.
+    ///
+    /// - meter: Unit of length - meter.
+    /// - kilometer: Unit of length - kilometer.
+    /// - feet: Unit of length - feet.
+    /// - yard: Unit of length - yard.
+    /// - mile: Unit of length - mile.
+    enum DistanceUnit {
```
LucianoPAlmeida: I think we should avoid create new types here.
Also, we can use the foundation `Measurement` API here. 
So we can basically use [UnitLength](https://developer.apple.com/documentation/foundation/unitlength). This also will make our job easy so we can avoid having to do the conversion by hand.

Sources/SwifterSwift/SwiftStdlib/ArrayExtensions.swift
```diff
@@ -137,3 +171,34 @@ public extension Array where Element: Equatable {
...
+
+    /// SwifterSwift: Calculates the sum of distances between each location in the array based on the curvature of the earth.
+    ///
+    /// - Parameter unit: The unit of length to return the distance in.
+    /// - Returns: A double of the distance in the specified unit.
+    func distance(unit: DistanceUnit) -> Double {
+        var distance = 0.0
+
+        for (index, location) in self.enumerated() {
+            if location == self.last { break }
```
LucianoPAlmeida: Use `self.reduce(0, ...)` here could be an option here?

Tests/SwiftStdlibTests/ArrayExtensionsTests.swift
```diff
@@ -91,4 +95,20 @@ final class ArrayExtensionsTests: XCTestCase {
...
+    
+    func testDistance() {
+        #if canImport(CoreLocation)
+        let locations = [
+            CLLocation(latitude: 41.0, longitude: -111.0),
+            CLLocation(latitude: 41.0, longitude: -112.0),
+            CLLocation(latitude: 41.0, longitude: -113.0)
+        ]
+
+        XCTAssertEqual(locations.distance(unit: .meter), 168270.37022814364)
```
LucianoPAlmeida: We don't have to do exactly comparations here, we can use a `XCTAssertEqual(FloatingPoint, FloatingPoint, accuracy: FloatingPoint)` e.g. `XCTAssertEqual(locations.distance(unit: .kilometer), 168.27, accuracy: 0.01)`

Sources/SwifterSwift/SwiftStdlib/ArrayExtensions.swift
trevorphillips: @LucianoPAlmeida I fixed all your suggestions. The only thing I wasn't sure of is if I put it in corelocationextensions or it's seperate file. I chose the later but let me know. thanks!

LucianoPAlmeida: I think what you did is fine :)) 
Can you just check why the CI is failing?

Sources/SwifterSwift/CoreLocation/CLLocationArrayExtensions.swift
```diff
@@ -0,0 +1,30 @@
...
+public extension Array where Element: CLLocation {
+
+    /// SwifterSwift: Calculates the sum of distances between each location in the array based on the curvature of the earth.
+    ///
+    /// - Parameter unit: The unit of length to return the distance in.
+    /// - Returns: The distance in the specified unit.
+    @available(iOS 10.0, tvOS 10.0, macOS 10.12, watchOS 3.0, *)
+    func distance(unit: UnitLength) -> Measurement<UnitLength> {
+        let distance = self.enumerated().reduce(0.0) {
+            if $1.1 == self.last { return $0 }
```
LucianoPAlmeida: To make this code more let's use named variables here, and if possible instead of tuple indexes use labels :))

trevorphillips: @LucianoPAlmeida sorry could you give an example, not quite sure what you mean?

named variables such as: `distance(unitLength unit: UnitLength)`?

and for tuple indexes you mean to use: `if $1.element == self.last { return $0 }` instead?

LucianoPAlmeida: In part yeah, but also replace the $0 an $1 to named variables :)) 
Also, one question if we have an empty or a one-element array? Can you add tests for those cases?

trevorphillips: @LucianoPAlmeida I think I did what you were asking for.

LucianoPAlmeida: That's nice @trevorphillips, thank you!
But taking a look in more detail even though I thought `reduce` was a good option here, it seems to make the code more complicated overall and we always have to check for last in the loop which maybe is not the better solution. 
What do you think about 
```swift
func distance(unitLength unit: UnitLength) -> Measurement<UnitLength> {
        guard count > 1 else {
          return Measurement(value: 0.0, unit: unit)
        }
        var distance = 0.0
        for idx in 0..<count-1 {
            distance += self[idx].distance(from: self[idx + 1])
        }
        return Measurement(value: distance, unit: UnitLength.meters).converted(to: unit)
}
```

trevorphillips: @LucianoPAlmeida yeah works for me. just updated my code.

LucianoPAlmeida: Nice, Thank you :)

LucianoPAlmeida:
@trevorphillips can you fix up the conflicts?

PR #804, opened 2020-02-14
Conform RawRepresentable to Equatable with its RawValue

guykogus:
🚀

When the raw values of a `RawRepresentable` type conform to `Equatable`, it makes sense that the type should be equatable with its raw value. This extension makes that possible, even when the type or its raw value are `Optional`.

## Checklist


LucianoPAlmeida:
@guykogus Glad to have you back \o/  I'll take a look as soon as possible :))

guykogus:
Thanks @LucianoPAlmeida, it's good to be back :)

I wish I could mark an extension as conforming to a protocol, but Swift doesn't like that. I feel like this should be part of the Swift stdlib itself.

Interestingly, if I declare the type as conforming to `Equatable`, e.g.
```Swift
enum Season: String, Equatable { ... }
```
then I don't need the `RawRepresentable` extension, only the `Optional` extension. So maybe I should remove that? But then we don't get the automatic implementation if that declaration isn't made.

Sources/SwifterSwift/SwiftStdlib/RawRepresentableExtensions.swift
```diff
@@ -0,0 +1,112 @@
...
+
+    /// Returns a Boolean value indicating whether two values are equal.
+    ///
+    /// Equality is the inverse of inequality. For any values `a` and `b`,
+    /// `a == b` implies that `a != b` is `false`.
+    ///
+    /// - Parameters:
+    ///   - lhs: A value to compare.
+    ///   - rhs: Another value to compare.
+    static func == (lhs: Self, rhs: RawValue) -> Bool {
```
LucianoPAlmeida: Do we really need those non-optional ones? If I understand that correctly, we already have an overload of type `(Self, RawValue?) -> Bool` that could handle both optional and non-option params, so maybe if don't need this and let the compiler handle the case by using the `(Self, RawValue?) -> Bool` optional param overload for cases like `Enum.a == "a"`. Unless there is a difference in behavior(which seems not), in that case, we may what to keep this.

LucianoPAlmeida:
> I wish I could mark an extension as conforming to a protocol, but Swift doesn't like that.
>

You mean conditional conformance like `extension RawRepresentable: Equatable where RawValue: Equatable`?

LucianoPAlmeida:
> then I don't need the RawRepresentable extension, only the Optional extension. So maybe I should remove that? But then we don't get the automatic implementation if that declaration isn't made.
>

Enums are a complicated case sometimes because they can involve associated data in enum cases so that makes generalizing some implementations a little bit hard to do :(

Sources/SwifterSwift/SwiftStdlib/RawRepresentableExtensions.swift
LucianoPAlmeida: This comment is also valid for the optional extensions :))

guykogus:
> > I wish I could mark an extension as conforming to a protocol, but Swift doesn't like that.
> 
> You mean conditional conformance like `extension RawRepresentable: Equatable where RawValue: Equatable`?

Yep, it's not allowed.

Sources/SwifterSwift/SwiftStdlib/RawRepresentableExtensions.swift
guykogus: Indeed there was no need for this :)

guykogus:
> > then I don't need the RawRepresentable extension, only the Optional extension. So maybe I should remove that? But then we don't get the automatic implementation if that declaration isn't made.
> 
> Enums are a complicated case sometimes because they can involve associated data in enum cases so that makes generalizing some implementations a little bit hard to do :(

If an enum has associated data then it doesn't have a base type and therefore doesn't conform to `RawRepresentable` by default.

LucianoPAlmeida:
> If an enum has associated data then it doesn't have a base type and therefore doesn't conform to `RawRepresentable` by default.
>

That's true

Sources/SwifterSwift/SwiftStdlib/OptionalExtensions.swift
```diff
@@ -111,6 +111,111 @@ public extension Optional where Wrapped: Collection {
...
+
+    /// Returns a Boolean value indicating whether two values are equal.
+    ///
+    /// Equality is the inverse of inequality. For any values `a` and `b`,
+    /// `a == b` implies that `a != b` is `false`.
+    ///
+    /// - Parameters:
+    ///   - lhs: A value to compare.
+    ///   - rhs: Another value to compare.
+    @inlinable static func == (lhs: Optional, rhs: Wrapped.RawValue) -> Bool {
```
LucianoPAlmeida: I think that this could be removed, we probably just need the optional overload `static func == (lhs: Optional, rhs: Wrapped.RawValue?) -> Bool` :)

LucianoPAlmeida:
> > > I wish I could mark an extension as conforming to a protocol, but Swift doesn't like that.
> > 
> > 
> > You mean conditional conformance like `extension RawRepresentable: Equatable where RawValue: Equatable`?
> 
> Yep, it's not allowed.

I still curious why that is 🤔

CHANGELOG.md
```diff
@@ -5,6 +5,10 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 ## Upcoming Release
 
 ### Added
+- **RawRepresentable**:
+  - Conform to `Equatable` when `RawValue` is `Equatable`. [#804](https://github.com/SwifterSwift/SwifterSwift/pull/804) by [guykogus](https://github.com/guykogus)
```
LucianoPAlmeida: This was removed from this PR, so we should remove this entry from changelog :))

Sources/SwifterSwift/SwiftStdlib/OptionalExtensions.swift
LucianoPAlmeida: @guykogus What you think about this? I really think we don't need this :)

PR #805, opened 2020-02-21
mkpolyline-init-locations

trevorphillips:
Added an init to make a polyline out of Cllocations.

## Checklist


Tests/MapKitTests/MKPolylineTests.swift
```diff
@@ -36,6 +44,20 @@ final class MKPolylineTests: XCTestCase {
...
 
+    func testInitWithLocations() {
+        var refCoordinates = locations.map {
+            return CLLocationCoordinate2D(latitude: $0.coordinate.latitude, longitude: $0.coordinate.longitude)
+        }
+
+        let polyline = MKPolyline(locations: locations)
+        let polyline2 = MKPolyline(coordinates: &refCoordinates, count: refCoordinates.count)
+
+        for (coordinate1, coordinate2) in zip(polyline.coordinates, polyline2.coordinates) {
```
LucianoPAlmeida: Why not just use `zip(polyline.coordinates, refCoordinates)`, because the way it is being done right now we testing against the same implementation, so the result will always be the same...

Sources/SwifterSwift/MapKit/MKPolylineExtensions.swift
```diff
@@ -21,6 +21,16 @@ public extension MKPolyline {
         self.init(coordinates: &refCoordinates, count: refCoordinates.count)
     }
 
+    /// SwifterSwift: Create a new MKPolyline from a provided Array of locations.
+    ///
+    /// - Parameter locations: Array of CLLocation(s).
+    convenience init(locations: [CLLocation]) {
```
LucianoPAlmeida: Maybe this would be nicer if we take an `[CLLocationCoordinate2D]` instead?

trevorphillips: There already exists one with that arg. I just thought it would be nice for users to have cllocation also.

Tests/MapKitTests/MKPolylineTests.swift
trevorphillips: i basically copied and pasted from the test above.
`    func testInitWithCoordinates() {
        var refCoordinates = coordinates

        let polyline = MKPolyline(coordinates: coordinates)
        let polyline2 = MKPolyline(coordinates: &refCoordinates, count: refCoordinates.count)

        for (coordinate1, coordinate2) in zip(polyline.coordinates, polyline2.coordinates) {
            XCTAssertEqual(coordinate1.latitude, coordinate2.latitude, accuracy: 0.000000001)
            XCTAssertEqual(coordinate1.longitude, coordinate2.longitude, accuracy: 0.000000001)
        }
    }`

Sources/SwifterSwift/MapKit/MKPolylineExtensions.swift
LucianoPAlmeida: Humm, I see now :)

Sources/SwifterSwift/MapKit/MKPolylineExtensions.swift
```diff
@@ -21,6 +21,16 @@ public extension MKPolyline {
...
+        var refCoordinates = locations.map {
```
guykogus: You can shorten this to 1 line
```suggestion
        var refCoordinates = locations.map { $0.coordinate }
```

Tests/MapKitTests/MKPolylineTests.swift
```diff
@@ -36,6 +44,20 @@ final class MKPolylineTests: XCTestCase {
         }
     }
 
+    func testInitWithLocations() {
+        var refCoordinates = locations.map {
```
guykogus: Same simplification
```suggestion
        var refCoordinates = locations.map { $0.coordinate }
```

Tests/MapKitTests/MKPolylineTests.swift
guykogus: I agree with @LucianoPAlmeida regarding re-using the implementation. I don't like that it was done above either. Just compare `polyline.coordinates` with `coordinates`.

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 ## Upcoming Release
 
 ### Added
+- **MKPolyline**:
+    - Added `init(locations:)` to MKPolyline. [#805](https://github.com/SwifterSwift/SwifterSwift/pull/805) by [trevorphillips](https://github.com/trevorphillips)
```
```suggestion
  - Added `init(locations:)` to MKPolyline. [#805](https://github.com/SwifterSwift/SwifterSwift/pull/805) by [trevorphillips](https://github.com/trevorphillips)
```

Sources/SwifterSwift/MapKit/MKPolylineExtensions.swift
guykogus: @trevorphillips why did you mark this as resolved? In case it's not clear, then function should look like this:
```Swift
convenience init(locations: [CLLocation]) {
    var coordinates = locations.map { $0.coordinate }
    self.init(coordinates: &coordinates, count: coordinates.count)
}
```

Tests/MapKitTests/MKPolylineTests.swift
```diff
@@ -24,13 +24,27 @@ final class MKPolylineTests: XCTestCase {
         (37.329767, -121.885813)
         ].map(CLLocationCoordinate2D.init)
 
-    func testInitWithCoordinates() {
-        var refCoordinates = coordinates
+    let locations = [
```
guykogus: You're just copying the coordinates above, which isn't necessary. I think what you should do is remove `.map(CLLocationCoordinate2D.init)` for `coordinates` above, then use that instead in `testInitWithCoordinates`. Similarly, you would then use `coordinates.map(CLLocation.init)` in `testInitWithLocations`. Does that make sense?

LucianoPAlmeida:
@trevorphillips , there are some conflicts and unaddressed comments here, can you please check them so we can move this forward, thank you :))

PR #806, opened 2020-02-21
cllocation-placemarks

trevorphillips:
## Checklist


Sources/SwifterSwift/CoreLocation/CLLocationExtensions.swift
```diff
@@ -70,6 +69,21 @@ public extension CLLocation {
         return (degrees+360).truncatingRemainder(dividingBy: 360)
     }
 
+    /// SwifterSwift: Retrieve the first placemark of a location.
+    ///
+    /// - Parameter completion: completion handler to run after reverseGeocodeLocation finishes.
+    /// - Returns: An array of  placemarks for a given location or nil if it didn't find any.
```
LucianoPAlmeida: The function doesn't return, maybe it is better to describe this on the completion paramter result :)

LucianoPAlmeida:
@guykogus I pretty much agree with all the points you've mention, and also there is the fact that this pretty much hides any underlaying error that may occur, so would not be possible to error handling correctly ...

trevorphillips:
@guykogus @LucianoPAlmeida I believe I solved the user not being able to see the error that occurs but not the retention of the CLGeocoder object.

LucianoPAlmeida:
> @guykogus @LucianoPAlmeida I believe I solved the user not being able to see the error that occurs but not the retention of the CLGeocoder object.

But I agree with @guykogus in the sense that it is not worth the abstraction ... meaning that we create a method that just calls another method with no difference on parameters and just forwards the response ... and in fact, as Guy mentioned, having a downside. So I agree that this would not be a good addition :)

PR #807, opened 2020-02-25
Updated center text and image extension for UIButton

qchenqizhi:
## Checklist


Sources/SwifterSwift/UIKit/UIButtonExtensions.swift
```diff
@@ -184,6 +184,39 @@ public extension UIButton {
...
+            left: -imageSize.width,
+            bottom: -(imageSize.height + spacing),
+            right: 0.0
+        )
+
+        let titleSize = text.size(withAttributes: [.font: font])
+        imageEdgeInsets = UIEdgeInsets(
+            top: -(titleSize.height + spacing),
+            left: 0.0,
+            bottom: 0.0, right: -titleSize.width
```
guykogus: Separate these lines for consistency

Tests/UIKitTests/UIButtonExtensionsTests.swift
```diff
@@ -158,6 +158,21 @@ final class UIButtonExtensionsTests: XCTestCase {
...
     }
 
+    func testCenterTextAndImageVertically() {
+        let button = UIButton(frame: CGRect(x: 10, y: 10, width: 100, height: 100))
+        let image = UIImage(color: .green, size: CGSize(width: 10, height: 10))
+        button.setTitleForAllStates("Title")
+        button.setImageForAllStates(image)
+        button.centerTextAndImageVertically(spacing: 20)
+
+        if let imageFrame = button.imageView?.frame,
```
guykogus: This should be `guard ... else { XCTFail() }`, otherwise this test could quietly sneak through without actually performing any checks.

Tests/UIKitTests/UIButtonExtensionsTests.swift
```diff
@@ -158,6 +158,21 @@ final class UIButtonExtensionsTests: XCTestCase {
         XCTAssertEqual(button.titleEdgeInsets, UIEdgeInsets(top: 0, left: 10, bottom: 0, right: -10))
     }
 
+    func testCenterTextAndImageVertically() {
```
guykogus: Perhaps add a test to ensure that the width of the frame equals the max width of the title or image frames, and that the button's height is the title + image frames + spacing? That's the expected result, isn't it?

Sources/SwifterSwift/UIKit/UIButtonExtensions.swift
qchenqizhi: Done

Tests/UIKitTests/UIButtonExtensionsTests.swift
qchenqizhi: Done

Tests/UIKitTests/UIButtonExtensionsTests.swift
qchenqizhi: No, the button must specify a size, it will not change its size automatically by setting the image and title. So such tests may not be successful

qchenqizhi:
> You're actually centring the image and label horizontally, with the image on top, IIUC? Can you please provide a screenshot to make it clear.

Yes! look like that:
image

LucianoPAlmeida:
cc @guykogus

Sources/SwifterSwift/UIKit/UIButtonExtensions.swift
```diff
@@ -184,6 +184,40 @@ public extension UIButton {
         contentEdgeInsets = UIEdgeInsets(top: 0, left: insetAmount, bottom: 0, right: insetAmount)
     }
 
+    /// SwifterSwift: Center vertically title text and image on UIButton
```
guykogus: You don't need to mention `UIButton` in the comment for a function in `UIButton`.

qchenqizhi: My english is not good. How about this new one: `alignTextAndImageVertically(imageAboveText:)`?The comments are copied from the previous method. You are very reasonable. I will modify it in the next commit.

qchenqizhi:
Seems need review again, @guykogus @LucianoPAlmeida please help.

Tests/UIKitTests/UIButtonExtensionsTests.swift
```diff
@@ -158,6 +158,47 @@ final class UIButtonExtensionsTests: XCTestCase {
...
+        guard let titleFrame = button.titleLabel?.frame else {
+            XCTFail("Unable to get title frame from button")
+            return
+        }
+
+        XCTAssert(titleFrame.midY > imageFrame.midY)
+        XCTAssertEqual(titleFrame.midX, imageFrame.midX, accuracy: 1.0)
+        XCTAssertEqual(titleFrame.minY - 20, imageFrame.maxY, accuracy: 1.0)
+
+        // Text above imag
```
LucianoPAlmeida:
```suggestion
        // Text above image
```

LucianoPAlmeida:
@qchenqizhi  Everything looks good to me, Just a minor typo :)

Sources/SwifterSwift/UIKit/UIButtonExtensions.swift
```diff
@@ -174,16 +174,53 @@ public extension UIButton {
...
         imageEdgeInsets = UIEdgeInsets(top: 0, left: -insetAmount, bottom: 0, right: insetAmount)
         titleEdgeInsets = UIEdgeInsets(top: 0, left: insetAmount, bottom: 0, right: -insetAmount)
         contentEdgeInsets = UIEdgeInsets(top: 0, left: insetAmount, bottom: 0, right: insetAmount)
     }
 
+    /// SwifterSwift: align title text and image vertically
+    /// - Parameters:
+    ///   - imageAboveText: set true for image above text (default is true).
+    ///   - spacing: spacing between title text and Image.
+    func alignTextAndImageVertically(imageAboveText: Bool = true, spacing: CGFloat = 6.0) {
```
guykogus: Again, you're not aligning them vertically. In fact you're doing the same as the `centerTextAndImage` except that you're not overlapping them, you're setting the vertical ordering, which is probably what the original function should have done in the first place. So I think use the same name but without default parameter values (at least for `imageAboveText`), otherwise you can end up with compiler errors
```suggestion
    func centerTextAndImage(imageAboveText: Bool, spacing: CGFloat) {
```

Tests/UIKitTests/UIButtonExtensionsTests.swift
```diff
@@ -158,6 +158,47 @@ final class UIButtonExtensionsTests: XCTestCase {
...
     }
 
+    func testAlignTextAndImageVertically() {
+        let button = UIButton(frame: CGRect(x: 10, y: 10, width: 100, height: 100))
+        let image = UIImage(color: .green, size: CGSize(width: 10, height: 10))
+        button.setTitleForAllStates("Title")
+        button.setImageForAllStates(image)
+
+        // Image above text
+        button.alignTextAndImageVertically(imageAboveText: true, spacing: 20)
```
guykogus: Create a constant value so that you can reuse it to avoid constantly hard-coding the value (e.g. in the assertions).
```suggestion
        let spacing = CGFloat(20)

        button.alignTextAndImageVertically(imageAboveText: true, spacing: spacing)
```

Sources/SwifterSwift/UIKit/UIButtonExtensions.swift
qchenqizhi: I named it is shorted from: align image and text along the vertical axis, and omitted axis.
But even if my explanation is OK, I know the name is not a good name because it is easy to misunderstand.

So, how about merge the two `centerTextAndImage` methods and providing options like this:

```swift
enum ImagePosition {
    case aboveText
    case belowText
    case onTheLeftOfText
    case onTheRightOfText
}

func centerTextAndImage(image position: ImagePosition = .onTheLeftOfText, spacing: CGFloat)
```

Tests/UIKitTests/UIButtonExtensionsTests.swift
qchenqizhi: Nice

Sources/SwifterSwift/UIKit/UIButtonExtensions.swift
LucianoPAlmeida: @qchenqizhi The problem with this approach is that specifically for the extensions on this project, we avoid creating new types, just as an overall rule to keep the extensions simpler, easy to just copy-paste ...

qchenqizhi: OK, in my case, the most commonly used is image above text and image on the left of text, so as guykogus suggested: use a bool parameter to identify the two types?

```
func centerTextAndImage(imageAboveText: Bool = false, spacing: CGFloat)
```

PR #810, opened 2020-03-04
prevent ugly animation when keyboard is on screen and alert is about to show

bigearsenal:
prevent ugly animation when keyboard is on screen and alert is about to show

## Checklist


guykogus:
I'm yet to see this issue when presenting an alert over a keyboard. I feel like your solution might be more specific to your project than this function itself. It also may result in unwanted side-effects by the callers.

LucianoPAlmeida:
> It also may result in unwanted side-effects by the callers.

+1 
I agree this should be done by the caller and not inside the function

But anyways, @bigearsenal Thank you for taking the time to submit this PR :)

Issue #815, opened 2020-03-16
String Safely subscript - out of bounds


Raber93:
> Please fill out this template when filing an issue.
>
> All comment lines beginning with an **>** instruct you with what info we expect. You can delete those lines once you've filled in the info.
>

[ ] I've read, understood, and done my best to follow the [Contributing guidelines](https://github.com/SwifterSwift/SwifterSwift/tree/master/CONTRIBUTING.md) before opening this issue.

### What did you do?
> Used safely subscript string within a closed range: "A"[safe: 0...1]?? "" 

### What did you expect to happen?
> Expected nil

### What happened instead?
> PIt crashed with: "Fatal error: index is out of bounds"

### SwifterSwift Environment

- SwifterSwift version: 5.1
- Xcode version: 11.3.1
- macOS version running Xcode: 10.15.3
- Swift version: 5.1
- Platform(s) running SwifterSwift:

### Demo Project
>Please link to or upload a project we can download that reproduces the issue.

LucianoPAlmeida:
This seems like a very simple detail in the implementation, I would say it is an easy fix ... PR's are welcome :))  

If someone wants to jump in the file where these subscripts are is `SwiftStdlib/StringExtensions.swift` :))

LucianoPAlmeida:
And @Raber93 Thank you for reporting :)

LucianoPAlmeida:
Fixed on master @Raber93 \o/

PR #816, opened 2020-03-17
String Safely subscript - out of bounds (Issue #815)

cHaLkdusT:
## Checklist

### DIFF
--- a/Sources/SwifterSwift/SwiftStdlib/StringExtensions.swift
+++ b/Sources/SwifterSwift/SwiftStdlib/StringExtensions.swift
@@ -611,6 +611,7 @@ public extension String {
         guard !isEmpty && range.lowerBound >= 0 else { return nil }
         guard let lowerIndex = index(startIndex, offsetBy: range.lowerBound, limitedBy: endIndex) else { return nil }
         guard let upperIndex = index(lowerIndex, offsetBy: range.upperBound - range.lowerBound, limitedBy: endIndex) else { return nil }
+
         return String(self[lowerIndex..<upperIndex])
     }
 
@@ -620,7 +621,7 @@ public extension String {
     ///		"Hello World!"[safe: 21...110] -> nil
     ///
     /// - Parameter range: Closed range.
-    subscript(safe range: ClosedRange<Int>) -> String?{
+    subscript(safe range: ClosedRange<Int>) -> String? {
         guard !isEmpty && range.lowerBound >= 0 else { return nil }
         guard let lowerIndex = index(startIndex, offsetBy: range.lowerBound, limitedBy: index(endIndex, offsetBy: -1)) else { return nil }
         guard let upperIndex = index(lowerIndex, offsetBy: range.upperBound - range.lowerBound, limitedBy: index(endIndex, offsetBy: -1)) else { return nil }
--- a/Sources/SwifterSwift/SwiftStdlib/StringExtensions.swift
+++ b/Sources/SwifterSwift/SwiftStdlib/StringExtensions.swift
@@ -600,51 +600,25 @@ public extension String {
         guard index >= 0 && index < count else { return nil }
         return self[self.index(startIndex, offsetBy: index)]
     }
-
-    /// SwifterSwift: Safely subscript string within a half-open range.
+    
+    /// SwifterSwift: Safely subscript string within a close or half-open range.
     ///
-    ///		"Hello World!"[safe: 6..<11] -> "World"
-    ///		"Hello World!"[safe: 21..<110] -> nil
+    ///        "Hello World!"[safe: 6..<11] -> "World"
+    ///        "Hello World!"[safe: 21..<110] -> nil
     ///
-    /// - Parameter substringRange: Half-open range.
-    subscript(safe substringRange: CountableRange<Int>) -> String? {
-        guard
-            !isEmpty,
-            substringRange.lowerBound >= 0,
-            let lowerIndex = index(startIndex, offsetBy: substringRange.lowerBound, limitedBy: endIndex),
-            let upperIndex = index(lowerIndex, offsetBy: substringRange.upperBound - substringRange.lowerBound, limitedBy: endIndex),
-            let range = range(of: self) else {
+    ///        "Hello World!"[safe: 6...11] -> "World!"
+    ///        "Hello World!"[safe: 21...110] -> nil
+    ///
+    /// - Parameter range: Range expression.
+    subscript<R>(safe range: R) -> String? where R: RangeExpression, R.Bound == Int {
+        let range = range.relative(to: Int.min..<Int.max)
+        guard range.lowerBound >= 0,
+            let lowerIndex = index(startIndex, offsetBy: range.lowerBound, limitedBy: endIndex),
+            let upperIndex = index(startIndex, offsetBy: range.upperBound, limitedBy: endIndex) else {
                 return nil
         }
 
-        if range ~= lowerIndex {
-            return String(self[lowerIndex..<upperIndex])
-        } else {
-            return nil
-        }
-    }
-
-    /// SwifterSwift: Safely subscript string within a closed range.
-    ///
-    ///		"Hello World!"[safe: 6...11] -> "World!"
-    ///		"Hello World!"[safe: 21...110] -> nil
-    ///
-    /// - Parameter substringRange: Closed range.
-    subscript(safe substringRange: ClosedRange<Int>) -> String? {
-        guard
-            !isEmpty,
-            substringRange.lowerBound >= 0,
-            let lowerIndex = index(startIndex, offsetBy: substringRange.lowerBound, limitedBy: endIndex),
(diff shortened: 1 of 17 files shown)


Sources/SwifterSwift/SwiftStdlib/StringExtensions.swift
```diff
@@ -608,8 +608,10 @@ public extension String {
     ///
     /// - Parameter range: Half-open range.
     subscript(safe range: CountableRange<Int>) -> String? {
-        guard let lowerIndex = index(startIndex, offsetBy: max(0, range.lowerBound), limitedBy: endIndex) else { return nil }
+        guard !isEmpty && range.lowerBound >= 0 else { return nil }
+        guard let lowerIndex = index(startIndex, offsetBy: range.lowerBound, limitedBy: endIndex) else { return nil }
         guard let upperIndex = index(lowerIndex, offsetBy: range.upperBound - range.lowerBound, limitedBy: endIndex) else { return nil }
```
guykogus: Combine the guards together. Do the same for the other function.
```suggestion
        guard !isEmpty,
            range.lowerBound >= 0,
            let lowerIndex = index(startIndex, offsetBy: range.lowerBound, limitedBy: endIndex),
            let upperIndex = index(lowerIndex, offsetBy: range.upperBound - range.lowerBound, limitedBy: endIndex) else {
                 return nil
        }
```

Tests/SwiftStdlibTests/StringExtensionsTests.swift
```diff
@@ -345,13 +345,47 @@ final class StringExtensionsTests: XCTestCase {
...
+        XCTAssertEqual(str[safe: 7...7], "o")
         XCTAssertNil(str[safe: 10...18])
+        XCTAssertEqual(str[safe: 11...11], "!")
+        XCTAssertNil(str[safe: 11...12])
+
+        let oneCharStr = "a"
+        XCTAssertEqual(oneCharStr[safe: 0..<0], "")
+        XCTAssertEqual(oneCharStr[safe: 0..<1], "a")
+        XCTAssertNil(oneCharStr[safe: 0..<2])
+        XCTAssertEqual(oneCharStr[safe: 1..<1], "")
```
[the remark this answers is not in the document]
cHaLkdusT: `XCTAssertEqual(oneCharStr[safe: 1..<1], "")` should now return `nil`

Sources/SwifterSwift/SwiftStdlib/StringExtensions.swift
cHaLkdusT: Combined guard statements for both subscripts

cHaLkdusT:
Hi @guykogus and @LucianoPAlmeida, I have made a few updates, could you take a look if the changes are fine?

Sources/SwifterSwift/SwiftStdlib/StringExtensions.swift
```diff
@@ -606,23 +606,45 @@ public extension String {
...
+        guard
+            !isEmpty,
+            substringRange.lowerBound >= 0,
+            let lowerIndex = index(startIndex, offsetBy: substringRange.lowerBound, limitedBy: endIndex),
+            let upperIndex = index(lowerIndex, offsetBy: substringRange.upperBound - substringRange.lowerBound, limitedBy: endIndex),
+            let range = range(of: self) else {
+                return nil
+        }
+
+        if range ~= lowerIndex {
```
guykogus: Why do you need this? You're already checking `!isEmpty` and `substringRange.lowerBound >= 0`.

Sources/SwifterSwift/SwiftStdlib/StringExtensions.swift
```diff
@@ -606,23 +606,45 @@ public extension String {
...
+        guard
+            !isEmpty,
+            substringRange.lowerBound >= 0,
+            let lowerIndex = index(startIndex, offsetBy: substringRange.lowerBound, limitedBy: endIndex),
+            let upperIndex = index(lowerIndex, offsetBy: substringRange.upperBound - substringRange.lowerBound, limitedBy: endIndex),
+            let range = range(of: self) else {
+                return nil
+        }
+        
+        if range ~= lowerIndex && range ~= upperIndex {
```
guykogus: Again, no need to check this.

Sources/SwifterSwift/SwiftStdlib/StringExtensions.swift
cHaLkdusT: Without `if range ~= lowerIndex` this test case fails `XCTAssertNil(oneCharStr[safe: 1..<1])`. It is returning empty string instead of `nil`

Sources/SwifterSwift/SwiftStdlib/StringExtensions.swift
cHaLkdusT: If there's is no additional checker, this test case fails `XCTAssertNil(oneCharStr[safe: 1..<1])`. It is returning empty String instead of `nil`

LucianoPAlmeida:
There are still some conflicts on the changelog ...

Sources/SwifterSwift/UIKit/UIButtonExtensions.swift
```diff
@@ -174,14 +174,35 @@ public extension UIButton {
...
 
-    /// SwifterSwift: Center align title text and image on UIButton
-    ///
-    /// - Parameter spacing: spacing between UIButton title text and UIButton Image.
-    func centerTextAndImage(spacing: CGFloat) {
-        let insetAmount = spacing / 2
-        imageEdgeInsets = UIEdgeInsets(top: 0, left: -insetAmount, bottom: 0, right: insetAmount)
-        titleEdgeInsets = UIEdgeInsets(top: 0, left: insetAmount, bottom: 0, right: -insetAmount)
-        contentEdgeInsets = UIEdgeInsets(top: 0, left: insetAmount, bottom: 0, right: insetAmount)
+    /// SwifterSwift: Center align title text and image
```
LucianoPAlmeida: Something is wrong here... probably an issue with merge

cHaLkdusT: I'm not sure how did that happen, I just updated the CHANGELOG.md and somehow it got affected the `UIButtonExtensions.swift` 😅. I have now updated the branch. Sorry.

LucianoPAlmeida: No problem @cHaLkdusT :)

Sources/SwifterSwift/SwiftStdlib/StringExtensions.swift
LucianoPAlmeida: This shouldn't be a problem here right? this is the ClosedRange method (...) so ..< will not be the case :))

cHaLkdusT: I've updated both ClosedRange and CountableRange methods 😅

Sources/SwifterSwift/SwiftStdlib/StringExtensions.swift
guykogus: Strangely this is the expected outcome in Swift. If you check what happens in `Array`, for example, you'll see that you can access the `endIndex` to get back an empty `ArraySlice`
```Swift
let oneNumArr = [0]
oneNumArr[0..<0] // ArraySlice([])
oneNumArr[1..<1] // ArraySlice([])
oneNumArr[2..<2] // Fatal error: Array index is out of range
```
I think for the sake of consistency, even though I agree that it doesn't make sense, we should maintain this result.

guykogus:
By the way, empty arrays also return an empty slice, so maybe remove the check for `!isEmpty` in the `guard`.

guykogus:
Also, my suggested update handles all types of ranges, not just `ClosedRange` and `CountableRange`. You're welcome ;)

cHaLkdusT:
Awesomesauce @guykogus! 🤩

Sources/SwifterSwift/SwiftStdlib/StringExtensions.swift
cHaLkdusT: I've also added a few test case so that it matches:
```
let emptyArr: [Int] = []
emptyArr[0..<0] // ArraySlice([])
emptyArr[0..<1] // Fatal error: Array index is out of range
```

Sources/SwifterSwift/SwiftStdlib/StringExtensions.swift
```diff
@@ -600,31 +600,27 @@ public extension String {
         guard index >= 0 && index < count else { return nil }
         return self[self.index(startIndex, offsetBy: index)]
     }
+    
+    /// SwifterSwift: Safely subscript string within a close or half-open range.
```
guykogus: There are other types of ranges
```suggestion
    /// SwifterSwift: Safely subscript string within a given range.
```

LucianoPAlmeida:
Thank you @cHaLkdusT @guykogus :)

PR #817, opened 2020-03-17
Added right align image on UIButton method to UIButton extension

ankermarco:
## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 ## Upcoming Release
 
 ### Added
+- **UIButton**:
+  - Added `rightAlignImage(spacing:)` to align button image on the right.
 - **NotificationCenter**:
   - `observeOnce(forName:object:queue:using:)` for observing a single posting of a notification. [#812](https://github.com/SwifterSwift/SwifterSwift/pull/812) by [guykogus](https://github.com/guykogus)
 - **Optional**:
--- a/Sources/SwifterSwift/UIKit/UIButtonExtensions.swift
+++ b/Sources/SwifterSwift/UIKit/UIButtonExtensions.swift
@@ -204,7 +204,28 @@ public extension UIButton {
             contentEdgeInsets = UIEdgeInsets(top: 0, left: insetAmount, bottom: 0, right: insetAmount)
         }
     }
-
+    
+    /// SwifterSwift: Put image to the right side of title
+    /// 
+    /// - Parameter spacing: spacing between title text and image.
+    func rightAlignImage(spacing: CGFloat) {
+        guard
+            let imageSize = imageView?.image?.size,
+            let text = titleLabel?.text,
+            let font = titleLabel?.font
+        else { return }
+        
+        let titleSize = text.size(withAttributes: [.font: font])
+        
+        let titleOffset = -(imageSize.width + spacing)
+        titleEdgeInsets = UIEdgeInsets(top: 0, left: titleOffset, bottom: 0, right: imageSize.width)
+        
+        let imageOffset = -(titleSize.width + spacing)
+        imageEdgeInsets = UIEdgeInsets(top: 0, left: titleSize.width, bottom: 0, right: imageOffset)
+        
+        let edgeOffset = spacing / 2.0
+        contentEdgeInsets = UIEdgeInsets(top: 0, left: edgeOffset, bottom: 0, right: edgeOffset)
+    }
 }
 
 #endif
--- a/Tests/UIKitTests/UIButtonExtensionsTests.swift
+++ b/Tests/UIKitTests/UIButtonExtensionsTests.swift
@@ -180,6 +180,25 @@ final class UIButtonExtensionsTests: XCTestCase {
         XCTAssertEqual(titleFrame.midX, imageFrame.midX, accuracy: 1.0)
         XCTAssertEqual(titleFrame.minY - spacing, imageFrame.maxY, accuracy: 1.0)
     }
+    
+    func testRightAlignImage() {
+        let button = UIButton(frame: CGRect(x: 10, y: 10, width: 100, height: 100))
+        let image = UIImage(color: .red, size: CGSize(width: 10, height: 10))
+        button.setTitleForAllStates("Title")
+        button.setImageForAllStates(image)
+        
+        XCTAssertNotNil(button.imageView)
+        XCTAssertNotNil(button.titleLabel)
+        
+        let spacing: CGFloat = 20
+        button.rightAlignImage(spacing: spacing)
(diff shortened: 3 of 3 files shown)


Sources/SwifterSwift/UIKit/UIButtonExtensions.swift
```diff
@@ -204,7 +204,28 @@ public extension UIButton {
             contentEdgeInsets = UIEdgeInsets(top: 0, left: insetAmount, bottom: 0, right: insetAmount)
         }
     }
-
+    
+    /// SwifterSwift: Put image to the right side of title
+    /// 
+    /// - Parameter spacing: spacing between title text and image.
+    func rightAlignImage(spacing: CGFloat) {
```
LucianoPAlmeida: I think this could be complemented with a `leftAlignImage` method ... maybe? What you think ?

LucianoPAlmeida:
@ankermarco Thank you for taking the time this PR :)
There is just a minor comment... but overall looks good \o/ 
Also, there are some lint warnings and could you attach some screenshots so we can see what the method really does :))

Sources/SwifterSwift/UIKit/UIButtonExtensions.swift
```diff
@@ -204,7 +204,28 @@ public extension UIButton {
...
+        guard
+            let imageSize = imageView?.image?.size,
+            let text = titleLabel?.text,
+            let font = titleLabel?.font
+        else { return }
+        
+        let titleSize = text.size(withAttributes: [.font: font])
```
qchenqizhi: There is similar code to [#807](https://github.com/SwifterSwift/SwifterSwift/pull/807). I think it could be combined into one method, but in order to avoid creating a new type (enum like `ImagePosition` in [a commit](https://github.com/SwifterSwift/SwifterSwift/pull/807/commits/bba4625a6c1330942e6c4a2fb1aa82bc07974786) of 807), Maybe you can implement it with two parameters, similar to `UIStackView`'s axis and alignment. such as (you should find a better name, my English is not good, :-)):

```swift
func centerImageAndText(axis: NSLayoutConstraint.Axis, imageAlignment: UIStackView.Alignment, spacing: CGFloat)
```

LucianoPAlmeida: +1

Sources/SwifterSwift/UIKit/UIButtonExtensions.swift
ankermarco: Hi @LucianoPAlmeida, I don't think so. `leftAlignImage` is the default behaviour of UIButton, right? Do we need to create an extension but not doing much? Or overriding the default behaviour?

Sources/SwifterSwift/UIKit/UIButtonExtensions.swift
ankermarco: Hi @qchenqizhi Thank you for your comments. But, do you mean I will have to stop adding my simple method `rightAlignImage(spacing:)`, instead, creating another method to handle all the different positioning of `image` and `label` for `UIButton`?

In summary, these are the reasons why I didn't mix `rightAlignImge(spacing:)` with your `centerImageAndText` method below:

1. personally, I think the extension method should be simple and effective.
2. positioning image to the left of label on UIButton is some default behaviour, do we need to override it?

Sources/SwifterSwift/UIKit/UIButtonExtensions.swift
LucianoPAlmeida: Makes sense :)

Sources/SwifterSwift/UIKit/UIButtonExtensions.swift
qchenqizhi: You are welcome. So we will have two methods:
one supports the image on the left of text, or the image above text,
and the other supports the image on the right of text.
Isn't this strange?

ankermarco: Why is it strange? 
It sounds to me that you are insisting I should create one big method to covering all the cases? I think that is strange. @qchenqizhi

ankermarco: It seems a deadlock for me to create this PR, however, I feel like to contribute to this project. Is there anyone else could help with this situation? Much appreciated!

qchenqizhi: Honestly, it's okay, but I don't think the code is very tidy. You guys decide. @LucianoPAlmeida @guykogus

guykogus: I think it's fine to have this method separate to the other one.
Personally I wouldn't use any of these anyway as they assume that the button's frame will never change. I have my own custom `UIButton` subclasses that override `titleRect(forContentRect:)` and `imageRect(forContentRect)`.

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 ## Upcoming Release
 
 ### Added
+- **UIButton**:
+  - Added `rightAlignImage(spacing:)` to align button image on the right.
```
```suggestion
  - Added `rightAlignImage(spacing:)` to align button image on the right. [#817](https://github.com/SwifterSwift/SwifterSwift/pull/817) by [ankermarco](https://github.com/ankermarco)
```

Tests/UIKitTests/UIButtonExtensionsTests.swift
```diff
@@ -180,6 +180,25 @@ final class UIButtonExtensionsTests: XCTestCase {
...
+    func testRightAlignImage() {
+        let button = UIButton(frame: CGRect(x: 10, y: 10, width: 100, height: 100))
+        let image = UIImage(color: .red, size: CGSize(width: 10, height: 10))
+        button.setTitleForAllStates("Title")
+        button.setImageForAllStates(image)
+        
+        XCTAssertNotNil(button.imageView)
+        XCTAssertNotNil(button.titleLabel)
+        
+        let spacing: CGFloat = 20
```
guykogus: This is the standard Swift coding style. Not a big deal either way, tbh, just an FYI :)
```suggestion
        let spacing = CGFloat(20)
```

Tests/UIKitTests/UIButtonExtensionsTests.swift
```diff
@@ -180,6 +180,25 @@ final class UIButtonExtensionsTests: XCTestCase {
...
+        button.rightAlignImage(spacing: spacing)
+        let imageFrame = button.imageView!.frame
+        let titleFrame = button.titleLabel!.frame
+        
+        XCTAssert(imageFrame.minX > titleFrame.maxX)
```
guykogus: This test is made redundant by the last test. You can delete this.

PR #818, opened 2020-03-19
Add initializer for a `URL` with an optional `String`

guykogus:
🚀

I often have an object loaded from a server that has optional paths, e.g.
```Swift
struct Food: Codable {
    imagePath: String?
}
```
With this extension users can simply call
```Swift
if let imageURL = URL(food.imagePath) { ... }
```
instead of
```Swift
if let imagePath = food.imagePath,
   let imageURL = URL(imagePath) { ... }
```

Alternatively I could have defined the property as `imagePath: URL?`, but I don't do this for decoded values because if the URL is invalid the entire decoding process fails as opposed to just returning `nil`, so this allows for smoother error handling.

## Checklist

### DIFF
--- a/Tests/FoundationTests/URLExtensionsTests.swift
+++ b/Tests/FoundationTests/URLExtensionsTests.swift
@@ -18,15 +18,6 @@ final class URLExtensionsTests: XCTestCase {
     let params = ["q": "swifter swift"]
     let queryUrl = URL(string: "https://www.google.com?q=swifter%20swift")!
 
-    func testAppendingQueryParameters() {
-        XCTAssertEqual(url.appendingQueryParameters(params), queryUrl)
-    }
-
-    func testAppendQueryParameters() {
-        url.appendQueryParameters(params)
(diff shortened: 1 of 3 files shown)


Tests/FoundationTests/URLExtensionsTests.swift
```diff
@@ -40,6 +31,29 @@ final class URLExtensionsTests: XCTestCase {
...
+        XCTAssertNotNil(baseURL)
+        XCTAssertNil(URL(string: nil, relativeTo: baseURL))
+
+        let string = "/index.html"
+        let optionalString: String? = string
+        XCTAssertEqual(URL(string: optionalString, relativeTo: baseURL), URL(string: string, relativeTo: baseURL))
+        XCTAssertEqual(URL(string: optionalString, relativeTo: baseURL)?.absoluteString, "https://www.example.com/index.html")
+    }
+
+    func testAppendingQueryParameters() {
```
guykogus: I re-arranged the tests so that they're in the same order as the extensions

PR #819, opened 2020-03-21
Make CATransform3D Swifty

guykogus:
🚀
CATransform only had the old Objective-C-style functions. It now gets the same treatment as `CGAffineTransform`, plus `mutating` functions to go with it.

Also added in a little fix for `CAGradientLayerExtensionsTests`, which wasn't included in all the tests.

## Checklist

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -39,3 +39,9 @@ disabled_rules:
 
 excluded:
   - vendor # Folder created by bundle install command that we run on CI
+
+identifier_name:
+  excluded:
+    - x
+    - y
+    - z
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -3,7 +3,7 @@ custom_rules:
     included: ".*\\.swift"
     excluded: ".*Tests\\.swift"
     name: "Missing SwifterSwift: prefix"
-    regex: "///\\s(?!SwifterSwift:\\s)\\w+[^\\n]*$"
+    regex: "\\n\\n(\\s*///\\s(?!SwifterSwift:\\s)\\w+[^\\n]*$)"
     match_kinds:
       - doccomment
     message: "Doc string should always start with the SwifterSwift: prefix"
--- a/Sources/SwifterSwift/CoreAnimation/CATransform3DExtensions.swift
+++ b/Sources/SwifterSwift/CoreAnimation/CATransform3DExtensions.swift
@@ -23,6 +23,7 @@ extension CATransform3D: Equatable {
     /// - Parameters:
     ///   - lhs: A value to compare.
     ///   - rhs: Another value to compare.
+    @inlinable
     public static func == (lhs: CATransform3D, rhs: CATransform3D) -> Bool {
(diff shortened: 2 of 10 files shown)


.swiftlint.yml
```diff
@@ -3,7 +3,7 @@ custom_rules:
     included: ".*\\.swift"
     excluded: ".*Tests\\.swift"
     name: "Missing SwifterSwift: prefix"
-    regex: "///\\s(?!SwifterSwift:\\s)\\w+[^\\n]*$"
+    regex: "\\n\\n(\\s*///\\s(?!SwifterSwift:\\s)\\w+[^\\n]*$)"
```
guykogus: This change means that only the first line of documentation comments is checked. The parentheses indicate where the warning is highlighted in the code.

Sources/SwifterSwift/CoreAnimation/CATransform3DExtensions.swift
```diff
@@ -0,0 +1,261 @@
...
+//
+
+#if canImport(QuartzCore)
+
+import QuartzCore
+
+// MARK: - Equatable
+extension CATransform3D: Equatable {
+
+    // swiftlint:disable missing_swifterswift_prefix
```
LucianoPAlmeida: Why disable this here?

guykogus: Cos it's an exact copy of the Swift documentation, it's not a custom function.

Sources/SwifterSwift/CoreAnimation/CATransform3DExtensions.swift
```diff
@@ -0,0 +1,281 @@
...
+        self = CATransform3DMakeTranslation(tx, ty, tz)
+    }
+
+    /// SwifterSwift: Returns a transform that scales by `(sx, sy, sz)`.
+    /// - Parameters:
+    ///   - sx: x-axis scale
+    ///   - sy: y-axis scale
+    ///   - sz: z-axis scale
+    @inlinable
+    init(scaleX sx: CGFloat, y sy: CGFloat, z sz: CGFloat) { // swiftlint:disable:this identifier_name
```
LucianoPAlmeida: This is just a comment for maybe another PR, we could add identifiers like : i, j, x, y, z as exceptions in for this rule in the yaml :)

guykogus:
@LucianoPAlmeida the SwiftLint rule didn't completely work because some of the variable names are `sx`, `tx`, etc. I don't want to change them because I prefer to keep them in line with the `CGAffine` functions.

LucianoPAlmeida:
> @LucianoPAlmeida the SwiftLint rule didn't completely work because some of the variable names are `sx`, `tx`, etc. I don't want to change them because I prefer to keep them in line with the `CGAffine` functions.

Humm... that's fine I prefer that too, and as you've done adding the `disable:this` is enough :))
TBH I'm personally don't like this `identifier_name` rule but that's a discussion for another time 😆

PR #827, opened 2020-04-19
Added Data.toHexString(), a highly customizable hexdump utilty

dannys42:
## Checklist


dannys42:
The original code was actually fairly trivial.  But I've found limitations with it (e.g. wanting to sometimes output as HTML).  Which is why I went for this route.  I tried to keep the classes and other things as "internally" functional as possible so that the usage is simpler.  But I agree the code is a bit complex to read.

LucianoPAlmeida:
> This is much more than a simple extension to Swift. I'm not sure how I feel about its inclusion, considering the necessity for a non-trivial class to go along with it.
> Thoughts @LucianoPAlmeida @omaralbeik?

Hey @dannys42 thank you for taking the time to put this PR up :)

But I agree with @guykogus, this seems much more than just a helper extension. Seems like this could be a whole `DataFormatter` class in itself or even a small library. Which to me doesn't really fit here since we work with simple and small extensions that in most cases can be copied around...

dannys42:
I see.  That's a shame... I wish there was a framework like your group that collected these larger tools.  I know others have mentioned a wish for a "boost" like framework for Swift, but this project is the closest I've seen to it.

omaralbeik:
Thanks for the PR @dannys42, I agree with @guykogus and @LucianoPAlmeida the extension is very complicated for SwifterSwift, but I like the idea of having a separate repo for utils like this

LucianoPAlmeida:
Since we all agree, closing this one!
Anyways, @dannys42 thank you for taking the time to this :)

PR #841, opened 2020-05-09
CGSize: Added new `+`, `+=`, `-` and `-=` operator extensions for tuple

hamtiko:
Added new `+`, `+=`, `-` and `-=` operator extensions for scalar.

## Checklist

### DIFF
--- a/Sources/SwifterSwift/CoreGraphics/CGSizeExtensions.swift
+++ b/Sources/SwifterSwift/CoreGraphics/CGSizeExtensions.swift
@@ -96,20 +96,6 @@ public extension CGSize {
         return CGSize(width: lhs.width + tuple.width, height: lhs.height + tuple.height)
     }
 
-    /// SwifterSwift: Add a tuple to CGSize.
-    ///
-    ///     let sizeA = CGSize(width: 5, height: 10)
-    ///     let result = (4, 5) + sizeA
-    ///     // result = CGSize(width: 9, height: 15)
-    ///
-    /// - Parameters:
-    ///   - scalar: tuple value.
-    ///   - rhs: CGSize to add to.
-    /// - Returns: The result comes from the addition of the given tuple and CGSize.
-    static func + (tuple: (width: CGFloat, height: CGFloat), rhs: CGSize) -> CGSize {
-        return CGSize(width: tuple.width + rhs.width, height: tuple.height + rhs.height)
-    }
-
     /// SwifterSwift: Add a CGSize to self.
     ///
     ///     var sizeA = CGSize(width: 5, height: 10)
--- a/Sources/SwifterSwift/CoreGraphics/CGSizeExtensions.swift
+++ b/Sources/SwifterSwift/CoreGraphics/CGSizeExtensions.swift
@@ -82,9 +82,37 @@ public extension CGSize {
         return CGSize(width: lhs.width + rhs.width, height: lhs.height + rhs.height)
     }
 
-    /// SwifterSwift: Add a CGSize to self.
+    /// SwifterSwift: Add a tuple to CGSize.
     ///
     ///     let sizeA = CGSize(width: 5, height: 10)
+    ///     let result = sizeA + (5, 4)
+    ///     // result = CGSize(width: 10, height: 14)
+    ///
+    /// - Parameters:
+    ///   - lhs: CGSize to add to.
+    ///   - tuple: tuple value.
+    /// - Returns: The result comes from the addition of the given CGSize and tuple.
+    static func + (lhs: CGSize, tuple: (width: CGFloat, height: CGFloat)) -> CGSize {
+        return CGSize(width: lhs.width + tuple.width, height: lhs.height + tuple.height)
+    }
+
+    /// SwifterSwift: Add a tuple to CGSize.
+    ///
+    ///     let sizeA = CGSize(width: 5, height: 10)
+    ///     let result = (4, 5) + sizeA
+    ///     // result = CGSize(width: 9, height: 15)
+    ///
+    /// - Parameters:
+    ///   - scalar: tuple value.
+    ///   - rhs: CGSize to add to.
+    /// - Returns: The result comes from the addition of the given tuple and CGSize.
+    static func + (tuple: (width: CGFloat, height: CGFloat), rhs: CGSize) -> CGSize {
+        return CGSize(width: tuple.width + rhs.width, height: tuple.height + rhs.height)
+    }
+
+    /// SwifterSwift: Add a CGSize to self.
+    ///
+    ///     var sizeA = CGSize(width: 5, height: 10)
     ///     let sizeB = CGSize(width: 3, height: 4)
     ///     sizeA += sizeB
     ///     // sizeA = CGPoint(width: 8, height: 14)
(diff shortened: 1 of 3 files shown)


cHaLkdusT:
Hi all, I think I may disagree with this PR. It is syntactically correct but is semantically incorrect. For consistency in operator overloading, it is best to follow the model's context. 

Just by looking at this code snippet, I need to explicitly look at what is the `result`s type and value. Is it a `CGSize` or a `CGFloat`?
```
let sizeA = CGSize(width: 5, height: 10)
let result = 5 + sizeA
```
Also, in Swift we are not allowed to do infix operation on two different types, they should of the same type.
```
let operand1 = 1
let operand2 = 1.10
let result = operand1 + operand2 // Error: Binary operator '+' cannot be applied to operands of type 'Int' and 'Double'
```

hamtiko:
@cHaLkdusT nice point. But we have operator for multiplying CGSize with scalar (*, *=), do you think these are also incorrect?

As an alternative I think we can have function extensions like this:

```swift
let sizeA = CGSize(width: 5, height: 10)
let result = sizeA.adding(scalar: 5)
```

or even like this:

```swift
let sizeA = CGSize(width: 5, height: 10)
let result = sizeA.resize(dWidth: 5, dHeight: -3)
// result = CGSize(width: 10, height: 7) 
```

cHaLkdusT:
Atleast for me, I think it would be better to deprecate these APIs:
* `static func * (lhs: CGSize, scalar: CGFloat) -> CGSize`
* `static func * (scalar: CGFloat, rhs: CGSize) -> CGSize` 
* `static func *= (lhs: inout CGSize, scalar: CGFloat)`

Or convert these to non-mutating/mutating methods to promote clarity on the API:
* `size.resized(by multiplier: CGFloat)`
* or `size.resize(by mutiplier: CGFloat)`

But if we are just simply mutating either `CGSize`'s `height` and/or `width` and not doing operations on two `CGSize`'s, you just have to declare `var` instead of `let`.

guykogus:
I don't necessarily agree that we can't mix types with infix operators, but I understand the confusion. Especially with something like adding a scalar to a 2D value, (doesn't matter if it were `CGPoint`, `CGSize`, `CGVector`, etc), it's not a mathematically-founded idea, i.e. matrices need to have the same number of rows and columns to be additive. And the reverse, adding a 2D value to a scalar (the example @cHaLkdusT gave in his first comment), does not seem correct to me at all.

However, you can very well multiply 2 matrices so long as the number of columns in the first matrix equals the number of rows in the second (imagine the 2D values to have 1 column and 2 rows).

To me, it makes sense, for example, to multiply `CGSize` by a scalar. You have an image of size 50x50 and you want to double it to 100x100. That's both a regular use case and conceptually obvious. For `CGVector` it also makes perfect sense. `CGPoint` I guess makes sense if you imagine it like a vector, but it's not the same. I'm on the fence on that one, but I think I've actually used it myself, and if you have a use case then that's a good starting point for arguing that it makes sense.

So far I haven't seen any reason for adding or subtracting scalars to `CGSize` that makes more semantic sense than using the operators for using 2 `CGSize` values.

TL;DR I also vote against this inclusion.

@omaralbeik care to weigh in?

LucianoPAlmeida:
> * size.resized(by multiplier: CGFloat)

@cHaLkdusT I really liked the idea, kind of makes it a clear API. But also think that multiply a scalar into a size  using `*` makes sense. 

>  I'm on the fence on that one, but I think I've actually used it myself, and if you have a use case then that's a good starting point for arguing that it makes sense.

@guykogus Maybe an idea that makes more sense is to have an operator that applies on tuples. 
The major gain I've seen in this is that you can make operation one-liners. e.g.
```swift
var size = CGSize(width: 10, height: 10)
size.width += 5
size.height += 5

size += 5 // Although it may seem not right it simplify syntax
```
So my proposal is to make it operate on tuples:
```swift
// We can make one liners
size += (width: 5, hight: 5) // That could simplify things and make more sense semantically than adding to a  scalar.
//or
size += (5, 5) 
```

guykogus:
@LucianoPAlmeida I like the idea of the tuple. It keeps it semantically correct and simplifies the syntax, which I believe was the original idea. How does that sound to you, @hamtiko?

hamtiko:
@guykogus @LucianoPAlmeida I like the idea too.

Will do updates today.

Thanks.

Sources/SwifterSwift/CoreGraphics/CGSizeExtensions.swift
```diff
@@ -82,9 +82,37 @@ public extension CGSize {
...
     ///
     ///     let sizeA = CGSize(width: 5, height: 10)
+    ///     let result = (4, 5) + sizeA
+    ///     // result = CGSize(width: 9, height: 15)
+    ///
+    /// - Parameters:
+    ///   - scalar: tuple value.
+    ///   - rhs: CGSize to add to.
+    /// - Returns: The result comes from the addition of the given tuple and CGSize.
+    static func + (tuple: (width: CGFloat, height: CGFloat), rhs: CGSize) -> CGSize {
```
guykogus, line 109: As mentioned in the discussion, this version of the operator doesn't make sense. Imagine that with `+` you should be able to have an equivalent assignment operator (`+=`). In the assignment operator for a `CGSize` to the tuple, e.g.
```Swift
var size = (CGFloat(), CGFloat())
size += CGSize.zero
```
`size` would still be a tuple, not a `CGSize`. Therefore it doesn't make sense if the equivalent `+` operator returns a `CGSize`.
I recommend that you delete this method, don't just change the return type.

hamtiko: Ok, I see. Will remove it.
Thanks.

Tests/CoreGraphicsTests/CGSizeExtensionsTests.swift
```diff
@@ -62,6 +62,13 @@ final class CGSizeExtensionsTests: XCTestCase {
         XCTAssertEqual(result.height, 14)
     }
 
+    func testAddTupleRight() {
```
guykogus, line 65:
```suggestion
    func testAddTuple() {
```

Tests/CoreGraphicsTests/CGSizeExtensionsTests.swift
```diff
@@ -86,6 +107,13 @@ final class CGSizeExtensionsTests: XCTestCase {
         XCTAssertEqual(sizeA.height, 6)
     }
 
+    func testSubtractTupleEqual() {
```
guykogus, line 110:
```suggestion
    func testSubtractEqualTuple() {
```

Tests/CoreGraphicsTests/CGSizeExtensionsTests.swift
```diff
@@ -62,6 +62,13 @@ final class CGSizeExtensionsTests: XCTestCase {
         XCTAssertEqual(result.height, 14)
     }
 
+    func testAddTupleRight() {
+        let sizeA = CGSize(width: 5, height: 10)
```
guykogus, line 66:
```suggestion
        let size = CGSize(width: 5, height: 10)
```

Tests/CoreGraphicsTests/CGSizeExtensionsTests.swift
```diff
@@ -70,6 +77,13 @@ final class CGSizeExtensionsTests: XCTestCase {
         XCTAssertEqual(sizeA.height, 14)
     }
 
+    func testAddEqualTuple() {
+        var sizeA = CGSize(width: 5, height: 10)
```
guykogus, line 81:
```suggestion
        var size = CGSize(width: 5, height: 10)
```

Tests/CoreGraphicsTests/CGSizeExtensionsTests.swift
```diff
@@ -78,6 +92,13 @@ final class CGSizeExtensionsTests: XCTestCase {
         XCTAssertEqual(result.height, 6)
     }
 
+    func testSubtractTuple() {
+        let sizeA = CGSize(width: 5, height: 10)
```
guykogus, line 96:
```suggestion
        let size = CGSize(width: 5, height: 10)
```

hamtiko: Nice catch.
Done.
Thanks.

PR #842, opened 2020-05-14
Added method for generating the QR Code and the 2D barcode for UIImageView

heart:
Added method for generating the QR Code and the 2D barcode for UIImageView

```Swift
imageView.generateQRCode(value: "some data")

imageView.generateBarcode128(value: "some data")
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
   - Added `last(where:equals:)` to find the last element of the sequence with having property by given key path equals to given value. [#838](https://github.com/SwifterSwift/SwifterSwift/pull/838) by [hamtiko](https://github.com/hamtiko)
 - **SKNode**:
   - `center`, `topLeft`, `topRight`, `bottomLeft`, `bottomRight` to get anchor position or set position using anchor. [#835](https://github.com/SwifterSwift/SwifterSwift/pull/835) by [rypyak](https://github.com/rypyak)
+- **CGSize**:
+  - Added new `+`, `+=`, `-` and `-=` operator extensions for tuple (width: CGFloat, height: CGFloat). [#841](https://github.com/SwifterSwift/SwifterSwift/pull/841) by [hamtiko](https://github.com/hamtiko)
 
 ### Changed
 - **NSAttributedStringExtensions.swift**:
--- a/Sources/SwifterSwift/CoreGraphics/CGSizeExtensions.swift
+++ b/Sources/SwifterSwift/CoreGraphics/CGSizeExtensions.swift
@@ -82,9 +82,23 @@ public extension CGSize {
         return CGSize(width: lhs.width + rhs.width, height: lhs.height + rhs.height)
     }
 
-    /// SwifterSwift: Add a CGSize to self.
+    /// SwifterSwift: Add a tuple to CGSize.
     ///
     ///     let sizeA = CGSize(width: 5, height: 10)
+    ///     let result = sizeA + (5, 4)
+    ///     // result = CGSize(width: 10, height: 14)
+    ///
+    /// - Parameters:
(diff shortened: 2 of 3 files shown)


LucianoPAlmeida:
> I think that this extension is misplaced. Instead of doing it directly in UIImageView it should be an initialiser for UIImage. I.e.

I would suggest those to not even be in `UIImage` but something under `CIFilter` extensions to abstract this is some way ...

guykogus:
> > I think that this extension is misplaced. Instead of doing it directly in UIImageView it should be an initialiser for UIImage. I.e.
> 
> I would suggest those to not even be in `UIImage` but something under `CIFilter` extensions to abstract this is some way ...

It seems that that's [already been created](https://developer.apple.com/documentation/coreimage/ciqrcodegenerator?language=objc), but I can't actually find any way to access it. It's magically in the documentation but not in the framework.

I personally prefer it being in `UIImage` because I don't think users should need to dig into the lower layers like Core Image to get something that has a real-world use case. It could be useful to have this in `CIImage` because then that can easily be converted to either a `UIImage` or `NSImage`, but then we might as well do that for users of this library and have an `NSImage` equivalent to these methods.

```Swift
convenience init?(qrCode: String, correctionLevel: String = "M") {
    guard let qrCodeData = qrCode.data(using: .isoLatin1),
        let filter = CIFilter(name: "CIQRCodeGenerator",
                              parameters: ["inputMessage": qrCodeData,
                                           "inputCorrectionLevel": correctionLevel]),
        let ciImage = filter.outputImage else {
            return nil
    }
    let ciImageRep = NSCIImageRep(ciImage: ciImage)
    self.init(size: ciImageRep.size)
    addRepresentation(ciImageRep)
}

convenience init?(code128Barcode: String, quietSpace: Double = 7.0) {
    guard let code128BarcodeData = code128Barcode.data(using: .ascii),
        let filter = CIFilter(name: "CICode128BarcodeGenerator",
                              parameters: ["inputMessage": code128BarcodeData,
                                           "inputQuietSpace": quietSpace]),
        let ciImage = filter.outputImage else {
            return nil
    }
    let ciImageRep = NSCIImageRep(ciImage: ciImage)
    self.init(size: ciImageRep.size)
    addRepresentation(ciImageRep)
}
```

If we were Apple then I'd be all for moving the logic to `CIFilter`, but I think to make this "SwifterSwift" it would be better to provide the high-level `(UI/NS)Image` implementations.

Tests/UIKitTests/UIImageExtensionsTests.swift
```diff
@@ -223,6 +223,19 @@ final class UIImageExtensionsTests: XCTestCase {
...
     }
 
-}
+    func testQRCode() {
+        let barcodeSize = CGSize(width: 100, height: 100)
+        let image = UIImage(qrCode: "qr code is awsome", correctionLevel: "M", size: barcodeSize)
```
guykogus, line 228: 😛
```suggestion
        let image = UIImage(qrCode: "qr code is awesome", correctionLevel: "M", size: barcodeSize)
```

Tests/UIKitTests/UIImageExtensionsTests.swift
```diff
@@ -223,6 +223,19 @@ final class UIImageExtensionsTests: XCTestCase {
...
+        XCTAssertNotNil(image)
+        XCTAssertEqual(barcodeSize, image?.size ?? CGSize.zero)
+    }
 
+    func test2DBarCode() {
+        let barcodeSize = CGSize(width: 300, height: 100)
+        let image = UIImage(code128Barcode: "code 128 is awsome", quietSpace: 7.0, size: CGSize(width: 300, height: 100))
```
guykogus, line 235: 😛
```suggestion
        let image = UIImage(code128Barcode: "code 128 is awesome", quietSpace: 7.0, size: CGSize(width: 300, height: 100))
```

PR #844, opened 2020-05-15
Refactored init(light:dark:) to remove deployment target version restrictions.

VincentSit:
## Checklist

See the earlier [discussion][link1] about the reason for creating this PR.

For `NSColor`, since `init (cgColor:)` returns an optional value, it may cause a break change. We may need further discussion.

[link1]:https://github.com/SwifterSwift/SwifterSwift/pull/722#issuecomment-629224728

### DIFF
--- a/Sources/SwifterSwift/UIKit/UIColorExtensions.swift
+++ b/Sources/SwifterSwift/UIKit/UIColorExtensions.swift
@@ -17,9 +17,13 @@ public extension UIColor {
     /// - Parameters:
     ///     - light: Color to use in light/unspecified mode.
     ///     - dark: Color to use in dark mode.
-    @available(iOS 13.0, tvOS 13.0, *)
-    convenience init(light: UIColor, dark: UIColor) {
-        self.init(dynamicProvider: { $0.userInterfaceStyle == .dark ? dark : light })
+    convenience init(light: UIColor, dark: UIColor?) {
+        if #available(iOS 13.0, tvOS 13.0, *),
+            let dark = dark {
+            self.init(dynamicProvider: { $0.userInterfaceStyle == .dark ? dark : light })
+        } else {
+            self.init(cgColor: light.cgColor)
+        }
     }
     #endif
 
--- a/Sources/SwifterSwift/UIKit/UIColorExtensions.swift
+++ b/Sources/SwifterSwift/UIKit/UIColorExtensions.swift
@@ -17,9 +17,8 @@ public extension UIColor {
     /// - Parameters:
     ///     - light: Color to use in light/unspecified mode.
     ///     - dark: Color to use in dark mode.
-    convenience init(light: UIColor, dark: UIColor?) {
-        if #available(iOS 13.0, tvOS 13.0, *),
-            let dark = dark {
+    convenience init(light: UIColor, dark: UIColor) {
+        if #available(iOS 13.0, tvOS 13.0, *) {
             self.init(dynamicProvider: { $0.userInterfaceStyle == .dark ? dark : light })
         } else {
             self.init(cgColor: light.cgColor)
--- a/Tests/UIKitTests/UIColorExtensionsTests.swift
+++ b/Tests/UIKitTests/UIColorExtensionsTests.swift
@@ -15,13 +15,17 @@ import UIKit
 final class UIColorExtensionsTests: XCTestCase {
 
     #if !os(watchOS)
-    @available(iOS 13.0, tvOS 13.0, *)
     func testInitLightDark() {
         let lightModeColor = UIColor.red
         let darkModeColor = UIColor.blue
         let color = UIColor(light: lightModeColor, dark: darkModeColor)
-        XCTAssertEqual(color.resolvedColor(with: UITraitCollection(userInterfaceStyle: .light)), lightModeColor)
-        XCTAssertEqual(color.resolvedColor(with: UITraitCollection(userInterfaceStyle: .dark)), darkModeColor)
+
+        if #available(iOS 13.0, tvOS 13.0, *) {
+            XCTAssertEqual(color.resolvedColor(with: UITraitCollection(userInterfaceStyle: .light)), lightModeColor)
+            XCTAssertEqual(color.resolvedColor(with: UITraitCollection(userInterfaceStyle: .dark)), darkModeColor)
+        } else {
+            XCTAssertEqual(color, lightModeColor)
+        }
     }
     #endif
 
--- a/Tests/UIKitTests/UIColorExtensionsTests.swift
+++ b/Tests/UIKitTests/UIColorExtensionsTests.swift
@@ -19,16 +19,12 @@ final class UIColorExtensionsTests: XCTestCase {
         let lightModeColor = UIColor.red
         let darkModeColor = UIColor.blue
         let color = UIColor(light: lightModeColor, dark: darkModeColor)
-        let color2 = UIColor(light: lightModeColor, dark: nil)
 
         if #available(iOS 13.0, tvOS 13.0, *) {
             XCTAssertEqual(color.resolvedColor(with: UITraitCollection(userInterfaceStyle: .light)), lightModeColor)
             XCTAssertEqual(color.resolvedColor(with: UITraitCollection(userInterfaceStyle: .dark)), darkModeColor)
-            XCTAssertEqual(color2.resolvedColor(with: UITraitCollection(userInterfaceStyle: .light)), lightModeColor)
-            XCTAssertEqual(color2.resolvedColor(with: UITraitCollection(userInterfaceStyle: .dark)), nil)
         } else {
             XCTAssertEqual(color, lightModeColor)
-            XCTAssertEqual(color2, lightModeColor)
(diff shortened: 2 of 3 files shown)


VincentSit:
> Please add a test for when the dark mode colour is `nil`.

Done.

Tests/UIKitTests/UIColorExtensionsTests.swift
```diff
@@ -15,13 +15,20 @@ import UIKit
...
 
     #if !os(watchOS)
-    @available(iOS 13.0, tvOS 13.0, *)
     func testInitLightDark() {
         let lightModeColor = UIColor.red
         let darkModeColor = UIColor.blue
         let color = UIColor(light: lightModeColor, dark: darkModeColor)
-        XCTAssertEqual(color.resolvedColor(with: UITraitCollection(userInterfaceStyle: .light)), lightModeColor)
-        XCTAssertEqual(color.resolvedColor(with: UITraitCollection(userInterfaceStyle: .dark)), darkModeColor)
+        let color2 = UIColor(light: lightModeColor, dark: nil)
```
guykogus, line 22: Check it in the `else` case too

VincentSit: Done.

Sources/SwifterSwift/UIKit/UIColorExtensions.swift
```diff
@@ -17,9 +17,13 @@ public extension UIColor {
...
-    convenience init(light: UIColor, dark: UIColor) {
-        self.init(dynamicProvider: { $0.userInterfaceStyle == .dark ? dark : light })
+    convenience init(light: UIColor, dark: UIColor?) {
+        if #available(iOS 13.0, tvOS 13.0, *),
+            let dark = dark {
+            self.init(dynamicProvider: { $0.userInterfaceStyle == .dark ? dark : light })
+        } else {
+            self.init(cgColor: light.cgColor)
+        }
     }
```
rypyak, line 27:
```suggestion
    convenience init(light: UIColor) {
        self.init(cgColor: light.cgColor)
    }
    
    convenience init(light: UIColor, dark: UIColor) {
        if #available(iOS 13, *) {
            self.init { $0.userInterfaceStyle == .dark ? dark : light }
        } else {
            self.init(cgColor: light.cgColor)
        }
    }
```

rypyak: I believe this would be neater like so ☝️
**Reasons being:**
The `dark: UIColor?` forces you to always provide a `nil` whenever you only want to use the `light:` theme. 
This makes it inconvenient to swap between use cases of just `UIColor(light:)` and both `UIColor(light:,dark:)`.

It is solvable by having a `convenience init(light: UIColor, dark: UIColor? = nil)` allowing you to have simply `UIColor(light: .red)`.
However it still doesn't prevent you from explicitly using *i.e*: `UIColor(light: .red, dark: nil)`.

Having two `convenience init` like above suggested, prevents any unnecessary code being written, as if you don't want the `dark:` you'd simply not declare it. It also removes the necessity of optional binding.

guykogus: The entire purpose of this initialiser is to provide callers with the option of providing 2 colours for light and dark mode. If they didn't want to provide a dark colour, why would they use this initialiser at all?

rypyak: That was my exact thought, @guykogus, the only reason for the second initialiser `UIColor(light:)` was the ease of removing the `dark:` argument on demand without having to rewrite the whole initialiser (for ease), but it’s totally discardable.

But in that case the `dark:` shouldn’t be `Optional` as you said it yourself and I quote.
> If they didn't want to provide a dark colour, why would they use this initialiser at all?

guykogus: I can foresee use cases where `dark` would be optional. If you've ever worked on a template app, where you swap between different colour/asset themes, there are times where in one situation you would have a light/dark colour for one theme but then use the same colour for another. There are probably other use cases, but that's the first that comes into my mind. I really don't see any harm in having `dark` as optional.

rypyak: There might be, I'm not denying such possibility, I am however am unaware of any particular case, if you could, please share more specific details on such cases or reference to some existing discussion/resource on this as I'd love to take a look myself.

If you're talking about the `Asset Catalog` and the ability to create `Color Set`s with **Appearances** either `None`, `Any, Dark` or `Any, Light, Dark`, + (whatever devices) then I believe it'd be a totally different scenario as either you reference those by a fail-able initialiser `UIColor?(named:)` or using drag and drop from the **Color Library** which creates `Color Literal` copies.

guykogus: I was thinking something along the lines of
```Swift
protocol AppTheme {
    var backgroundColor: UIColor { get }
    var darkModeBackgroundColor: UIColor? { get }
}

struct SimpleAppTheme: AppTheme {
    var backgroundColor: UIColor { .white }
    var darkModeBackgroundColor: UIColor? { nil }
}

struct FancyAppTheme: AppTheme {
    var backgroundColor: UIColor { .white }
    var darkModeBackgroundColor: UIColor? { .black }
}

func applyBackgroundColor(theme: AppTheme) {
    backgroundColor = UIColor(light: theme.backgroundColor, dark: theme.darkModeBackgroundColor)
}
```
but I guess it's not really necessary, you could just replace it with:
```Swift
protocol AppTheme {
    var backgroundColor: UIColor { get }
}

struct SimpleAppTheme: AppTheme {
    var backgroundColor: UIColor { .white }
}

struct FancyAppTheme: AppTheme {
    var backgroundColor: UIColor { UIColor(light: .white, dark: .black) }
}

func applyBackgroundColor(theme: AppTheme) {
    backgroundColor = theme.backgroundColor
}
```

Regardless, in general I wish that APIs would be more forgiving with optional values. I often find myself creating extensions to allow them because it simplifies the calls, e.g. `DateFormatter.date(from: String?)`. This is often needed because when you are responding to data coming from the server you need to handle missing information, and passing along optionals can save you from a lot of `if-else` statements. I really see no harm in making `dark` optional here, and it can potentially do good.

rypyak: Exactly @guykogus , It'd be no longer required to provide an explicitly allocated variable for `darkModeBackgroundColor` as it would be handled dynamically using the `UIColor(light:,dark:)`. 
Otherwise just use any other initialiser like `UIColor(red:,green:,blue:,alpha)` instead.

And nothing would prevent you from still using the mentioned `AppTheme` approach ignoring the dynamic provider init altogether.

There's one other reason I'd avoid `Optional` for that `dark:` parameter that I'll formulate later. As I'm currently busy to give a full blown example 🚀

rypyak: Regarding the data coming from server, I do agree, it's useful to have an occasional `Optional` to avoid a Parsing/Decoding exception right away, and handle it somewhere down the execution line.
But when it comes to `UIColor` I don't thing there'd be a valid reason to have `dark:` as `Optional` as it promotes an unintended use of the initialiser. 

The `dynamicProvider: @escaping (UITraitCollection) -> UIColor` returns a non-optional type `UIColor`. And the current behaviour of this init is already somewhat unpredictable, yes it may seem logical to jump back to `light:` value in case the `dark:` isn't supported by `iOS` version. but how would you handle an addition third mode like `gold:` in a future `iOS` version (i.e: `UIColor(light:,dark:,gold:)`)? Would you jump back to the original `light:` value? or maybe `dark:` instead?
Hence their `Color Set` **Appearance** signature is either `None`, `Any, Dark` or `Any, Light, Dark`, future-proofing potential additions of new modes if necessary.

One possibility would be to replicate the `Color Set` and have:
* `UIColor(none:)` (totally discardable as doesn't do anything new)
* `UIColor(any:,dark:)`
* `UIColor(any:,light:,dark:)` 
Then no matter what it'd quite clear which colour is being taken when the mode is not supported.
This would allow easy future extension without requirement of modification of existing `init`s.

guykogus: OK, I think I'm in agreement to not let `dark` be optional.

As for the better initialisers - I totally agree, we should be 100% replicating the way it's done in `Color Set`. For now we don't want any breaking changes, and I like that the purpose of this PR is just to move the remove the version restriction for callers, so in a future PR we can deprecate this function and replace it with the second 2 initialisers that you suggested (or probably even do it in 1).

@VincentSit tl;dr can you please make `dark` non-optional again?

VincentSit: @guykogus Done.

LucianoPAlmeida:
I didn't participate on the discussion, so I'll defer the second approval to @rypyak :)

LucianoPAlmeida:
Thanks all :)

PR #851, opened 2020-05-25
Added WKWebView extensions

hamtiko:
I have copied old `UIWebView` extensions and added them as a `WKWebview` extensions with small compatibility changes.

## Checklist


hamtiko:
It seems there is some problem with `Swiftlint` check. I have tried to rerun but it fails again.

It is failing with following message:

```
No runner matching the specified labels was found: macos.
```

I am assuming GitHub removes `macos` yaml label for runners.

LucianoPAlmeida:
I tried to adjust, can you rebase with master and try again :)

Sources/SwifterSwift/WebKit/WKWebViewExtensions.swift
```diff
@@ -0,0 +1,32 @@
...
+
+    /// SwifterSwift: Navigate to url using `String`.
+    /// - Parameter urlString: The string specifying the URL to navigate to.
+    /// - Returns: A new navigation for given `urlString`.
+    @discardableResult
+    public func loadURLString(_ urlString: String) -> WKNavigation? {
+        guard let url = URL(string: urlString) else { return nil }
+        return load(URLRequest(url: url))
+    }
+}
```
guykogus, line 31: Add an empty line above

Tests/WebKitTest/WKWebViewExtensionsTests.swift
```diff
@@ -0,0 +1,67 @@
...
+
+    func testLoadURL() {
+        let url = URL(string: "https://www.w3schools.com/")!
+        webView.loadURL(url)
+
+        DispatchQueue.main.asyncAfter(deadline: .now() + 2.0) {
+            guard self.webView.url != nil else {
+                XCTFail("No URL in webView")
+                return
+            }
```
guykogus, line 31: You still need to fulfil the expectation so that the test knows to continue. Just assert that the `url` is not `nil`.
```suggestion
            XCTAssertNotNil(self.webView.url)
```

guykogus: The same comment applies to the other tests.

LucianoPAlmeida: +1

hamtiko:
@guykogus thanks for you comments. 

I have refactored tests based on your suggestions.

SwifterSwift.xcodeproj/project.pbxproj
```diff
@@ -1131,6 +1138,7 @@
 				748B191B23B4F4DA0030FABB /* StoreKitTests */,
 				077BA0931F6BE93000D9C4AC /* SwiftStdlibTests */,
 				07C50D0D1F5EB03200F46E5A /* UIKitTests */,
+				985C1B9124797911008AAB0E /* WebKitTest */,
```
guykogus, line 1141: Rename the folder to `WebKitTests`

hamtiko:
@guykogus done. I have increased test expectation timeout as well.

hamtiko:
It seems 3 seconds of timeout is not enough. What you will suggest? Is it worth to increase timeout more?

guykogus:
Maybe use a simpler website to load, like http://example.com. Although I wonder if it will fail because it's not https 🤔 
I think it might be enough to consider the expectation fulfilled as soon as `webView(_:didStartProvisionalNavigation:)` is called. You don't need to wait for the entire website to finish loading. Check that this fails when you have a dead URL, like https://invalidurl.com.
At the same time, you should respond to the `WKNavigationDelegate` error callbacks to fail on errors and fulfil the expectation.

hamtiko:
@guykogus When I am trying to add test for dead url the `webView(_:didStartProvisionalNavigation:)` delegate function is calling and expectation fulfills before getting fail function called.

Is it ok? or you suggest to go back to use `webView(_:didFinish:)` function?

guykogus:
> @guykogus When I am trying to add test for dead url the `webView(_:didStartProvisionalNavigation:)` delegate function is calling and expectation fulfills before getting fail function called.
> 
> Is it ok? or you suggest to go back to use `webView(_:didFinish:)` function?

@hamtiko you should actually be waiting for a failure expectation in this case. I think what might be better is to use custom `XCTestExpectation` classes. Try using these:

```Swift
class WebViewSuccessExpectation: XCTestExpectation, WKNavigationDelegate {
    init(description: String, webView: WKWebView) {
        super.init(description: description)
        webView.navigationDelegate = self
    }

    func webView(_ webView: WKWebView, didCommit navigation: WKNavigation!) {
        fulfill()
    }
}

class WebViewFailureExpectation: XCTestExpectation, WKNavigationDelegate {
    init(description: String, webView: WKWebView) {
        super.init(description: description)
        webView.navigationDelegate = self
    }

    func webView(_ webView: WKWebView, didFail navigation: WKNavigation!, withError error: Error) {
        fulfill()
    }

    func webView(_ webView: WKWebView, didFailProvisionalNavigation navigation: WKNavigation!, withError error: Error) {
        fulfill()
    }
}
```
This way each test can use the appropriate expectation. Notice that for the success I changed to using `webView(_:didCommit:)`, because as you pointed out even on failure `didStartProvisionalNavigation` is called. This way we don't wait for the entire website to load, we just wait till loading starts to succeed.

hamtiko:
@guykogus Great idea! Thanks for your advice.

Issue #850, opened 2020-05-22
Request add the calculate text size method to UILabel


tospery:
Hope the UILabelExtensions can add the method, that support to calculate text size.
Like TTTAttributedLabel's sizeThatFitsAttributedString:withConstraints:limitedToNumberOfLines function

daliadanila:
@tospery you mean something like this?

let string = //some long string

let attributes = [NSAttributedString.Key.foregroundColor: UIColor.red, NSAttributedString.Key.font: UIFont.boldSystemFont(ofSize: 16) ]

let attributedString = NSAttributedString(string: string, attributes: attributes)

let label = UILabel()

label.numberOfLines = 20

label.attributedText = attributedString

let singleLineSize = UILabel.size(thatFitsAttributedString: attributedString, withConstraints: CGSize(width: 100, height: CGFloat.greatestFiniteMagnitude))

label.backgroundColor = UIColor.white

label.frame = CGRect.init(x: 0, y: 0, width: singleLineSize.width, height: singleLineSize.height)

let testView = UIView.init(frame: CGRect.init(x: 0, y: 0, width: 500, height: singleLineSize.height))

testView.backgroundColor = UIColor.blue

testView.addSubview(label)

<img width="512" alt="Screenshot 2020-05-27 at 15 53 44" src="https://user-images.githubusercontent.com/4118291/83036228-4387be80-a032-11ea-88ca-b3014530f2ec.png">

tospery:
Yes, I mean that

daliadanila:
@tospery cool, I’ll make a PR shortly

tospery:
I think it's better to pass in UILabel's numberOfLines with the limitedToNumberOfLines parameter of the method

tospery:
You can refer to the realization of the OC version: https://github.com/TTTAttributedLabel/TTTAttributedLabel/blob/master/TTTAttributedLabel/TTTAttributedLabel.m（TTTAttributedLabel->sizeThatFitsAttributedString:withConstraints:limitedToNumberOfLines）

daliadanila:
I had a look, but there doesn't seem to be an equivalent in Swift for `CTFramesetterSuggestFrameSizeForAttributedStringWithConstraints` or a similar method in CT that receives the number of lines.

tospery:
CTFramesetterSuggestFrameSizeForAttributedStringWithConstraints is not a system method, you can see its implementation

[...]

PR #853, opened 2020-05-28
UILabel extension to calculate text size

daliadanila:
Added UILabel extension for calculating the size that best fits an attributed string #850 

## Checklist

### DIFF
--- a/Examples/Examples.playground/Pages/03-UIKitExtensions.xcplaygroundpage/Contents.swift
+++ b/Examples/Examples.playground/Pages/03-UIKitExtensions.xcplaygroundpage/Contents.swift
@@ -102,4 +102,32 @@ view.cornerRadius = 30
 // Add shadow to view
 view.addShadow(ofColor: .black, radius: 3, opacity: 0.5)
 
+//: UILabel extension
+
+let string = "Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur."
+
+let attributes = [NSAttributedString.Key.foregroundColor: UIColor.red, NSAttributedString.Key.font: UIFont.boldSystemFont(ofSize: 16) ]
+
+let attributedString = NSAttributedString(string: string, attributes: attributes)
+
+let label = UILabel()
+
+label.numberOfLines = 20
+
+label.attributedText = attributedString
+
+let singleLineSize = UILabel.size(thatFitsAttributedString: attributedString, withConstraints: CGSize(width: 100, height: CGFloat.greatestFiniteMagnitude))
+
+label.backgroundColor = UIColor.white
+
+label.frame = CGRect.init(x: 0, y: 0, width: singleLineSize.width, height: singleLineSize.height)
+
+let testView = UIView.init(frame: CGRect.init(x: 0, y: 0, width: 500, height: singleLineSize.height))
+
+testView.backgroundColor = UIColor.blue
+
+testView.addSubview(label)
+
+
 //: [Next](@next)
+//
--- a/Examples/Examples.playground/Pages/03-UIKitExtensions.xcplaygroundpage/Contents.swift
+++ b/Examples/Examples.playground/Pages/03-UIKitExtensions.xcplaygroundpage/Contents.swift
@@ -128,6 +128,5 @@ testView.backgroundColor = UIColor.blue
 
 testView.addSubview(label)
 
-
 //: [Next](@next)
 //
--- a/Sources/SwifterSwift/UIKit/UILabelExtensions.swift
+++ b/Sources/SwifterSwift/UIKit/UILabelExtensions.swift
@@ -38,7 +38,6 @@ public extension UILabel {
     ///   - attributedString: The attributed string.
     ///   - size: The maximum dimensions used to calculate size.
     /// - Returns: The size that fits the attributed string within the specified constraints.
-    @available(iOS 10, *)
     static func size(thatFitsAttributedString attributedString: NSAttributedString,
                      withConstraints size: CGSize) -> CGSize {
 
--- a/Sources/SwifterSwift/UIKit/UILabelExtensions.swift
+++ b/Sources/SwifterSwift/UIKit/UILabelExtensions.swift
@@ -38,7 +38,7 @@ public extension UILabel {
     ///   - attributedString: The attributed string.
     ///   - size: The maximum dimensions used to calculate size.
     /// - Returns: The size that fits the attributed string within the specified constraints.
-    
+    @available(iOS 10, *)
     static func size(thatFitsAttributedString attributedString: NSAttributedString,
                      withConstraints size: CGSize) -> CGSize {
         
--- a/Sources/SwifterSwift/UIKit/UILabelExtensions.swift
+++ b/Sources/SwifterSwift/UIKit/UILabelExtensions.swift
@@ -11,7 +11,7 @@ import UIKit
 
 // MARK: - Properties
 public extension UILabel {
-    
+
     /// SwifterSwift: Required height for a label
     var requiredHeight: CGFloat {
         let label = UILabel(frame: CGRect(x: 0, y: 0, width: frame.width, height: CGFloat.greatestFiniteMagnitude))
@@ -27,7 +27,7 @@ public extension UILabel {
 
 // MARK: - Methods
 public extension UILabel {
-    
+
     /// SwifterSwift:  Calculate and return the size that best fits an attributed string, given the specified constraints on size
     ///
     ///   let yourAttributedString = ...
@@ -41,34 +41,32 @@ public extension UILabel {
     @available(iOS 10, *)
     static func size(thatFitsAttributedString attributedString: NSAttributedString,
                      withConstraints size: CGSize) -> CGSize {
-        
+
         guard attributedString.length > 0 else { return CGSize.zero}
-        
+
         let framesetter = CTFramesetterCreateWithAttributedString(attributedString)
-        
+
         let calculatedSize = CTFramesetterSuggestFrameSizeWithConstraints(framesetter,
-                                                                          CFRange(location: 0,length: attributedString.length),
+                                                                          CFRange(location: 0, length: attributedString.length),
                                                                           nil,
                                                                           size,
                                                                           nil)
-        
+
         return calculatedSize
-        
-        
-        
+
     }
 }
 
 // MARK: - Initializers
 
 public extension UILabel {
-    
+
     /// SwifterSwift: Initialize a UILabel with text
     convenience init(text: String?) {
         self.init()
         self.text = text
     }
-    
+
     /// SwifterSwift: Initialize a UILabel with a text and font style.
     ///
     /// - Parameters:
--- a/Sources/SwifterSwift/UIKit/UILabelExtensions.swift
+++ b/Sources/SwifterSwift/UIKit/UILabelExtensions.swift
@@ -9,7 +9,6 @@
 #if canImport(UIKit) && !os(watchOS)
 import UIKit
 
-
 // MARK: - Properties
 public extension UILabel {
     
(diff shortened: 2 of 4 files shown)


Sources/SwifterSwift/UIKit/UILabelExtensions.swift
```diff
@@ -9,7 +9,56 @@
 #if canImport(UIKit) && !os(watchOS)
 import UIKit
 
+// MARK: - Properties
+public extension UILabel {
+
+    /// SwifterSwift: Required height for a label
+    var requiredHeight: CGFloat {
+        let label = UILabel(frame: CGRect(x: 0, y: 0, width: frame.width, height: CGFloat.greatestFiniteMagnitude))
```
LucianoPAlmeida, line 17: I'm not sure about this... I think creating another label object is not the best way to move this forward.
Is there any way to calculate this only based on text and font size (maybe some other system information e.g. dynamic size)?

Sources/SwifterSwift/UIKit/UILabelExtensions.swift
```diff
@@ -9,7 +9,56 @@
...
+    ///
+    ///   let yourAttributedString = ...
+    ///   let yourConstraintSize = ...
+    ///   let size = UILabel.size(thatFitsAttributedString: yourAttributedString, withConstraints: yourConstraintSize)
+    ///
+    /// - Parameters:
+    ///   - attributedString: The attributed string.
+    ///   - size: The maximum dimensions used to calculate size.
+    /// - Returns: The size that fits the attributed string within the specified constraints.
+    @available(iOS 10, *)
```
LucianoPAlmeida, line 41: The minimum target we have is already 10 I think, so I think we don't need this :)

Sources/SwifterSwift/UIKit/UILabelExtensions.swift
```diff
@@ -9,7 +9,56 @@
...
+    static func size(thatFitsAttributedString attributedString: NSAttributedString,
+                     withConstraints size: CGSize) -> CGSize {
+
+        guard attributedString.length > 0 else { return CGSize.zero}
+
+        let framesetter = CTFramesetterCreateWithAttributedString(attributedString)
```
LucianoPAlmeida, line 47: Could we use this for `requiredHeight: CGFloat`? Or is it not the same?

Sources/SwifterSwift/UIKit/UILabelExtensions.swift
LucianoPAlmeida: Oh I just noted this was already here, sorry 😄

LucianoPAlmeida: We probably need to revisit this extension ... cc @omaralbeik

Examples/Examples.playground/Pages/03-UIKitExtensions.xcplaygroundpage/Contents.swift
```diff
@@ -102,4 +102,31 @@ view.cornerRadius = 30
...
+
+let attributedString = NSAttributedString(string: string, attributes: attributes)
+
+let label = UILabel()
+
+label.numberOfLines = 20
+
+label.attributedText = attributedString
+
+let singleLineSize = UILabel.size(thatFitsAttributedString: attributedString, withConstraints: CGSize(width: 100, height: CGFloat.greatestFiniteMagnitude))
```
LucianoPAlmeida, line 119: @daliadanila 
Throwing an idea here ... what do you think about an API like 
```swift
extension UILabel {
   func adjustToFit(string: String,  limitedBy size: CGSize) -> CGSize { ... }
}
```
Most use cases I think probably will use the size to adjust the label anyway ... 
What do you think?

daliadanila: @LucianoPAlmeida That's how I thought about it initially, but both methods from CoreText use the NSAttributedString that have style info in it.
Unless I'd create a local instance of NSAttributedString that contains all extracted properties from UILabel (font, size, foregroundColor, etc) 🤔

Sources/SwifterSwift/UIKit/UILabelExtensions.swift
daliadanila: Fixed now.

Examples/Examples.playground/Pages/03-UIKitExtensions.xcplaygroundpage/Contents.swift
LucianoPAlmeida: Humm, I thought it was just font, but even in that case we can get this information from the self label object at least font size and colors so we build our own attributed string internally... what else it is used for calculate this? 
And also we can have both API’s string and attract string :)

daliadanila: There are around 30 different properties style related that the attributed string can have.  Found there's a handy property on the label called `attributedText` that does exactly that! 😄
I'll change the extension to match your suggested signature.

LucianoPAlmeida: Let’s wait a bit for others opinions before changing :) what you think?

LucianoPAlmeida: Vc @omaralbeik @guykogus

daliadanila: > Let’s wait a bit for others opinions before changing :) what you think?

No problem 😃

Sources/SwifterSwift/UIKit/UILabelExtensions.swift
omaralbeik: > We probably need to revisit this extension ... cc @omaralbeik

Definitely, there are lots of UIKit extensions that can benefit from a revisit 💯

Sources/SwifterSwift/UIKit/UILabelExtensions.swift
```diff
@@ -9,7 +9,55 @@
...
+    ///   let size = UILabel.size(thatFitsAttributedString: yourAttributedString, withConstraints: yourConstraintSize)
+    ///
+    /// - Parameters:
+    ///   - attributedString: The attributed string.
+    ///   - size: The maximum dimensions used to calculate size.
+    /// - Returns: The size that fits the attributed string within the specified constraints.
+    static func size(thatFitsAttributedString attributedString: NSAttributedString,
+                     withConstraints size: CGSize) -> CGSize {
+
+        guard attributedString.length > 0 else { return CGSize.zero}
```
omaralbeik, line 44: zero

Sources/SwifterSwift/UIKit/UILabelExtensions.swift
```diff
@@ -9,7 +9,55 @@
...
+
+        let framesetter = CTFramesetterCreateWithAttributedString(attributedString)
+
+        let calculatedSize = CTFramesetterSuggestFrameSizeWithConstraints(framesetter,
+                                                                          CFRange(location: 0, length: attributedString.length),
+                                                                          nil,
+                                                                          size,
+                                                                          nil)
```
omaralbeik, line 52: If we extract the range out, we can fill the function in one line

```swift
let range = CFRange(location: 0, length: attributedString.length)
return CTFramesetterSuggestFrameSizeWithConstraints(framesetter, range, nil, size, nil)
```

Sources/SwifterSwift/UIKit/UILabelExtensions.swift
```diff
@@ -9,7 +9,55 @@
...
+    ///   - size: The maximum dimensions used to calculate size.
+    /// - Returns: The size that fits the attributed string within the specified constraints.
+    static func size(thatFitsAttributedString attributedString: NSAttributedString,
+                     withConstraints size: CGSize) -> CGSize {
+
+        guard attributedString.length > 0 else { return CGSize.zero}
+
+        let framesetter = CTFramesetterCreateWithAttributedString(attributedString)
+
+        let calculatedSize = CTFramesetterSuggestFrameSizeWithConstraints(framesetter,
```
omaralbeik, line 48: Since this is coming from CoreText, please wrap the function with a `canImport` check

```swift
public extension UILabel {
    #if canImport(CoreText)
    func ...
    #endif
}
```

Sources/SwifterSwift/UIKit/UILabelExtensions.swift
```diff
@@ -9,7 +9,55 @@
...
+    ///   let yourAttributedString = ...
+    ///   let yourConstraintSize = ...
+    ///   let size = UILabel.size(thatFitsAttributedString: yourAttributedString, withConstraints: yourConstraintSize)
+    ///
+    /// - Parameters:
+    ///   - attributedString: The attributed string.
+    ///   - size: The maximum dimensions used to calculate size.
+    /// - Returns: The size that fits the attributed string within the specified constraints.
+    static func size(thatFitsAttributedString attributedString: NSAttributedString,
+                     withConstraints size: CGSize) -> CGSize {
```
omaralbeik, line 42: How about the following function signature?
```swift
static func size(fitting attributedString: NSAttributedString, limitedBy size: CGSize) -> CGSize
```

Examples/Examples.playground/Pages/03-UIKitExtensions.xcplaygroundpage/Contents.swift
omaralbeik: @LucianoPAlmeida I see your point, tbh I prefer a function that does not change the frame. but return the required size instead

guykogus: I'm just not sure I understand the purpose of this. View's have `sizeThatFits(_:)` and `sizeToFit()` functions. Why not just set the `attributedText` of a label and call one of those? Is your purpose to avoid creating a `UILabel`?

LucianoPAlmeida: @guykogus TBH I really don’t know, I personally never use something like this, but seem some people use something like this, so I figure it may be useful for some case :)

omaralbeik: @guykogus I'm not sure as well, but I remember having to deal with layouting `UILabels` a couple of years ago in one app where `sizeThatFits` and `sizeToFit` was very buggy in complex `UICollectionView` layouts

@daliadanila could you please explain where you use this extension? what benefits does it provide over `sizeThatFits` and `sizeToFit`?

LucianoPAlmeida: @tospery suggested this on the issue, maybe he can provide some input here too :)

daliadanila: @guykogus @omaralbeik @LucianoPAlmeida  I'd think people might find it useful when they'd have to work with strings/labels that have complex styling. In the original issue that was requesting this feature, [TTTAttributedLabel](https://github.com/TTTAttributedLabel/TTTAttributedLabel/blob/master/TTTAttributedLabel/TTTAttributedLabel.m) was mentioned that is implementing a similar method 
（sizeThatFitsAttributedString:withConstraints:limitedToNumberOfLines. 

Maybe @tospery can help with a use case when you'd use the CT methods over sizeThatFits:/sizeToFit methods.

daliadanila:
> Nice work @daliadanila, please refer to my comments in code.

Thank you for the review! 🙏

Examples/Examples.playground/Pages/03-UIKitExtensions.xcplaygroundpage/Contents.swift
guykogus: `TTTAttributedLabel` was created before you could assign an attributed string to a label. If you're still using it for any purpose then for the love of all that's holy please stop! 😅

LucianoPAlmeida: >  I'd think people might find it useful when they'd have to work with strings/labels that have complex styling.

I think if `sizeToFit()` and `sizeThatFits` works in most cases and this is more used for very specific scenarios and complex things, I'm not sure we should add this here ... I mean we would increase the library with an addition that would be rarely be used if used... 
Let me know what you think @daliadanila @guykogus @omaralbeik :)

LucianoPAlmeida: Just making sure, everyone agrees or are we moving forward with this? :)

LucianoPAlmeida:
@daliadanila Thank you for taking the time to put this up and collaboration on the discussion 💯 
But as per what was discussed we are closing this. But feel free to reopen this if you want to restart discussion.
Also, If you still feel like contributing to SwifterSwift, take a look at our issues with help wanted and good first issue labels :)

PR #854, opened 2020-05-31
Added extensions to DispatchQueue, String, Optional

VatoKo:
## Checklist

### DIFF
--- a/Sources/SwifterSwift/Dispatch/DispatchQueueExtensions.swift
+++ b/Sources/SwifterSwift/Dispatch/DispatchQueueExtensions.swift
@@ -41,6 +41,34 @@ public extension DispatchQueue {
 
         return DispatchQueue.getSpecific(key: key) != nil
     }
+    
+    /// SwifterSwift: Runs passed closure on main thread.
+    ///
+    /// - Parameter block: The closure to run on main thread
+    static func asyncOnMain(code block: @escaping () -> Void) {
+        DispatchQueue.main.async {
+            block()
+        }
+    }
+    
+    /// SwifterSwift: Runs passed closure on background thread.
+    ///
+    /// - Parameter block: The closure to run on background thread
+    static func asyncOnBg(code block: @escaping () -> Void) {
+        DispatchQueue.global().async {
+            block()
+        }
+    }
+    
+    /// SwifterSwift: Runs passed closure asynchronous after certain time interval
+    ///
+    /// - Parameter time: The time inverval after which the closure will run.
+    /// - Parameter block: The closure to run after certain time interval.
+    static func asyncAfter(time: TimeInterval, code block: @escaping () -> Void) {
+        DispatchQueue.main.asyncAfter(deadline: .now() + time) {
+            block()
+        }
+    }
 
 }
 
--- a/Sources/SwifterSwift/SwiftStdlib/StringExtensions.swift
+++ b/Sources/SwifterSwift/SwiftStdlib/StringExtensions.swift
@@ -957,6 +957,19 @@ public extension String {
         return range(of: pattern, options: .regularExpression, range: nil, locale: nil) != nil
     }
     #endif
+    
+    /// SwifterSwift: Overload Swift's 'contains' operator for matching regex pattern
+    ///
+    /// - Parameter lhs: String to check on regex pattern.
+    /// - Parameter rhs: Pattern to verify.
+    /// - Returns: true if string matches the pattern.
+    static func ~= (lhs: String, rhs: String) -> Bool {
+        // https://www.hackingwithswift.com/articles/108/how-to-use-regular-expressions-in-swift
+        
+        guard let regex = try? NSRegularExpression(pattern: rhs) else { return false }
+        let range = NSRange(location: 0, length: lhs.utf16.count)
+        return regex.firstMatch(in: lhs, options: [], range: range) != nil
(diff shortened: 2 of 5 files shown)


Sources/SwifterSwift/Dispatch/DispatchQueueExtensions.swift
```diff
@@ -41,6 +41,34 @@ public extension DispatchQueue {
...
+        DispatchQueue.global().async {
+            block()
+        }
+    }
+    
+    /// SwifterSwift: Runs passed closure asynchronous after certain time interval
+    ///
+    /// - Parameter time: The time inverval after which the closure will run.
+    /// - Parameter block: The closure to run after certain time interval.
+    static func asyncAfter(time: TimeInterval, code block: @escaping () -> Void) {
```
guykogus, line 67: This is the only one of the `DispatchQueue` functions that I think is useful, but only if it were non-static.
```suggestion
   func asyncAfter(time: TimeInterval, execute block: @escaping () -> Void) {
```

Sources/SwifterSwift/SwiftStdlib/StringExtensions.swift
```diff
@@ -957,6 +957,19 @@ public extension String {
         return range(of: pattern, options: .regularExpression, range: nil, locale: nil) != nil
     }
     #endif
+    
+    /// SwifterSwift: Overload Swift's 'contains' operator for matching regex pattern
+    ///
+    /// - Parameter lhs: String to check on regex pattern.
+    /// - Parameter rhs: Pattern to verify.
+    /// - Returns: true if string matches the pattern.
+    static func ~= (lhs: String, rhs: String) -> Bool {
```
guykogus, line 966: We already have this functionality implemented in the `matches(pattern:)` function, so feel free to copy that as it's more concise.
```Swift
return range(of: pattern, options: .regularExpression) != nil
```

You also need to wrap the function in `#if canImport(Foundation)`, since you're using `Foundation` classes/functions for it like in other functions. Don't forget to do the same for the unit tests.

LucianoPAlmeida:
As @guykogus pointed the Optional extension isn't really useful since we already have `map`. 
But I think it would be nice to put the String and Dispatch one in different PR's :)

Sources/SwifterSwift/Dispatch/DispatchQueueExtensions.swift
rypyak, line 67: I like this idea. totally better than the `.now() + 123` syntax.

I wonder what do you think about an extension on `DispatchTime` instead?
like you have `.now()` there could be a `.wait(_:)` (or any other name such as `.after(_:)` or `.delay(_:)`)
```
extension DispatchTime {

    static func wait(_ interval: TimeInterval) -> Self { .now() + interval }
}
```

resulting in syntax as:
```
DispatchQueue.main.asyncAfter(deadline: .wait(10)) {
    // Do stuff...
}
```

VatoKo:
Thank you very much for your feedback, I'll take all that into account and will make a new PR.

Sources/SwifterSwift/Dispatch/DispatchQueueExtensions.swift
guykogus: I don't really see much benefit in replacing `.now() + delay` with `.wait(delay)`. You're just replacing one function call with another. A cleaner option would actually be to extend `DispatchTime` with `ExpressibleByIntegerLiteral` and `ExpressibleByFloatLiteral`.

```Swift
public extension DispatchTime: ExpressibleByIntegerLiteral {
    public init(integerLiteral value: Int) {
        self.init(uptimeNanoseconds: (.now() + .seconds(value)).uptimeNanoseconds)
    }
}

extension DispatchTime: ExpressibleByFloatLiteral {
    public init(floatLiteral value: TimeInterval) {
        self.init(uptimeNanoseconds: (.now() + value).uptimeNanoseconds)
    }
}
```
The first problem is that it's semantically wrong.
There's also the limitation here is that you can't actually use constant values, only literals. I.e.
```Swift
DispatchQueue.main.asyncAfter(deadline: 10) { ... } // OK
DispatchQueue.main.asyncAfter(deadline: delay) { ... } // Cannot convert value of type 'Int' to expected argument type 'DispatchTime'
```
I think the currently suggested extension (`dispatchAfter(delay:...)`) is better, although I'd recommend making it more complete:
```Swift
public extension DispatchQueue {
    func asyncAfter(delay: Double,
                    qos: DispatchQoS = .unspecified,
                    flags: DispatchWorkItemFlags = [],
                    execute work: @escaping @convention(block) () -> Void) {
        asyncAfter(deadline: .now() + delay, qos: qos, flags: flags, execute: work)
    }
}
```

rypyak: I agree, having a direct `TimeInterval` is much easier to use, let’s keep it as you proposed.

I just originally thought that creating a whole new `DispatchQueue` function to fix **one of the parameters** would be a little too much. 
With hope that `DispatchTime` function would allow for more granular, composable and reusable solution, not limited to `DispatchQueue` case. but as you said it’s totally just replacing one function call to another. 🚀 Plus it’s probably only ever is used in combination of both.

guykogus:
Closing, as this has been split into #858 and #859

PR #860, opened 2020-06-04
Added new protocol for UIViewController to instantiate it from storyboard

VatoKo:
## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 
 ### Added
 - **UIViewController**:
--   Added  `instantiate(from:bundle:identifier:)` function to  `UIViewController` to make it easier to instantiate it from storyboard. [#860](https://github.com/SwifterSwift/SwifterSwift/pull/860) by [VatoKo](https://github.com/VatoKo)
+  - Added `instantiate(from:bundle:identifier:)` function to `UIViewController` to make it easier to instantiate it from storyboard. [#860](https://github.com/SwifterSwift/SwifterSwift/pull/860) by [VatoKo](https://github.com/VatoKo)
 -**String**:
   - Overloaded Swift's 'contains' operator (`~=`) for `String` to check regex matching. [#858](https://github.com/SwifterSwift/SwifterSwift/pull/858) by [VatoKo](https://github.com/VatoKo)
 -**DispatchQueue**:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 ### Added
 - **UIViewController**:
 -   Added  `instantiate(from:bundle:identifier:)` function to  `UIViewController` to make it easier to instantiate it from storyboard. [#860](https://github.com/SwifterSwift/SwifterSwift/pull/860) by [VatoKo](https://github.com/VatoKo)
+-**String**:
+  - Overloaded Swift's 'contains' operator (`~=`) for `String` to check regex matching. [#858](https://github.com/SwifterSwift/SwifterSwift/pull/858) by [VatoKo](https://github.com/VatoKo)
+-**DispatchQueue**:
+- Added `asyncAfter(delay:qos:flags:execute:)` method to use it more conveniently without typing unnecessary `.now() + delay`. [#859](https://github.com/SwifterSwift/SwifterSwift/pull/859) by [VatoKo](https://github.com/VatoKo)
 - **RangeReplaceableCollection**:
   - `subscript(offset:)` and `subscript(range:)` to access and replace elements by the index offsets. [#826](https://github.com/SwifterSwift/SwifterSwift/pull/826) by [guykogus](https://github.com/guykogus)
 - **Sequence**:
@@ -30,6 +34,8 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 ### Removed
 
 ### Fixed
+- **CAGradientLayerExtensions.swift**:
+- CAGradientLayer extensions inaccessible through internal level protection. [#856](https://github.com/SwifterSwift/SwifterSwift/pull/856) by [Den Andreychuk](https://github.com/denandreychuk).
 
 ### Security
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 ### Added
 - **UIViewController**:
   - Added `instantiate(from:bundle:identifier:)` function to `UIViewController` to make it easier to instantiate it from storyboard. [#860](https://github.com/SwifterSwift/SwifterSwift/pull/860) by [VatoKo](https://github.com/VatoKo)
--**String**:
+- **String**:
   - Overloaded Swift's 'contains' operator (`~=`) for `String` to check regex matching. [#858](https://github.com/SwifterSwift/SwifterSwift/pull/858) by [VatoKo](https://github.com/VatoKo)
 -**DispatchQueue**:
 - Added `asyncAfter(delay:qos:flags:execute:)` method to use it more conveniently without typing unnecessary `.now() + delay`. [#859](https://github.com/SwifterSwift/SwifterSwift/pull/859) by [VatoKo](https://github.com/VatoKo)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
   - `center`, `topLeft`, `topRight`, `bottomLeft`, `bottomRight` to get anchor position or set position using anchor. [#835](https://github.com/SwifterSwift/SwifterSwift/pull/835) by [rypyak](https://github.com/rypyak)
 - **CGSize**:
   - Added new `+`, `+=`, `-` and `-=` operator extensions for tuple (width: CGFloat, height: CGFloat). [#841](https://github.com/SwifterSwift/SwifterSwift/pull/841) by [hamtiko](https://github.com/hamtiko)
+- **WKWebView**:
+  - Added `loadURL(_:)` and `loadURLString(_:)` extensions for `WkWebView`. [#851](https://github.com/SwifterSwift/SwifterSwift/pull/851) by [hamtiko](https://github.com/hamtiko)
 
 ### Changed
 - **NSAttributedStringExtensions.swift**:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 ### Changed
 - **NSAttributedStringExtensions.swift**:
   - `applying(attributes:)` changed access modifier from `fileprivate` to `public`. [#832](https://github.com/SwifterSwift/SwifterSwift/pull/832) by [cHaLkdusT](https://github.com/cHaLkdusT)
+- **Color**:
+  - Refactored `init(light:dark:)` to remove deployment target version restrictions. [#844](https://github.com/SwifterSwift/SwifterSwift/pull/844) by [VincentSit](https://github.com/vincentsit).
 
 ### Deprecated
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 ## Upcoming Release
 
 ### Added
+- **UIViewController**:
+-   Added  `instantiate(from:bundle:identifier:)` function to  `UIViewController` to make it easier to instantiate it from storyboard. [#860](https://github.com/SwifterSwift/SwifterSwift/pull/860) by [VatoKo](https://github.com/VatoKo)
 - **RangeReplaceableCollection**:
   - `subscript(offset:)` and `subscript(range:)` to access and replace elements by the index offsets. [#826](https://github.com/SwifterSwift/SwifterSwift/pull/826) by [guykogus](https://github.com/guykogus)
 - **Sequence**:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,8 +9,8 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
   - Added `instantiate(from:bundle:identifier:)` function to `UIViewController` to make it easier to instantiate it from storyboard. [#860](https://github.com/SwifterSwift/SwifterSwift/pull/860) by [VatoKo](https://github.com/VatoKo)
 - **String**:
   - Overloaded Swift's 'contains' operator (`~=`) for `String` to check regex matching. [#858](https://github.com/SwifterSwift/SwifterSwift/pull/858) by [VatoKo](https://github.com/VatoKo)
--**DispatchQueue**:
-- Added `asyncAfter(delay:qos:flags:execute:)` method to use it more conveniently without typing unnecessary `.now() + delay`. [#859](https://github.com/SwifterSwift/SwifterSwift/pull/859) by [VatoKo](https://github.com/VatoKo)
+- **DispatchQueue**:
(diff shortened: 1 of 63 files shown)


VatoKo:
Here I thought about making extension to UIViewController directly, but then every time you instantiated a custom UIViewController, you would have to cast it to your type as the return type for the function would be 'UIViewController'. In this way that casting won't be necessary.

Sources/SwifterSwift/UIKit/UIViewControllerExtensions.swift
```diff
@@ -9,6 +9,33 @@
 #if canImport(UIKit) && !os(watchOS)
 import UIKit
 
+// MARK: - Helper Protocols
+protocol Storyboarded {
```
LucianoPAlmeida, line 13: We have kind of a "rule" against creating new types, protocols, and stuff... to ensure we keep this simple and only quick and helpful extensions that can be copy/pasted. That one of the things that allow this library not to blow out of proportion :)

VatoKo: I see. Understood

VatoKo:
Okay, I'll try. Thanks 😊

VatoKo:
@guykogus I updated the code the way you suggested.

Sources/SwifterSwift/UIKit/UIViewControllerExtensions.swift
```diff
@@ -22,6 +22,23 @@ public extension UIViewController {
...
+    /// - Parameters:
+    ///   - storyboard: Name of the storyboard where the UIViewController is located
+    ///   - bundle: Bundle in which storyboard is located
+    ///   - identifier: UIViewController's storyboard identifier
+    /// - Returns: Custom UIViewController instantiated from storyboard
+    class func instantiate(from storyboard: String = "Main", bundle: Bundle? = nil, identifier: String? = nil) -> Self {
+        let viewControllerIdentifier = identifier ?? String(describing: self)
+        let storyboard = UIStoryboard(name: storyboard, bundle: bundle)
+        let vc = storyboard.instantiateViewController(withIdentifier: viewControllerIdentifier) as! Self
+        return vc
```
guykogus, line 40:
```suggestion
        return storyboard.instantiateViewController(withIdentifier: viewControllerIdentifier) as! Self // swiftlint:disable:this force_cast
```

Sources/SwifterSwift/UIKit/UIViewControllerExtensions.swift
```diff
@@ -22,6 +22,23 @@ public extension UIViewController {
 
 // MARK: - Methods
 public extension UIViewController {
+    
+    
+    /// SwifterSwift: Instantiate UIViewController from storyboard
+    ///
+    ///     let myViewController = MyViewController.instantiate()
```
guykogus, line 29: I don't think this example is necessary.

LucianoPAlmeida:
@VatoKo There are a couple of swiftlint issues which needs to be addressed

Tests/UIKitTests/UIViewControllerExtensionsTests.swift
```diff
@@ -54,6 +54,18 @@ final class UIViewControllerExtensionsTests: XCTestCase {
...
     }
+    
+    class MyViewController: UIViewController {
+        var foo: String?
+    }
+    
+    func testInstantiate() {
+        let bar = "Bar"
+        let vc = MyViewController.instantiate(from: "TestStoryboard")
+        XCTAssertNotNil(vc)
```
guykogus, line 65: I'm not sure you can even get a `nil` result, so this test doesn't seem necessary.

Tests/UIKitTests/UIViewControllerExtensionsTests.swift
```diff
@@ -54,6 +54,18 @@ final class UIViewControllerExtensionsTests: XCTestCase {
         NotificationCenter.default.post(name: notificationIdentifier, object: nil)
         XCTAssertFalse(viewController.notificationFired)
     }
+    
+    class MyViewController: UIViewController {
+        var foo: String?
+    }
```
guykogus, line 60: Move this declaration to the top of the class, under `MockNotificationViewController`.

Tests/UIKitTests/UIViewControllerExtensionsTests.swift
```diff
@@ -54,6 +54,18 @@ final class UIViewControllerExtensionsTests: XCTestCase {
...
+    
+    func testInstantiate() {
+        let bar = "Bar"
+        let vc = MyViewController.instantiate(from: "TestStoryboard")
+        XCTAssertNotNil(vc)
+        vc.foo = bar
+        XCTAssertEqual(vc.foo, bar)
```
guykogus, line 67: It feels like you're doing this to test that you have a `MyViewController` object. Just check
```Swift
XCTAssert(vc is MyViewController)
```
You can also then remove the `foo` property from the class, as it's not really necessary.

LucianoPAlmeida: But based on the method signature, this will be always a `MyViewController` so not sure this assertion would do something ... Maybe we could do two tests here one like this asserting no throw and one where we try to instantiate a view controller that isn't on the storyboard so we expect a force unwrap error or an assertion. What you think @guykogus ?

Sources/SwifterSwift/UIKit/UIViewControllerExtensions.swift
```diff
@@ -22,6 +22,20 @@ public extension UIViewController {
...
+    ///
+    /// - Parameters:
+    ///   - storyboard: Name of the storyboard where the UIViewController is located
+    ///   - bundle: Bundle in which storyboard is located
+    ///   - identifier: UIViewController's storyboard identifier
+    /// - Returns: Custom UIViewController instantiated from storyboard
+    class func instantiate(from storyboard: String = "Main", bundle: Bundle? = nil, identifier: String? = nil) -> Self {
+        let viewControllerIdentifier = identifier ?? String(describing: self)
+        let storyboard = UIStoryboard(name: storyboard, bundle: bundle)
+        return storyboard.instantiateViewController(withIdentifier: viewControllerIdentifier) as! Self
```
LucianoPAlmeida, line 37: Could we throw a `preconditionFailure` here if this`storyboard.instantiateViewController(withIdentifier: viewControllerIdentifier)` method returns nil, I think will be good because we can have a message explaining the reasons why this may be nil. What do you think @guykogus?

Tests/UIKitTests/UIViewControllerExtensionsTests.swift
LucianoPAlmeida: +1

Sources/SwifterSwift/UIKit/UIViewControllerExtensions.swift
guykogus: Yeah, I think that'll be clearer than crashing on a forced cast.

VatoKo: @LucianoPAlmeida The thing is `storyboard.instantiateViewController(withIdentifier: viewControllerIdentifier)` doesn't return `Optional` and comparing the return value to `nil` will always result `false`

guykogus: You can do it like this:
```Swift
guard let viewController = storyboard.instantiateViewController(withIdentifier: viewControllerIdentifier) as? Self else {
    preconditionFailure("Unable to instantiate view controller with identifier \(viewControllerIdentifier) as type \(String(describing: Self))")
}
return viewController
```

Sources/SwifterSwift/UIKit/UIViewControllerExtensions.swift
```diff
@@ -22,6 +22,23 @@ public extension UIViewController {
 
 // MARK: - Methods
 public extension UIViewController {
+    
+
```
guykogus, line 26: Remove this extra line

Tests/UIKitTests/UIViewControllerExtensionsTests.swift
```diff
@@ -55,6 +59,14 @@ final class UIViewControllerExtensionsTests: XCTestCase {
         XCTAssertFalse(viewController.notificationFired)
     }
 
+    func testInstantiate() {
+        let bar = "Bar"
+        let myViewController = MyViewController.instantiate(from: "TestStoryboard")
+        XCTAssert(myViewController is MyViewController)
+        myViewController.foo = bar
+        XCTAssertEqual(myViewController.foo, bar)
```
guykogus, line 67: Still not sure what tests make sense here. @LucianoPAlmeida @omaralbeik any suggestions?

LucianoPAlmeida: I think we shouldn't assert anything, the test would fail if it crashes... so we should just call the function inside the test and expect the function to return successfully. 
IMO we don't need to assert that :)

LucianoPAlmeida:
Closing due inactivity, @VatoKo feel free to reopen if you still want to move this forward :)

VatoKo:
@LucianoPAlmeida I was waiting for the final decision from you actually 😄I'll make changes according to the way you suggested.

CHANGELOG.md
```diff
@@ -5,7 +5,9 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 ## Upcoming Release
 
 ### Added
-- **String**:
+- **UIViewController**:
+  - Added `instantiate(from:bundle:identifier:)` function to `UIViewController` to make it easier to instantiate it from storyboard. [#860](https://github.com/SwifterSwift/SwifterSwift/pull/860) by [VatoKo](https://github.com/VatoKo)
+-**String**:
```
guykogus, line 10:
```suggestion
- **String**:
```

Sources/SwifterSwift/UIKit/UIViewControllerExtensions.swift
```diff
@@ -23,6 +23,22 @@ public extension UIViewController {
...
+    /// - Parameters:
+    ///   - storyboard: Name of the storyboard where the UIViewController is located
+    ///   - bundle: Bundle in which storyboard is located
+    ///   - identifier: UIViewController's storyboard identifier
+    /// - Returns: Custom UIViewController instantiated from storyboard
+    class func instantiate(from storyboard: String = "Main", bundle: Bundle? = nil, identifier: String? = nil) -> Self {
+        let viewControllerIdentifier = identifier ?? String(describing: self)
+        let storyboard = UIStoryboard(name: storyboard, bundle: bundle)
+        guard let viewController = storyboard.instantiateViewController(withIdentifier: viewControllerIdentifier) as? Self else {
+            preconditionFailure("Unable to instantiate view controller with identifier \(viewControllerIdentifier) as type \(String(describing: self))")
```
guykogus, line 37:
```suggestion
            preconditionFailure("Unable to instantiate view controller with identifier \(viewControllerIdentifier) as type \(type(of: self))")
```

Tests/UIKitTests/UIViewControllerExtensionsTests.swift
```diff
@@ -22,6 +22,10 @@ final class UIViewControllerExtensionsTests: XCTestCase {
         }
     }
 
+    class MyViewController: UIViewController {
+        var foo: String?
```
guykogus, line 26: Replace this property with an `IBOutlet` that you set in the storyboard, and then call `XCTAssertNotNil` in the test. I think that will more correctly ensure that the new function works correctly.

guykogus:
By the way, notice that the tests are failing...

VatoKo:
@guykogus I get "Could not find a storyboard named 'TestStoryboard' in bundle..." error on unit test. I checked already existing unit test where this storyboard is used and those tests are failing too. Any idea what could be causing this?

guykogus:
> @guykogus I get "Could not find a storyboard named 'TestStoryboard' in bundle..." error on unit test. I checked already existing unit test where this storyboard is used and those tests are failing too. Any idea what could be causing this?

I'm not really sure. It may be to do with catalyst. You may need to add `#if !targetEnvironment(macCatalyst)`. Give it a try.
Also, make sure that you're open the project in Xcode via the `SwifterSwift.xcworkspace` file in Finder. If you try to open the project directly from Xcode it will open it as a Swift package, which doesn't support resources (until Xcode 12).

VatoKo:
@guykogus I've always opened the project via `SwifterSwift.xcworkspace`, no problem with that. Tried `#if !targetEnvironment(macCatalyst)` check and it's still failing as you can see. 😕

Tests/UIKitTests/UIViewControllerExtensionsTests.swift
```diff
@@ -55,6 +59,15 @@ final class UIViewControllerExtensionsTests: XCTestCase {
         XCTAssertFalse(viewController.notificationFired)
     }
 
+    #if !targetEnvironment(macCatalyst)
+    
+    func testInstantiate() {
+        let myViewController = MyViewController.instantiate(from: "TestStoryboard")
```
guykogus, line 65: Found it! When running unit tests you need to define the `Bundle`.
```suggestion
        let myViewController = MyViewController.instantiate(from: "TestStoryboard", bundle: Bundle(for: UIViewControllerExtensionsTests.self))
```

Tests/UIKitTests/UIViewControllerExtensionsTests.swift
```diff
@@ -22,6 +22,10 @@ final class UIViewControllerExtensionsTests: XCTestCase {
         }
     }
 
+    class MyViewController: UIViewController {
```
guykogus, line 25: You need to move this class declaration outside of the test class otherwise the storyboard won't recognise it. In fact maybe just make a new file for it in the `ResourcesTests` folder.

Tests/UIKitTests/UIViewControllerExtensionsTests.swift
```diff
@@ -55,6 +59,15 @@ final class UIViewControllerExtensionsTests: XCTestCase {
         XCTAssertFalse(viewController.notificationFired)
     }
 
+    #if !targetEnvironment(macCatalyst)
+    
+    func testInstantiate() {
+        let myViewController = MyViewController.instantiate(from: "TestStoryboard")
+        XCTAssertNotNil(myViewController.testLabel)
```
guykogus, line 66: You will need to call `myViewController.loadViewIfNeeded()` before this or it will fail.

Tests/UIKitTests/UIViewControllerExtensionsTests.swift
```diff
@@ -55,6 +55,16 @@ final class UIViewControllerExtensionsTests: XCTestCase {
         XCTAssertFalse(viewController.notificationFired)
     }
 
+    #if !targetEnvironment(macCatalyst)
```
guykogus, line 58: Remove this now

Tests/ResourcesTests/MyViewController.swift
```diff
@@ -0,0 +1,13 @@
+//
+//  MyViewController.swift
+//  SwifterSwift
+//
+//  Created by Vato Kostava on 7/1/20.
+//  Copyright © 2020 SwifterSwift
+//
+
+import Foundation
```
guykogus, line 9: Fix this for Linux builds
```suggestion
#if canImport(UIKit) && !os(watchOS)
import UIKit
```

LucianoPAlmeida:
We have a problem with the build 
```
Tests/UIKitTests/UIViewControllerExtensionsTests.swift#L59: use of unresolved identifier ‘MyViewController’ 
let myViewController = MyViewController.instantiate(from: "TestStoryboard", bundle: Bundle(for: UIViewControllerExtensionsTests.self))
```

guykogus:
> We have a problem with the build
> 
> ```
> Tests/UIKitTests/UIViewControllerExtensionsTests.swift#L59: use of unresolved identifier ‘MyViewController’ 
> let myViewController = MyViewController.instantiate(from: "TestStoryboard", bundle: Bundle(for: UIViewControllerExtensionsTests.self))
> ```

Do you have any more context for this error?

LucianoPAlmeida:
Nope, I just copied what danger is saying :)

VatoKo:
I don't have such problem. The code compiles and even passes the test successfully.

LucianoPAlmeida:
Ok I'll take a look ASAP :)

guykogus:
@VatoKo I went ahead and fixed it for you. I was actually curious to see if I could push to your fork. I guess I can cos I have superpowers? 😅

The main issue was for the tvOS tests. We needed a separate storyboard for tvOS, since storyboards aren't cross-platform.

VatoKo:
@guykogus I guess I have to say thank you 😄

PR #863, opened 2020-06-08
`RandomAccessCollection`/`Collection` generalisations

guykogus:
🚀

Loosen some protocol restrictions to make functions more generic.

## Checklist


Sources/SwifterSwift/SwiftStdlib/CollectionExtensions.swift
```diff
@@ -89,26 +81,47 @@ public extension Collection where Index == Int {
     ///   - slice: size of array in each interation.
     ///   - body: a closure that takes an array of slice size as a parameter.
     func forEach(slice: Int, body: ([Element]) throws -> Void) rethrows {
```
guykogus, line 83: I'd really love to make this `SubSequence` instead of `[Element]`, but I don't want a breaking change.

LucianoPAlmeida: That was discussed back when implemented a long time ago ... I agree, but not sure it's worth a breaking change at this point.

Sources/SwifterSwift/SwiftStdlib/CollectionExtensions.swift
```diff
@@ -89,26 +81,47 @@ public extension Collection where Index == Int {
...
-        guard slice > 0, !isEmpty else { return }
+        guard slice > 0 else { return }
 
-        var value: Int = 0
-        while value < count {
-            try body(Array(self[Swift.max(value, startIndex)..<Swift.min(value + slice, endIndex)]))
-            value += slice
```
LucianoPAlmeida, line 97: What about 
```swift
var start = startIndex
repeat {
     let end = index(start, offsetBy: slice, limitedBy: endIndex) ?? endIndex
     try body(Array(self[start..<end]))
     start = end
} while (end != endIndex)
```
Suggestion, not sure this compiles(just sketch here)/ or handles all the cases, but I just searching for a way to avoid/simplify the last check and the duplicated index :)

Sources/SwifterSwift/SwiftStdlib/CollectionExtensions.swift
```diff
@@ -89,26 +81,47 @@ public extension Collection where Index == Int {
...
+        var start = startIndex
+        while let end = index(start, offsetBy: slice, limitedBy: endIndex) {
+            try body(Array(self[start..<end]))
+            start = index(start, offsetBy: slice)
```
LucianoPAlmeida, line 89:
```suggestion
start = end
```
Could it be that?

Sources/SwifterSwift/SwiftStdlib/CollectionExtensions.swift
LucianoPAlmeida: Maybe that would break on empty collection ...

guykogus: It doesn't compile because `end` is declared inside the loop so it isn't recognised outside the scope. I found a solution though :)

LucianoPAlmeida: Nice :)

LucianoPAlmeida:
We have to fix these timeout tests at some point :)

LucianoPAlmeida:
@guykogus Could this failing time tests fix enter in your clean up branch? It is often failing so I think we should give a little increase on the limit there.

guykogus:
> @guykogus Could this failing time tests fix enter in your clean up branch? It is often failing so I think we should give a little increase on the limit there.

Yep, fixed it: https://github.com/SwifterSwift/SwifterSwift/pull/864/commits/0c2d3dc139270da54ad9b3b00749e0222a3b4aee#diff-9029f943f4f9fe91d3038cdd34d323b8R63

LucianoPAlmeida:
Let's move on with this because it's really not ideal having to updating all the time since it's ready :)

PR #865, opened 2020-06-12
Update StringExtensions.swift

philCc:
-fixed: crashed when self.length=0


philCc:
> LGTM! Mind updates the Unit tests and add the "" case?

yes

PR #869, opened 2020-06-16
Reinstate debounce function

guykogus:
🚀
When moving to version 5.0.0 (commit 42353f5787fc946e0936aed014b4d88f3a4d327d) this function was removed. It was part of the deleted `struct SwifterSwift`. The function has been moved into `DispatchQueue`.

## Checklist


Sources/SwifterSwift/Dispatch/DispatchQueueExtensions.swift
```diff
@@ -56,6 +56,21 @@ public extension DispatchQueue {
         asyncAfter(deadline: .now() + delay, qos: qos, flags: flags, execute: work)
     }
 
+    func debounce(millisecondsDelay: Int, action: @escaping () -> Void) -> () -> Void {
```
LucianoPAlmeida, line 59: Why milliseconds, not a `DispatchTimeInterval`?

guykogus: It's how the original function was written. More than happy to make it a `TimeInterval`, as it's more standard.

LucianoPAlmeida: I think that it would be nicer :)

PR #870, opened 2020-06-26
add self use extensions

iMazy:
## Checklist

### DIFF
--- a/Sources/SwifterSwift/SwiftStdlib/Number+Extension.swift
+++ b/Sources/SwifterSwift/SwiftStdlib/Number+Extension.swift
@@ -1,0 +1,25 @@
+//
+//  Number+Extension.swift
+//  SwifterSwift
+//
+//  Created by Mazy on 2020/6/26.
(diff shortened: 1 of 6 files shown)


LucianoPAlmeida:
Hi @iMazy :)

Thanks for submitting this PR 
But unfortunately, I don't think those extensions fit here because of the following reasons:
1. Some of these are dependent on another framework(CryptoSwift) which we can't accept here because all the extensions here should be only dependent on the standard library or some native first-party Framework e.g. Foundation, CoreGraphics.
2. Some of them like `topmost` were previously discussed and not accepted.
3. Some of them like `group` we already have (or had, in this case, now we deprecated in favor the stdlib one)
4. The other ones related to `hairlineImageView`  look to be very specific. So I don't think they are a good fit for this library.

Those are my thoughts on this, but if you think that any of those extensions are still worth discussing, please open a PR with each specific extension(or group of related ones) + unit tests + changelog and we could continue the discussion there :)

PR #874, opened 2020-07-01
add RawRepresentableExtensions and Tests 

fanglinwei:
## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,20 +6,31 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 
 ### Added
 - **String**:
+  
   - Overloaded Swift's 'contains' operator (`~=`) for `String` to check regex matching. [#858](https://github.com/SwifterSwift/SwifterSwift/pull/858) by [VatoKo](https://github.com/VatoKo)
 - **DispatchQueue**:
+  
   - Added `asyncAfter(delay:qos:flags:execute:)` method to use it more conveniently without typing unnecessary `.now() + delay`. [#859](https://github.com/SwifterSwift/SwifterSwift/pull/859) by [VatoKo](https://github.com/VatoKo)
 - **RangeReplaceableCollection**:
+  
   - `subscript(offset:)` and `subscript(range:)` to access and replace elements by the index offsets. [#826](https://github.com/SwifterSwift/SwifterSwift/pull/826) by [guykogus](https://github.com/guykogus)
 - **Sequence**:
   - Added `first(where:equals:)` to find the first element of the sequence with having property by given key path equals to given value. [#836](https://github.com/SwifterSwift/SwifterSwift/pull/836) by [hamtiko](https://github.com/hamtiko)
   - Added `last(where:equals:)` to find the last element of the sequence with having property by given key path equals to given value. [#838](https://github.com/SwifterSwift/SwifterSwift/pull/838) by [hamtiko](https://github.com/hamtiko)
 - **SKNode**:
+  
   - `center`, `topLeft`, `topRight`, `bottomLeft`, `bottomRight` to get anchor position or set position using anchor. [#835](https://github.com/SwifterSwift/SwifterSwift/pull/835) by [rypyak](https://github.com/rypyak)
 - **CGSize**:
+  
   - Added new `+`, `+=`, `-` and `-=` operator extensions for tuple (width: CGFloat, height: CGFloat). [#841](https://github.com/SwifterSwift/SwifterSwift/pull/841) by [hamtiko](https://github.com/hamtiko)
 - **WKWebView**:
+  
   - Added `loadURL(_:)` and `loadURLString(_:)` extensions for `WkWebView`. [#851](https://github.com/SwifterSwift/SwifterSwift/pull/851) by [hamtiko](https://github.com/hamtiko)
+- **RawRepresentable**:
+  
+  - Added `init?(rawValue: RawValue?)` extensions for `RawRepresentable`. [#872](https://github.com/SwifterSwift/SwifterSwift/pull/872) by [fanglinwei](https://github.com/fanglinwei)
+  
+    
 
 ### Changed
 - **NSAttributedStringExtensions.swift**:
--- a/Sources/SwifterSwift/SwiftStdlib/RawRepresentableExtensions.swift
+++ b/Sources/SwifterSwift/SwiftStdlib/RawRepresentableExtensions.swift
@@ -1,0 +1,21 @@
+//
+//  RawRepresentableExtensions.swift
+//  SwifterSwift
+//
+//  Created by 方林威 on 2020/7/1.
+//  Copyright © 2020 SwifterSwift
+//
+
+// MARK: - Initializers
+public extension RawRepresentable {
+    
+    /// SwifterSwift: Initializes an  enumeration type. If `rawValue` was malformed, returns `nil`.
+    /// - Parameters:
+    ///   - rawValue: Initialize the value of the enum type..
+    init?(rawValue: RawValue?) {
+        guard let value = rawValue else {
+            return nil
+        }
+        self.init(rawValue: value)
+    }
+}
--- a/Tests/SwiftStdlibTests/RawRepresentableExtensionsTests.swift
+++ b/Tests/SwiftStdlibTests/RawRepresentableExtensionsTests.swift
@@ -1,0 +1,29 @@
+//
+//  RawRepresentableExtensionsTests.swift
+//  SwifterSwift
+//
+//  Created by 方林威 on 2020/7/1.
+//  Copyright © 2020 SwifterSwift
+//
+
+import XCTest
+@testable import SwifterSwift
+
+private enum Animal: String {
+    case cat
+    case dog
+}
+
+final class RawRepresentableExtensionsTests: XCTestCase {
+
+    func testOptionalRawValueInitializer() {
+        XCTAssertNil(Animal(rawValue: nil))
+
+        let optionalString1: String? = "cat"
+        let optionalString2: String? = "dog"
+        XCTAssertEqual(Animal(rawValue: optionalString1), Animal.cat)
+        XCTAssertEqual(Animal(rawValue: optionalString1)?.rawValue, Animal.cat.rawValue)
+        XCTAssertEqual(Animal(rawValue: optionalString2), Animal.dog)
+        XCTAssertEqual(Animal(rawValue: optionalString2)?.rawValue, Animal.dog.rawValue)
+    }
+}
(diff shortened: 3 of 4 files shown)


Sources/SwifterSwift/SwiftStdlib/RawRepresentableExtensions.swift
```diff
@@ -0,0 +1,21 @@
...
+//  Created by 方林威 on 2020/7/1.
+//  Copyright © 2020 SwifterSwift
+//
+
+// MARK: - Initializers
+public extension RawRepresentable {
+    
+    /// SwifterSwift: Initializes an  enumeration type. If `rawValue` was malformed, returns `nil`.
+    /// - Parameters:
+    ///   - rawValue: Initialize the value of the enum type..
```
LucianoPAlmeida, line 14:
```suggestion
    ///   - rawValue: Initialize the value of the enum type.
```

Sources/SwifterSwift/SwiftStdlib/RawRepresentableExtensions.swift
```diff
@@ -0,0 +1,21 @@
...
+//  SwifterSwift
+//
+//  Created by 方林威 on 2020/7/1.
+//  Copyright © 2020 SwifterSwift
+//
+
+// MARK: - Initializers
+public extension RawRepresentable {
+    
+    /// SwifterSwift: Initializes an  enumeration type. If `rawValue` was malformed, returns `nil`.
```
LucianoPAlmeida, line 12: I think its worth adding that the difference of this from the default native rawValue initializer is the argument is optional...

guykogus:
> @fanglinwei Thank you for taking the time to put this up! :)
> 
> TBH I don't see a real useful use case for it (since we have the native init) so I'm not sure it is worth adding here... so I'll defer to @omaralbeik @guykogus

I've never needed this, but it's potentially useful. Sometimes I wish that APIs were more flexible with optionals.

CHANGELOG.md
```diff
@@ -6,20 +6,31 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 
 ### Added
 - **String**:
+
```
guykogus, line 9: Please remove all these new empty lines you added.

CHANGELOG.md
```diff
@@ -6,20 +6,31 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
...
   - `center`, `topLeft`, `topRight`, `bottomLeft`, `bottomRight` to get anchor position or set position using anchor. [#835](https://github.com/SwifterSwift/SwifterSwift/pull/835) by [rypyak](https://github.com/rypyak)
 - **CGSize**:
+  
   - Added new `+`, `+=`, `-` and `-=` operator extensions for tuple (width: CGFloat, height: CGFloat). [#841](https://github.com/SwifterSwift/SwifterSwift/pull/841) by [hamtiko](https://github.com/hamtiko)
 - **WKWebView**:
+  
   - Added `loadURL(_:)` and `loadURLString(_:)` extensions for `WkWebView`. [#851](https://github.com/SwifterSwift/SwifterSwift/pull/851) by [hamtiko](https://github.com/hamtiko)
+- **RawRepresentable**:
+  
+  - Added `init?(rawValue: RawValue?)` extensions for `RawRepresentable`. [#872](https://github.com/SwifterSwift/SwifterSwift/pull/872) by [fanglinwei](https://github.com/fanglinwei)
```
guykogus, line 31:
```suggestion
  - Added `init?(rawValue:)` for creating a `RawRepresentable` with an optional raw value. [#872](https://github.com/SwifterSwift/SwifterSwift/pull/872) by [fanglinwei](https://github.com/fanglinwei)
```

Sources/SwifterSwift/SwiftStdlib/RawRepresentableExtensions.swift
guykogus, line 14:
```suggestion
    ///   - rawValue: The optional raw value to use for the new instance.
```

Sources/SwifterSwift/SwiftStdlib/RawRepresentableExtensions.swift
```diff
@@ -0,0 +1,21 @@
...
+//  RawRepresentableExtensions.swift
+//  SwifterSwift
+//
+//  Created by 方林威 on 2020/7/1.
+//  Copyright © 2020 SwifterSwift
+//
+
+// MARK: - Initializers
+public extension RawRepresentable {
+
```
guykogus, line 11: This is giving you a SwiftLint error.
```suggestion

```

Sources/SwifterSwift/SwiftStdlib/RawRepresentableExtensions.swift
```diff
@@ -0,0 +1,21 @@
...
+// MARK: - Initializers
+public extension RawRepresentable {
+    
+    /// SwifterSwift: Initializes an  enumeration type. If `rawValue` was malformed, returns `nil`.
+    /// - Parameters:
+    ///   - rawValue: Initialize the value of the enum type..
+    init?(rawValue: RawValue?) {
+        guard let value = rawValue else {
+            return nil
+        }
```
guykogus, line 18: Put these on 1 line
```suggestion
        guard let value = rawValue else { return nil }
```

Tests/SwiftStdlibTests/RawRepresentableExtensionsTests.swift
```diff
@@ -0,0 +1,29 @@
...
+        XCTAssertNil(Animal(rawValue: nil))
+
+        let optionalString1: String? = "cat"
+        let optionalString2: String? = "dog"
+        XCTAssertEqual(Animal(rawValue: optionalString1), Animal.cat)
+        XCTAssertEqual(Animal(rawValue: optionalString1)?.rawValue, Animal.cat.rawValue)
+        XCTAssertEqual(Animal(rawValue: optionalString2), Animal.dog)
+        XCTAssertEqual(Animal(rawValue: optionalString2)?.rawValue, Animal.dog.rawValue)
+    }
+}
```
guykogus, line 29:
```suggestion
    }

}
```

Tests/SwiftStdlibTests/RawRepresentableExtensionsTests.swift
```diff
@@ -0,0 +1,29 @@
...
+//  SwifterSwift
+//
+//  Created by 方林威 on 2020/7/1.
+//  Copyright © 2020 SwifterSwift
+//
+
+import XCTest
+@testable import SwifterSwift
+
+private enum Animal: String {
```
guykogus, line 12: I think you can move this inside the `testOptionalRawValueInitializer` function.

Sources/SwifterSwift/SwiftStdlib/RawRepresentableExtensions.swift
fanglinwei: Thank you for your suggestion, I will pay attention to the specification of SwiftLint next time

Sources/SwifterSwift/SwiftStdlib/RawRepresentableExtensions.swift
fanglinwei: yeah!

Sources/SwifterSwift/SwiftStdlib/RawRepresentableExtensions.swift
fanglinwei: Sorry, I was careless

fanglinwei: Yes, I think this description would be better

Sources/SwifterSwift/SwiftStdlib/RawRepresentableExtensions.swift
fanglinwei: Hahaha, I usually write it this way, because I see some code in swifter is line break.

fanglinwei: I agree with this way of writing

Tests/SwiftStdlibTests/RawRepresentableExtensionsTests.swift
fanglinwei: Yes, it can indeed be in the testOptionalRawValueInitializer function

omaralbeik:
Thanks for putting the time to work on this PR @fanglinwei !
While I see how this might be useful to map optionals. I'm afraid adding such extensions will open the door for more extensions that do the same of all native methods with slight changes, bloating the project more than it is necessary

fanglinwei:
> Thanks for putting the time to work on this PR @fanglinwei !
> While I see how this might be useful to map optionals. I'm afraid adding such extensions will open the door for more extensions that do the same of all native methods with slight changes, bloating the project more than it is necessary

@omaralbeik 
I think this initialization can make the code more concise, and the URL can also initialize an object from the string optional value

image

image

guykogus:
> > Thanks for putting the time to work on this PR @fanglinwei !
> > While I see how this might be useful to map optionals. I'm afraid adding such extensions will open the door for more extensions that do the same of all native methods with slight changes, bloating the project more than it is necessary
> 
> @omaralbeik
> I think this initialization can make the code more concise, and the URL can also initialize an object from the string optional value

I'm guessing that the problem that you're trying to solve is receiving `nil` for the `animal` property of your model. Why don't you use this?

```Swift
enum Animal: String, Codable {
    case dog
    case cat
}

struct Model: Codable {
    let animal: Animal?
}
```

fanglinwei:
> > > Thanks for putting the time to work on this PR @fanglinwei !
> > > While I see how this might be useful to map optionals. I'm afraid adding such extensions will open the door for more extensions that do the same of all native methods with slight changes, bloating the project more than it is necessary
> > 
> > 
> > @omaralbeik
> > I think this initialization can make the code more concise, and the URL can also initialize an object from the string optional value
> 
> I'm guessing that the problem that you're trying to solve is receiving `nil` for the `animal` property of your model. Why don't you use this?
> 
> ```swift
> enum Animal: String, Codable {
>     case dog
>     case cat
> }
> 
> struct Model: Codable {
>     let animal: Animal?
> }
> ```

Why do URLs have similar extensions, I think their functions are similar

guykogus:
> > > > Thanks for putting the time to work on this PR @fanglinwei !
> > > > While I see how this might be useful to map optionals. I'm afraid adding such extensions will open the door for more extensions that do the same of all native methods with slight changes, bloating the project more than it is necessary
> > > 
> > > 
> > > @omaralbeik
> > > I think this initialization can make the code more concise, and the URL can also initialize an object from the string optional value
> > 
> > 
> > I'm guessing that the problem that you're trying to solve is receiving `nil` for the `animal` property of your model. Why don't you use this?
> > ```swift
> > enum Animal: String, Codable {
> >     case dog
> >     case cat
> > }
> > 
> > struct Model: Codable {
> >     let animal: Animal?
> > }
> > ```
> 
> Why do URLs have similar extensions, I think their functions are similar

I think that was @omaralbeik's point, that we'll bloat the project if we keep adding these types of functions.

LucianoPAlmeida:
Have we come up with a final decision about this?

fanglinwei:
> Have we come up with a final decision about this?

Optional type initialization should be customized for business scenarios. It is not suitable as a public extension library

PR #880, opened 2020-07-21
Make √ generic for FloatingPoint values

guykogus:
🚀
The underlying function is generic, so might as well do the same.

## Checklist

### DIFF
--- a/Sources/SwifterSwift/SwiftStdlib/FloatingPointExtensions.swift
+++ b/Sources/SwifterSwift/SwiftStdlib/FloatingPointExtensions.swift
@@ -81,3 +81,15 @@ public prefix func ± <T: FloatingPoint> (number: T) -> (T, T) {
     return 0 ± number
 }
 // swiftlint:enable identifier_name
+
+// swiftlint:disable identifier_name
+prefix operator √
+/// SwifterSwift: Square root of float.
+///
+/// - Parameter float: float value to find square root for
+/// - Returns: square root of given float.
+public prefix func √<T> (float: T) -> T where T: FloatingPoint {
+    // http://nshipster.com/swift-operators/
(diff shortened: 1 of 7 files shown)


Sources/SwifterSwift/SwiftStdlib/FloatingPointExtensions.swift
```diff
@@ -81,3 +81,15 @@ public prefix func ± <T: FloatingPoint> (number: T) -> (T, T) {
     return 0 ± number
 }
 // swiftlint:enable identifier_name
+
+// swiftlint:disable identifier_name
```
LucianoPAlmeida, line 85: I think that would be a good thing to add those exceptions on the swiftlint.yaml file ... common cases like those to avoid this bunch of enable disable ... no in this PR though, might be good for an issue with starter task :)

Sources/SwifterSwift/SwiftStdlib/FloatingPointExtensions.swift
```diff
@@ -81,3 +81,15 @@ public prefix func ± <T: FloatingPoint> (number: T) -> (T, T) {
...
+prefix operator √
+/// SwifterSwift: Square root of float.
+///
+/// - Parameter float: float value to find square root for
+/// - Returns: square root of given float.
+public prefix func √<T> (float: T) -> T where T: FloatingPoint {
```
LucianoPAlmeida, line 91: I wonder if its time to think about inline such trivial numeric operators like this, the implementation is trivial so maybe we should throw an `@inlinable` here (and possibly audit other places where we could do it as well). This is good especially for generic functions like this one because it would allow the compiler to specialize such functions on clients that install this library as a module.

Issue #879, opened 2020-07-21
want more helpers?


gurgeous:
Thanks for creating such a great library! SwifterSwift is amazing and helped me get rid of many of our custom extensions. Want a few more? Here are additional helpers that I can add if you want:

- CGFloat.random
- CGRect x, y, width, height, center, centerX, centerY, left, top, right, bottom
- CGSize init(square: CGFloat)
- String escapeRegex
- UIColor image(ofSize: CGSize)
- UIImage averageColor

### UIView
- loadFromNib(withClass name: T.Type)
- frame - origin, width, height, size, top, left, bottom, right
- bounds - boundsWidth, boundsHeight, boundsSize
- constraints - allConstraints, findConstraint(attribute: NSLayoutConstraint.Attribute), widthConstraint, heightConstraint, leadingConstraint, trailingConstraint, topConstraint, bottomConstraint
^ these are for finding existing constraints and are typically used to modify a constant

Let me know if any of these are of interest and I'll put together a PR.

(I also have a fair number of NSRegularExpression, NSString and NSMutableAttributedString extensions but I think those are of less interest. Let me know if you'd like to see those too.)

guykogus:
> * CGFloat.random **- Part of the standard library**
> * CGRect x, y, width, height, center, centerX, centerY, left, top, right, bottom **- already have min/mid/max(X/Y) etc**
> * CGSize init(square: CGFloat) **- Not a huge saving, IMO**
> * String escapeRegex **- 👍**
> * UIColor image(ofSize: CGSize) **- what's the benefit of this?**
> * UIImage averageColor **- could be interesting**
> 
> ### UIView
> * loadFromNib(withClass name: T.Type) **- Good idea, for now we just have it returning `UIView`**
> * frame - origin, width, height, size, top, left, bottom, right **- Same as with `CGRect`, no need**
> * bounds - boundsWidth, boundsHeight, boundsSize **- not much of a benefit here calling `boundsWidth` instead of `bounds.width`, for example**
> * constraints - allConstraints, findConstraint(attribute: NSLayoutConstraint.Attribute), widthConstraint, heightConstraint, leadingConstraint, trailingConstraint, topConstraint, bottomConstraint **- 👍**
>   ^ these are for finding existing constraints and are typically used to modify a constant
> 
[...]

PR #884, opened 2020-07-21
UIImage.averageColor

gurgeous:
Calculates the average color for an entire UIImage. This is useful for determining whether an image is light or dark, among other things. See #879.

## Checklist

### DIFF
--- a/Sources/SwifterSwift/UIKit/UIImageExtensions.swift
+++ b/Sources/SwifterSwift/UIKit/UIImageExtensions.swift
@@ -36,7 +36,6 @@ public extension UIImage {
     /// SwifterSwift: Average color for this image
     func averageColor() -> UIColor? {
         // https://stackoverflow.com/questions/26330924
-        // note: self.ciImage appears to always be nil
         guard let ciImage = ciImage ?? CIImage(image: self) else { return nil }
 
         // CIAreaAverage returns a single-pixel image that contains the average color for a given region of an image.
--- a/Sources/SwifterSwift/UIKit/UIImageExtensions.swift
+++ b/Sources/SwifterSwift/UIKit/UIImageExtensions.swift
@@ -32,8 +32,8 @@ public extension UIImage {
         return withRenderingMode(.alwaysTemplate)
     }
 
-    /// SwifterSwift: Average color for this image
     #if canImport(CoreImage)
+    /// SwifterSwift: Average color for this image
     func averageColor() -> UIColor? {
         // https://stackoverflow.com/questions/26330924
         guard let ciImage = CIImage(image: self) else {
--- a/Sources/SwifterSwift/UIKit/UIImageExtensions.swift
+++ b/Sources/SwifterSwift/UIKit/UIImageExtensions.swift
@@ -39,7 +39,7 @@ public extension UIImage {
             return nil
         }
 
-        // run CIAreaAverage filter to get average
+        // CIAreaAverage returns a single-pixel image that contains the average color for a given region of an image.
         let parameters = [kCIInputImageKey: ciImage, kCIInputExtentKey: CIVector(cgRect: ciImage.extent)]
         guard let outputImage = CIFilter(name: "CIAreaAverage", parameters: parameters)?.outputImage else {
             return nil
--- a/Sources/SwifterSwift/UIKit/UIImageExtensions.swift
+++ b/Sources/SwifterSwift/UIKit/UIImageExtensions.swift
@@ -34,7 +34,7 @@ public extension UIImage {
 
     /// SwifterSwift: Average color for this image
     #if canImport(CoreImage)
-    var averageColor: UIColor? {
+    func averageColor() -> UIColor? {
         // https://stackoverflow.com/questions/26330924
         guard let ciImage = CIImage(image: self) else {
             return nil
--- a/Sources/SwifterSwift/UIKit/UIImageExtensions.swift
+++ b/Sources/SwifterSwift/UIKit/UIImageExtensions.swift
@@ -37,9 +37,7 @@ public extension UIImage {
     func averageColor() -> UIColor? {
         // https://stackoverflow.com/questions/26330924
         // note: self.ciImage appears to always be nil
-        guard let ciImage = CIImage(image: self) else {
-            return nil
-        }
+        guard let ciImage = ciImage ?? CIImage(image: self) else { return nil }
 
         // CIAreaAverage returns a single-pixel image that contains the average color for a given region of an image.
         let parameters = [kCIInputImageKey: ciImage, kCIInputExtentKey: CIVector(cgRect: ciImage.extent)]
--- a/Sources/SwifterSwift/UIKit/UIImageExtensions.swift
+++ b/Sources/SwifterSwift/UIKit/UIImageExtensions.swift
@@ -36,6 +36,7 @@ public extension UIImage {
     /// SwifterSwift: Average color for this image
     func averageColor() -> UIColor? {
         // https://stackoverflow.com/questions/26330924
+        // note: self.ciImage appears to always be nil
         guard let ciImage = CIImage(image: self) else {
             return nil
         }
@@ -48,7 +49,8 @@ public extension UIImage {
 
         // After getting the single-pixel image from the filter extract pixel's RGBA8 data
         var bitmap = [UInt8](repeating: 0, count: 4)
-        let context = CIContext(options: [.workingColorSpace: NSNull()])
+        let workingColorSpace: Any = cgImage?.colorSpace ?? NSNull()
+        let context = CIContext(options: [.workingColorSpace: workingColorSpace])
         context.render(outputImage,
                        toBitmap: &bitmap,
                        rowBytes: 4,
--- a/Sources/SwifterSwift/UIKit/UIImageExtensions.swift
+++ b/Sources/SwifterSwift/UIKit/UIImageExtensions.swift
@@ -32,11 +32,39 @@ public extension UIImage {
         return withRenderingMode(.alwaysTemplate)
     }
 
+    /// SwifterSwift: Average color for this image
+    var averageColor: UIColor? {
+        // https://stackoverflow.com/questions/26330924
+        guard let ciImage = CIImage(image: self) else {
+            return nil
+        }
+
+        // run CIAreaAverage filter to get average
+        let parameters = [kCIInputImageKey: ciImage, kCIInputExtentKey: CIVector(cgRect: ciImage.extent)]
+        guard let outputImage = CIFilter(name: "CIAreaAverage", parameters: parameters)?.outputImage else {
+            return nil
+        }
+
+        // extract single pixel of RGBA8 data
+        var bitmap = [UInt8](repeating: 0, count: 4)
+        let context = CIContext(options: [.workingColorSpace: NSNull()])
+        context.render(outputImage,
+                       toBitmap: &bitmap,
+                       rowBytes: 4,
+                       bounds: CGRect(x: 0, y: 0, width: 1, height: 1),
+                       format: .RGBA8,
+                       colorSpace: nil)
+
+        // convert to UIColor
+        return UIColor(red: CGFloat(bitmap[0]) / 255.0,
+                       green: CGFloat(bitmap[1]) / 255.0,
+                       blue: CGFloat(bitmap[2]) / 255.0,
+                       alpha: CGFloat(bitmap[3]) / 255.0)
+    }
 }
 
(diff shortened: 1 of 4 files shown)


Sources/SwifterSwift/UIKit/UIImageExtensions.swift
```diff
@@ -32,6 +32,35 @@ public extension UIImage {
...
     }
 
+    /// SwifterSwift: Average color for this image
+    var averageColor: UIColor? {
+        // https://stackoverflow.com/questions/26330924
+        guard let ciImage = CIImage(image: self) else {
+            return nil
+        }
+
+        // run CIAreaAverage filter to get average
```
LucianoPAlmeida, line 42:
```suggestion
        // CIAreaAverage returns a single-pixel image that contains the average color for a given region of an image.
```

Sources/SwifterSwift/UIKit/UIImageExtensions.swift
```diff
@@ -32,6 +32,35 @@ public extension UIImage {
...
+        let parameters = [kCIInputImageKey: ciImage, kCIInputExtentKey: CIVector(cgRect: ciImage.extent)]
+        guard let outputImage = CIFilter(name: "CIAreaAverage", parameters: parameters)?.outputImage else {
+            return nil
+        }
+
+        // extract single pixel of RGBA8 data
```
LucianoPAlmeida, line 48:
```suggestion
        // After getting the single-pixel image from the filter extract pixel's RGBA8 data
```

Sources/SwifterSwift/UIKit/UIImageExtensions.swift
```diff
@@ -32,6 +32,35 @@ public extension UIImage {
...
+        var bitmap = [UInt8](repeating: 0, count: 4)
+        let context = CIContext(options: [.workingColorSpace: NSNull()])
+        context.render(outputImage,
+                       toBitmap: &bitmap,
+                       rowBytes: 4,
+                       bounds: CGRect(x: 0, y: 0, width: 1, height: 1),
+                       format: .RGBA8,
+                       colorSpace: nil)
+
+        // convert to UIColor
```
LucianoPAlmeida, line 58:
```suggestion
        // Convert pixel data to UIColor
```

Sources/SwifterSwift/UIKit/UIImageExtensions.swift
```diff
@@ -32,6 +32,35 @@ public extension UIImage {
         return withRenderingMode(.alwaysTemplate)
     }
 
+    /// SwifterSwift: Average color for this image
+    var averageColor: UIColor? {
```
LucianoPAlmeida, line 36: I think we are missing a can import here? CoreImage?

gurgeous: Does this look correct? I am unfamiliar with some of these issues, unfortunately

gurgeous:
Sure, not a problem

gurgeous:
Let me know if you'd like me to take further action here. Not sure who's in charge, or maybe if that's me now :)

Sources/SwifterSwift/UIKit/UIImageExtensions.swift
```diff
@@ -32,6 +32,37 @@ public extension UIImage {
         return withRenderingMode(.alwaysTemplate)
     }
 
+    /// SwifterSwift: Average color for this image
+    #if canImport(CoreImage)
```
guykogus, line 36:
```suggestion
    #if canImport(CoreImage)
    /// SwifterSwift: Average color for this image
```

Sources/SwifterSwift/UIKit/UIImageExtensions.swift
```diff
@@ -32,6 +32,37 @@ public extension UIImage {
...
+
+        // CIAreaAverage returns a single-pixel image that contains the average color for a given region of an image.
+        let parameters = [kCIInputImageKey: ciImage, kCIInputExtentKey: CIVector(cgRect: ciImage.extent)]
+        guard let outputImage = CIFilter(name: "CIAreaAverage", parameters: parameters)?.outputImage else {
+            return nil
+        }
+
+        // After getting the single-pixel image from the filter extract pixel's RGBA8 data
+        var bitmap = [UInt8](repeating: 0, count: 4)
+        let context = CIContext(options: [.workingColorSpace: NSNull()])
```
guykogus, line 51: Why do you set this color space? The documentation says
> To request that Core Image perform no color management, specify the NSNull object as the value for this key. Use this option for images that don’t contain color data (such as elevation maps, normal vector maps, and sampled function tables).

but this image may have a color space, no? I.e. `cgImage.colorSpace`. So maybe here you could use
`[.workingColorSpace: cgImage.colorSpace ?? NSNull()]`

Sources/SwifterSwift/UIKit/UIImageExtensions.swift
```diff
@@ -32,6 +32,37 @@ public extension UIImage {
         return withRenderingMode(.alwaysTemplate)
     }
 
+    /// SwifterSwift: Average color for this image
+    #if canImport(CoreImage)
+    func averageColor() -> UIColor? {
+        // https://stackoverflow.com/questions/26330924
+        guard let ciImage = CIImage(image: self) else {
+            return nil
+        }
```
guykogus, line 41: `UIImage` already has this value as a property. You can also one-line this.
```suggestion
        guard let ciImage = ciImage else { return nil }
```

Tests/UIKitTests/UIImageExtensionsTests.swift
```diff
@@ -13,6 +13,13 @@ import XCTest
 import UIKit
 
 final class UIImageExtensionsTests: XCTestCase {
+    func testAverageColor() {
+        // note that not all colors precisely survive the roundtrip due to CI colorspace
+        // management. Blue & brown seem safe, though.
```
guykogus, line 18: If you don't get exact values you can still use `XCTAssertEqualWithAccuracy` to check the components of the colours.
And how about adding, for example, an image with size (0, 0)? Or a combination of colours?

Sources/SwifterSwift/UIKit/UIImageExtensions.swift
gurgeous: UIImage.ciImage is often nil according to my testing (and googling). Gotta create a new CIImage, unfortunately.

Sources/SwifterSwift/UIKit/UIImageExtensions.swift
gurgeous: good idea

Sources/SwifterSwift/UIKit/UIImageExtensions.swift
guykogus: Actually yeah, the documentation even says:
> If the UIImage object was initialized using a CGImage, the value of the property is nil.

How about this then?
```suggestion
        guard let ciImage = ciImage ?? CIImage(image: self) else { return nil }
```

Tests/UIKitTests/UIImageExtensionsTests.swift
gurgeous: ok, how about this? (see latest)

Tests/UIKitTests/UIImageExtensionsTests.swift
```diff
@@ -13,6 +13,40 @@ import XCTest
 import UIKit
 
 final class UIImageExtensionsTests: XCTestCase {
+    func testAverageColor() {
+        // note that not all colors precisely survive the roundtrip due to colorspace
+        // management. Use XCTAssertEqual w/ accuracy.
+        func assertColorsEqual(_ color1: UIColor, _ color2: UIColor, accuracy: CGFloat = 0.01) {
```
guykogus, line 19: This looks to me like a perfect extension to `XCTestCase` 😏

Sources/SwifterSwift/UIKit/UIImageExtensions.swift
guykogus: I think it'll be good to use my suggestion as it can avoid recreating `CIImage` objects if one already exists.

Tests/UIKitTests/UIImageExtensionsTests.swift
```diff
@@ -13,6 +13,40 @@ import XCTest
...
+            var r1: CGFloat = 0, g1: CGFloat = 0, b1: CGFloat = 0, a1: CGFloat = 0
+            var r2: CGFloat = 0, g2: CGFloat = 0, b2: CGFloat = 0, a2: CGFloat = 0
+            color1.getRed(&r1, green: &g1, blue: &b1, alpha: &a1)
+            color2.getRed(&r2, green: &g2, blue: &b2, alpha: &a2)
```
guykogus, line 23: In tests we don't have the same limitation for not being able to use other extensions. In `UIColor` we have `rgbComponents` and `alpha`, so that can simplify things for you. E.g.
```Swift
let (r1, g1, b1) = color1.rgbComponents
let (r2, g2, b2) = color2.rgbComponents
```

Sources/SwifterSwift/UIKit/UIImageExtensions.swift
gurgeous: 👍

Tests/UIKitTests/UIImageExtensionsTests.swift
gurgeous: I experimented a bit. I think `rgbComponents` doesn't work quite as well because it returns Ints, which makes it more difficult to use XCTAssertEqual with `accuracy`. I'd have to cast to floats every step of the way. Good idea, though.

Tests/UIKitTests/UIImageExtensionsTests.swift
gurgeous: You are 100% correct. 😃 Hesitant to jam that in this PR without a bit more direction, though. Where would it live? Is this for end users, or just our own tests? I don't want to slow down the PR further with build decisions...

guykogus: I ended spending a long time looking into this 😬 See #889

gurgeous: Excellent, thank you! I'm glad I didn't try to attack this one because I'm sure I would've gotten stuck in build issues. Looks perfect, though. It's a bit hard to tell from the project diff - is this for internal or external use?

Let's get #889 merged and then I can revisit `testAverageColor`

LucianoPAlmeida:
We now have some conflicts, can you please resolve then? :)

gurgeous:
Certainly, done.

LucianoPAlmeida:
@gurgeous @guykogus what is missing on this one?

guykogus:
@gurgeous you can now use the new colour tests 😃

gurgeous:
Thanks for adding the color test - this is done.

Looks like the build failed installing `danger-swiftlint`. I'm not familiar with this stuff, but it looks like that gem was deprecated a while back. https://github.com/ashfurrow/danger-swiftlint. Maybe it finally rotted?

Sources/SwifterSwift/UIKit/UIImageExtensions.swift
guykogus: @gurgeous you marked this as resolved but didn't make the change?

Tests/UIKitTests/UIImageExtensionsTests.swift
```diff
@@ -13,6 +13,27 @@ import XCTest
 import UIKit
 
 final class UIImageExtensionsTests: XCTestCase {
+    func testAverageColor() {
```
guykogus, line 16: Minor nitpick, but can you please add an empty space?
```suggestion

    func testAverageColor() {
```

Sources/SwifterSwift/UIKit/UIImageExtensions.swift
```diff
@@ -32,6 +32,37 @@ public extension UIImage {
         return withRenderingMode(.alwaysTemplate)
     }
 
+    #if canImport(CoreImage)
+    /// SwifterSwift: Average color for this image
+    func averageColor() -> UIColor? {
+        // https://stackoverflow.com/questions/26330924
+        // note: self.ciImage appears to always be nil
```
guykogus, line 39: Do we need this comment? I don't think it helps.

guykogus (changes requested):

guykogus (approved):

gurgeous:
🎂

PR #885, opened 2020-07-21
loadFromNib(withClass)

gurgeous:
Loads a UIView of a particular type from a nib file. See #879.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,8 +33,8 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 - **UIView**:
   - Added `masksToBounds` (IBInspectable) extension. [#877](https://github.com/SwifterSwift/SwifterSwift/pull/877) by [hamtiko](https://github.com/hamtiko)
   - Added `loadFromNib(withClass:)`, which loads a UIView of a particular type from a nib file. [#885](https://github.com/SwifterSwift/SwifterSwift/pull/885) by [gurgeous](https://github.com/gurgeous)
-  - Added `findConstraint` for finding an existing constraint
-  - Added `widthConstraint`, `heightConstraint`, `leadingConstraint`, `trailingConstraint`, `topConstraint`, and `bottomConstraint` for finding specific constraints
+  - Added `findConstraint` for finding an existing constraint. [#886](https://github.com/SwifterSwift/SwifterSwift/pull/886) by [gurgeous]
+  - Added `widthConstraint`, `heightConstraint`, `leadingConstraint`, `trailingConstraint`, `topConstraint`, and `bottomConstraint` for finding specific constraints. [#886](https://github.com/SwifterSwift/SwifterSwift/pull/886) by [gurgeous]
 
 ### Changed
 - **NSAttributedString**:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
   - Added `isStandGoalMet`, `isExerciseTimeGoalMet`, and `isEnergyBurnedGoalMet`. [#875](https://github.com/SwifterSwift/SwifterSwift/pull/875) by [lhygilbert](https://github.com/lhygilbert)
 - **UIView**:
   - Added `masksToBounds` (IBInspectable) extension. [#877](https://github.com/SwifterSwift/SwifterSwift/pull/877) by [hamtiko](https://github.com/hamtiko)
+  - Added `loadFromNib(withClass)`, which loads a UIView of a particular type from a nib file.
 
 ### Changed
 - **NSAttributedString**:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,7 +32,7 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
   - Added `isStandGoalMet`, `isExerciseTimeGoalMet`, and `isEnergyBurnedGoalMet`. [#875](https://github.com/SwifterSwift/SwifterSwift/pull/875) by [lhygilbert](https://github.com/lhygilbert)
 - **UIView**:
   - Added `masksToBounds` (IBInspectable) extension. [#877](https://github.com/SwifterSwift/SwifterSwift/pull/877) by [hamtiko](https://github.com/hamtiko)
-  - Added `loadFromNib(withClass:)`, which loads a UIView of a particular type from a nib file.
+  - Added `loadFromNib(withClass:)`, which loads a UIView of a particular type from a nib file. [#885](https://github.com/SwifterSwift/SwifterSwift/pull/885) by [gurgeous](https://github.com/gurgeous)
   - Added `findConstraint` for finding an existing constraint
   - Added `widthConstraint`, `heightConstraint`, `leadingConstraint`, `trailingConstraint`, `topConstraint`, and `bottomConstraint` for finding specific constraints
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,7 +31,7 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
   - Added `isStandGoalMet`, `isExerciseTimeGoalMet`, and `isEnergyBurnedGoalMet`. [#875](https://github.com/SwifterSwift/SwifterSwift/pull/875) by [lhygilbert](https://github.com/lhygilbert)
 - **UIView**:
   - Added `masksToBounds` (IBInspectable) extension. [#877](https://github.com/SwifterSwift/SwifterSwift/pull/877) by [hamtiko](https://github.com/hamtiko)
-  - Added `loadFromNib(withClass)`, which loads a UIView of a particular type from a nib file.
+  - Added `loadFromNib(withClass:)`, which loads a UIView of a particular type from a nib file.
 
 ### Changed
 - **NSAttributedString**:
--- a/Sources/SwifterSwift/UIKit/UIViewExtensions.swift
+++ b/Sources/SwifterSwift/UIKit/UIViewExtensions.swift
@@ -337,7 +337,7 @@ public extension UIView {
     /// - Returns: UIView
     class func loadFromNib<T: UIView>(withClass name: T.Type, bundle: Bundle? = nil) -> T {
         let named = String(describing: name)
-        guard let view = UINib(nibName: name, bundle: bundle).instantiate(withOwner: nil, options: nil)[0] as? T else {
+        guard let view = UINib(nibName: named, bundle: bundle).instantiate(withOwner: nil, options: nil)[0] as? T else {
             fatalError("First element in xib file \(named) is not of type \(named)")
         }
         return view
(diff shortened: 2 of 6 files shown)


LucianoPAlmeida:
CI is failing on TVOS Build, can you take a look @gurgeous? :)

gurgeous:
I believe this is fixed. The loadFromNib test uses UITableViewCell.xib, so that will only work on iOS.

Sources/SwifterSwift/UIKit/UIViewExtensions.swift
```diff
@@ -329,6 +329,20 @@ public extension UIView {
...
 
+    /// SwifterSwift: Load view of a certain type from nib
+    ///
+    /// - Parameters:
+    ///   - withClass: UIView type.
+    ///   - bundle: bundle of nib (default is nil).
+    /// - Returns: UIView
+    class func loadFromNib<T: UIView>(withClass name: T.Type, bundle: Bundle? = nil) -> T {
+        let named = String(describing: name)
+        guard let view = loadFromNib(named: named, bundle: bundle) as? T else {
```
LucianoPAlmeida, line 340:
```suggestion
        guard let view = UINib(nibName: name, bundle: bundle).instantiate(withOwner: nil, options: nil)[0] as? T else {
```

Sources/SwifterSwift/UIKit/UIViewExtensions.swift
```diff
@@ -329,6 +329,20 @@ public extension UIView {
...
+    /// SwifterSwift: Load view of a certain type from nib
+    ///
+    /// - Parameters:
+    ///   - withClass: UIView type.
+    ///   - bundle: bundle of nib (default is nil).
+    /// - Returns: UIView
+    class func loadFromNib<T: UIView>(withClass name: T.Type, bundle: Bundle? = nil) -> T {
+        let named = String(describing: name)
+        guard let view = UINib(nibName: named, bundle: bundle).instantiate(withOwner: nil, options: nil)[0] as? T else {
+            fatalError("First element in xib file \(named) is not of type \(named)")
```
omaralbeik, line 341: `fatalError` is so harsh and untestable here, what do you think of making the function return an optional view instead?

gurgeous: There's a reason for this - it's intended to match `dequeueReusableCell`, which is also quite harsh. I almost always use optionals in my own code as well, but I think there are a few spots (like nib loading) where it's simply "developer error" if something goes wrong. UIKit will also throw an exception if a nib file is not found, for example. Reasonable behavior in my opinion. Not sure a user could reasonably handle the error and continue here.

Having said that, happy to change it if you feel otherwise. Users can always throw their own fatalErrors when calling this function.

LucianoPAlmeida: I used to do that in the past as well but the thing is discussing this one time, I've learned and it's been working for me since, that for loading static resources which should always be there unless someone deletes it does make sense to fail fatal or at least assert it, so we verify in development that it is something wrong with our asset, xib ... or any other static resource. To me, the ideal is always returning a non-optional(because the resource should be there) and having a `assert` that only triggers in debug builds to break when something wrong with it. With that, we achieve both safety when loading resources and avoid a harsh crash in a production application.
Something like this 
```swift
guard let view = UINib(nibName: named, bundle: bundle).instantiate(withOwner: nil, options: nil)[0] as? T else {
  assertionFailure("First element in xib file \(named) is not of type \(named)")// Assert so we are warned on development
  return T()  // Always return something because the static resource it is expected to be there always, but don't crash in production and avoid the nil checks around calls.
}
```
Those are my 2 cents :)

gurgeous: Happy to make that change. @omaralbeik, what do you think?

omaralbeik: I see your point, let's go with the `fatalError` approach then 👍

gurgeous:
Sounds like we're good to go, then. Should I resolve and merge? LMK

CHANGELOG.md
```diff
@@ -31,6 +31,7 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
   - Added `isStandGoalMet`, `isExerciseTimeGoalMet`, and `isEnergyBurnedGoalMet`. [#875](https://github.com/SwifterSwift/SwifterSwift/pull/875) by [lhygilbert](https://github.com/lhygilbert)
 - **UIView**:
   - Added `masksToBounds` (IBInspectable) extension. [#877](https://github.com/SwifterSwift/SwifterSwift/pull/877) by [hamtiko](https://github.com/hamtiko)
+  - Added `loadFromNib(withClass)`, which loads a UIView of a particular type from a nib file.
```
guykogus, line 34: Sorry to be so picky 😬
```suggestion
  - Added `loadFromNib(withClass:)`, which loads a UIView of a particular type from a nib file.
```

Tests/UIKitTests/UIViewExtensionsTests.swift
```diff
@@ -322,6 +322,14 @@ final class UIViewExtensionsTests: XCTestCase {
         XCTAssertEqual(view1.transform, view3.transform)
     }
 
+    #if os(iOS) // UITableViewCell nib only works on iOS
```
guykogus, line 325: You can create a separate XIB file for tvOS. You may need to create a separate folder for tvOS-specific resources so that you can use the same file name.

gurgeous: This is a good idea. I'm slightly inclined to save it for the next PR since it's a minor testing improvement, though. What do you think?

guykogus: That sounds like one of those `TODO`s that will exist forevermore in your code 😅

gurgeous: You can read me like an open book :)

Happy to do it if you want, though. Give me a nudge either way!

guykogus: Yes, please do it. It should take you 2 minutes

gurgeous: I will give it a whirl. I'm not familiar with the project and I've never used tvOS, so I doubt this will take me two minutes.

I think I can create a `UIImageView.xib` for each test target since that view is supported on both platforms. Are you envisioning a `tvOsResourcesTests` folder inside `Tests`?

guykogus: Indeed. In `Test/ResourcesTests` create a `tvOS` folder and create the XIB in there. Remember to make sure that you select the `tvOS` section when creating the XIB file so that it will load correctly.

Screen Shot 2020-07-28 at 17 59 07

gurgeous: That was helpful, thank you. How about this?

Sources/SwifterSwift/UIKit/UIViewExtensions.swift
```diff
@@ -329,6 +329,20 @@ public extension UIView {
         return UINib(nibName: name, bundle: bundle).instantiate(withOwner: nil, options: nil)[0] as? UIView
     }
 
+    /// SwifterSwift: Load view of a certain type from nib
+    ///
+    /// - Parameters:
+    ///   - withClass: UIView type.
+    ///   - bundle: bundle of nib (default is nil).
+    /// - Returns: UIView
+    class func loadFromNib<T: UIView>(withClass name: T.Type, bundle: Bundle? = nil) -> T {
```
LucianoPAlmeida, line 338: Since we are still here, one question what do you think about the naming `instanceFromNib`? It is just curiosity, I found the name `load` a bit misleading since the action is really to instantiate ... but is more a question than an actual comment :)

gurgeous: The name was vaguely chosen to match `loadNibNamed`. Don't have a strong opinion, though.

Tests/UIKitTests/UIViewExtensionsTests.swift
```diff
@@ -322,6 +322,13 @@ final class UIViewExtensionsTests: XCTestCase {
         XCTAssertEqual(view1.transform, view3.transform)
     }
 
+    func testLoadFromNib() {
+        // versions of this nib are present for both iOS and tvOS tests
```
guykogus, line 326: No need for this comment

Sources/SwifterSwift/UIKit/UIViewExtensions.swift
guykogus: I like the current name as it basically replaces the one above. It's also what I call it in my project except that I have something a little sexier:
```Swift
public extension UIView {
    class func loadFromNib(in bundle: Bundle = .main) -> Self? {
        return bundle.loadFirstObject(inNibNamed: String(describing: self))
    }
}
```
The only reason I can do it like that is because I have another extension:
```Swift
public extension Bundle {
    func loadFirstObject<T>(inNibNamed name: String, owner: Any? = nil, options: [UINib.OptionsKey: Any]? = nil) -> T? {
        return loadNibNamed(name, owner: owner, options: options)?.first as? T
    }
}
```

gurgeous: That's nice too. No objection if you want to scrap this and add UIView.loadFromNib -> Self

guykogus: We're limited in this project to never call other functions because it's designed to be copyable piecemeal. It's a bit frustrating, but that seems to be the way that most people use it. Personally I think that the current implementation is great.
@gurgeous can you please just remove the comment from the test? Then I'll approve it :)

guykogus: And resolve conflicts, of course

gurgeous: 👍

CHANGELOG.md
```diff
@@ -32,6 +32,7 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
   - Added `isStandGoalMet`, `isExerciseTimeGoalMet`, and `isEnergyBurnedGoalMet`. [#875](https://github.com/SwifterSwift/SwifterSwift/pull/875) by [lhygilbert](https://github.com/lhygilbert)
 - **UIView**:
   - Added `masksToBounds` (IBInspectable) extension. [#877](https://github.com/SwifterSwift/SwifterSwift/pull/877) by [hamtiko](https://github.com/hamtiko)
+  - Added `loadFromNib(withClass:)`, which loads a UIView of a particular type from a nib file.
```
guykogus, line 35: I just realised a few of these are missing their PRs 😞
```suggestion
  - Added `loadFromNib(withClass:)`, which loads a UIView of a particular type from a nib file. [#885](https://github.com/SwifterSwift/SwifterSwift/pull/885) by [gurgeous](https://github.com/gurgeous)
```

PR #886, opened 2020-07-21
constraint finders (widthConstraint, heightConstraint, leadingConstraint, trailingConstraint, topConstraint, bottomConstraint)

gurgeous:
Handy helpers for finding existing constraints on a view. Under the hood they use `allConstraints`, which enumerates all ancestors to find all constraints for a view. See #879.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,7 +31,8 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
   - Added `isStandGoalMet`, `isExerciseTimeGoalMet`, and `isEnergyBurnedGoalMet`. [#875](https://github.com/SwifterSwift/SwifterSwift/pull/875) by [lhygilbert](https://github.com/lhygilbert)
 - **UIView**:
   - Added `masksToBounds` (IBInspectable) extension. [#877](https://github.com/SwifterSwift/SwifterSwift/pull/877) by [hamtiko](https://github.com/hamtiko)
-  - Added `widthConstraint`, `heightConstraint`, `leadingConstraint`, `trailingConstraint`, `topConstraint`, and `bottomConstraint` for finding existing constraints
+  - Added `findConstraint` for finding an existing constraint
+  - Added `widthConstraint`, `heightConstraint`, `leadingConstraint`, `trailingConstraint`, `topConstraint`, and `bottomConstraint` for finding specific constraints
 
 ### Changed
 - **NSAttributedString**:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
   - Added `isStandGoalMet`, `isExerciseTimeGoalMet`, and `isEnergyBurnedGoalMet`. [#875](https://github.com/SwifterSwift/SwifterSwift/pull/875) by [lhygilbert](https://github.com/lhygilbert)
 - **UIView**:
   - Added `masksToBounds` (IBInspectable) extension. [#877](https://github.com/SwifterSwift/SwifterSwift/pull/877) by [hamtiko](https://github.com/hamtiko)
+  - Added `widthConstraint`, `heightConstraint`, `leadingConstraint`, `trailingConstraint`, `topConstraint`, and `bottomConstraint` for finding existing constraints
 
 ### Changed
 - **NSAttributedString**:
--- a/Sources/SwifterSwift/UIKit/UIViewExtensions.swift
+++ b/Sources/SwifterSwift/UIKit/UIViewExtensions.swift
@@ -232,54 +232,6 @@ public extension UIView {
             frame.origin.y = newValue
         }
     }
-
-    /// SwifterSwift: Return constraints for this view. This must enumerate superviews.
-    var allConstraints: [NSLayoutConstraint] {
-        // https://stackoverflow.com/questions/47053727
-
-        // build array of self and all superviews
-        var views = [self]
-        var current = self
-        while let superview = current.superview {
-            views.append(superview)
-            current = superview
-        }
-
-        // now collect constraints affecting this view
-        return views.flatMap { $0.constraints }.filter {
-            $0.firstItem as? UIView == self || $0.secondItem as? UIView == self
-        }
-    }
-
-    /// SwifterSwift: First width constraint for this view
-    var widthConstraint: NSLayoutConstraint? {
-        findConstraint(attribute: .width)
-    }
-
-    /// SwifterSwift: First height constraint for this view
-    var heightConstraint: NSLayoutConstraint? {
-        findConstraint(attribute: .height)
-    }
-
-    /// SwifterSwift: First leading constraint for this view
-    var leadingConstraint: NSLayoutConstraint? {
-        findConstraint(attribute: .leading)
-    }
-
-    /// SwifterSwift: First trailing constraint for this view
-    var trailingConstraint: NSLayoutConstraint? {
-        findConstraint(attribute: .trailing)
-    }
-
-    /// SwifterSwift: First top constraint for this view
-    var topConstraint: NSLayoutConstraint? {
-        findConstraint(attribute: .top)
-    }
-
-    /// SwifterSwift: First bottom constraint for this view
-    var bottomConstraint: NSLayoutConstraint? {
-        findConstraint(attribute: .bottom)
-    }
 }
 
 // MARK: - Methods
@@ -626,6 +578,27 @@ public extension UIView {
     func ancestorView<T: UIView>(withClass name: T.Type) -> T? {
         return ancestorView(where: { $0 is T }) as? T
     }
+}
+
+// MARK: - findConstraint and friends
+public extension UIView {
+    /// SwifterSwift: Return constraints for this view. This must enumerate superviews.
+    var allConstraints: [NSLayoutConstraint] {
+        // https://stackoverflow.com/questions/47053727
+
+        // build array of self and all superviews
+        var views = [self]
+        var current = self
+        while let superview = current.superview {
+            views.append(superview)
+            current = superview
+        }
+
+        // now collect constraints affecting this view
+        return views.flatMap { $0.constraints }.filter {
+            $0.firstItem as? UIView == self || $0.secondItem as? UIView == self
+        }
+    }
 
     /// SwifterSwift: Search constraints for this view until we find one for the given attribute.
     ///
@@ -640,6 +613,36 @@ public extension UIView {
             ($0.firstAttribute == attribute && $0.firstItem as? UIView == self) ||
             ($0.secondAttribute == attribute && $0.secondItem as? UIView == self)
         }
+    }
+
+    /// SwifterSwift: First width constraint for this view
+    var widthConstraint: NSLayoutConstraint? {
+        findConstraint(attribute: .width)
+    }
+
+    /// SwifterSwift: First height constraint for this view
+    var heightConstraint: NSLayoutConstraint? {
+        findConstraint(attribute: .height)
+    }
+
+    /// SwifterSwift: First leading constraint for this view
+    var leadingConstraint: NSLayoutConstraint? {
+        findConstraint(attribute: .leading)
+    }
+
+    /// SwifterSwift: First trailing constraint for this view
+    var trailingConstraint: NSLayoutConstraint? {
+        findConstraint(attribute: .trailing)
+    }
+
+    /// SwifterSwift: First top constraint for this view
+    var topConstraint: NSLayoutConstraint? {
+        findConstraint(attribute: .top)
+    }
+
+    /// SwifterSwift: First bottom constraint for this view
+    var bottomConstraint: NSLayoutConstraint? {
+        findConstraint(attribute: .bottom)
     }
 }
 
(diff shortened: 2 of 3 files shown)


Sources/SwifterSwift/UIKit/UIViewExtensions.swift
```diff
@@ -580,6 +627,16 @@ public extension UIView {
         return ancestorView(where: { $0 is T }) as? T
     }
 
+    /// SwifterSwift: Search constraints for this view until we find one for the given attribute.
+    ///
+    /// - Parameter attribute: the attribute to find
+    /// - Returns: matching constraint
+    func findConstraint(attribute: NSLayoutConstraint.Attribute) -> NSLayoutConstraint? {
+        allConstraints.first {
+            ($0.firstAttribute == attribute && $0.firstItem as? UIView == self) ||
```
LucianoPAlmeida, line 636: All constraints already make this filter on == self right? Or am I missing something?

Sources/SwifterSwift/UIKit/UIViewExtensions.swift
```diff
@@ -233,6 +233,53 @@ public extension UIView {
...
+    var trailingConstraint: NSLayoutConstraint? {
+        findConstraint(attribute: .trailing)
+    }
+
+    /// SwifterSwift: First top constraint for this view
+    var topConstraint: NSLayoutConstraint? {
+        findConstraint(attribute: .top)
+    }
+
+    /// SwifterSwift: First bottom constraint for this view
```
LucianoPAlmeida, line 279: We normally have a rules where all extensions should be self-contained, but for these cases, since they are all related I'm not sure if we should apply it, because then the code will become such a mess ...

Sources/SwifterSwift/UIKit/UIViewExtensions.swift
```diff
@@ -233,6 +233,53 @@ public extension UIView {
...
     }
 
+    /// SwifterSwift: Return constraints for this view. This must enumerate superviews.
+    var allConstraints: [NSLayoutConstraint] {
+        // https://stackoverflow.com/questions/47053727
+
+        // build array of self and all superviews
+        var views = [self]
+        var current = self
+        while let superview = current.superview {
```
LucianoPAlmeida, line 243: Should this consider subviews here too?

LucianoPAlmeida: Not really sure ...

Sources/SwifterSwift/UIKit/UIViewExtensions.swift
```diff
@@ -233,6 +233,53 @@ public extension UIView {
         }
     }
 
+    /// SwifterSwift: Return constraints for this view. This must enumerate superviews.
+    var allConstraints: [NSLayoutConstraint] {
+        // https://stackoverflow.com/questions/47053727
```
LucianoPAlmeida, line 238: I'm going to use this to raise a concern here, I know we already have some StackOverflow extensions on the project, but recently I've been looking a lot at that software licensing stuff. So question is, should we allow this? Because all content published in SO, is under Creative Commons Attribution-ShareAlike 4.0 see [here](https://stackoverflow.com/help/licensing) for reference. And there are some requirements on [license conditions section](https://creativecommons.org/licenses/by-sa/4.0/legalcode) e.g. "if You Share Adapted Material You produce, the following conditions also apply... The Adapter’s License You apply must be a Creative Commons license with the same License Elements, this version or later, or a BY-SA Compatible License."
That makes me wonder if that means that to use this content SwifterSwift should be released with the same license? 

I really don't know, I'm just starting to look/study those things and also I'm no lawyer so just raising the point ... 
 
@gurgeous This is not related to your changes, I'm just picking this as an example :)

LucianoPAlmeida: cc @omaralbeik @guykogus

Sources/SwifterSwift/UIKit/UIViewExtensions.swift
gurgeous: This is a bit subtle. While looking for .height, it's entirely possible that firstAttribute is .height but firstItem is not self (secondItem would be self in that case). That's not what we want to return.

I'll add a comment.

Sources/SwifterSwift/UIKit/UIViewExtensions.swift
gurgeous: would it help to group these in an extension down below?

Sources/SwifterSwift/UIKit/UIViewExtensions.swift
gurgeous: I think constraints are always placed on the common ancestor (which is self or above). No need to check subviews.

Sources/SwifterSwift/UIKit/UIViewExtensions.swift
gurgeous: I'll let you guys hash it out, but here's what Jeff Atwood (Stack Overflow's founder) has to say about it:

https://meta.stackexchange.com/questions/25956/what-is-up-with-the-source-code-license-on-stack-overflow/25957#25957

> ...
> a snippet of code falls under excerpt category and thus should be free to use under fair use. Heck, we don't even support giant masses of code being posted, so to me, by definition, everything would be an excerpt. We're not sourceforge, github, or codeplex.
>
> I am not sure this is really a practical ongoing concern except for the truly paranoid
> ...
>
> \- Jeff Atwood

LucianoPAlmeida: Sure, this is a more general concern of mine, just wanted to bring up. To give you more context the company I work for have strict rules about "never copy code from the internet" because the code "is not owned by the company" so that's something I'm always paying attention to :) which may sound paranoid sometimes haha 
Probably not a big deal here, since as I mentioned we had some already, but just raising the point :)

Sources/SwifterSwift/UIKit/UIViewExtensions.swift
LucianoPAlmeida: Sure! Is better if related are all inn the same place :)

gurgeous: this is done, thanks

Sources/SwifterSwift/UIKit/UIViewExtensions.swift
```diff
@@ -579,7 +578,72 @@ public extension UIView {
...
+    /// SwifterSwift: Search constraints for this view until we find one for the given attribute.
+    ///
+    /// - Parameter attribute: the attribute to find
+    /// - Returns: matching constraint
+    func findConstraint(attribute: NSLayoutConstraint.Attribute) -> NSLayoutConstraint? {
+        // This is a bit subtle. While looking for .height, it's entirely
+        // possible that firstAttribute is .height but firstItem is not self
+        // (secondItem would be self in that case). That's not what we want to
+        // return.
+        allConstraints.first {
```
LucianoPAlmeida, line 612: Maybe it does make sense to not reuse `allConstraints` here, because if we implement the search here we can optimize and return early if we find then right away. What do you think?

Tests/UIKitTests/UIViewExtensionsTests.swift
```diff
@@ -470,6 +470,31 @@ final class UIViewExtensionsTests: XCTestCase {
         XCTAssertEqual(buttonSubview.ancestorView(withClass: UITableView.self), tableView)
     }
 
+    func testConstraintFinders() {
+        let container = UIView()
+        let view = UIView()
+        container.addSubview(view)
+
+        // setup constraints, some in container and some in view
```
LucianoPAlmeida, line 478: Can you please separate the test cases? Like one for all constraints, one for finding direct calls and one for computed properties?

Sources/SwifterSwift/UIKit/UIViewExtensions.swift
gurgeous: The constraint properties are the useful extensions here. allConstraints and findConstraint are just dependencies. The properties are unlikely to be performant because of how autolayout works, but we can certainly make them a bit faster by removing allConstraints entirely. I'll take a look.

gurgeous: how about this?

Tests/UIKitTests/UIViewExtensionsTests.swift
gurgeous: done

Sources/SwifterSwift/UIKit/UIViewExtensions.swift
gurgeous: btw, since allConstraints was removed there is no longer any meaningful code from stackoverflow in this extension

Sources/SwifterSwift/UIKit/UIViewExtensions.swift
```diff
@@ -579,7 +578,52 @@ public extension UIView {
     func ancestorView<T: UIView>(withClass name: T.Type) -> T? {
         return ancestorView(where: { $0 is T }) as? T
     }
+}
+
+// MARK: - findConstraint and friends
+public extension UIView {
+    /// SwifterSwift: Search constraints until we find one for the given view and attribute. This will enumerate ancestors since constraints are always added to the common ancestor.
```
LucianoPAlmeida, line 585: Can you break this into two lines? :)

CHANGELOG.md
```diff
@@ -31,6 +31,7 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
   - Added `isStandGoalMet`, `isExerciseTimeGoalMet`, and `isEnergyBurnedGoalMet`. [#875](https://github.com/SwifterSwift/SwifterSwift/pull/875) by [lhygilbert](https://github.com/lhygilbert)
 - **UIView**:
   - Added `masksToBounds` (IBInspectable) extension. [#877](https://github.com/SwifterSwift/SwifterSwift/pull/877) by [hamtiko](https://github.com/hamtiko)
+  - Added `widthConstraint`, `heightConstraint`, `leadingConstraint`, `trailingConstraint`, `topConstraint`, and `bottomConstraint` for finding existing constraints
```
LucianoPAlmeida, line 35: Should we add another line here for findConstraint since it is public?

Tests/UIKitTests/UIViewExtensionsTests.swift
```diff
@@ -470,6 +470,40 @@ final class UIViewExtensionsTests: XCTestCase {
         XCTAssertEqual(buttonSubview.ancestorView(withClass: UITableView.self), tableView)
     }
 
+    func testFindConstraint() {
+        let view = UIView()
+        let container = UIView()
+        container.addSubview(view)
+        view.widthAnchor.constraint(equalToConstant: 1).isActive = true
```
LucianoPAlmeida, line 477: Can you add a test case for a view that is not on the container hierarchy... a view that It is not a subview of container?

Tests/UIKitTests/UIViewExtensionsTests.swift
```diff
@@ -470,6 +470,40 @@ final class UIViewExtensionsTests: XCTestCase {
...
+        XCTAssertNil(view.findConstraint(view: view, attribute: .height))
+    }
+
+    func testConstraintProperties() {
+        let container = UIView()
+        let view = UIView()
+        container.addSubview(view)
+
+        // setup constraints, some in container and some in view
+        view.widthAnchor.constraint(equalToConstant: 1).isActive = true
```
LucianoPAlmeida, line 489: Nit: Let's prefer NSLayoutconstraint.activate([ ... ]), in other places too :)

Sources/SwifterSwift/UIKit/UIViewExtensions.swift
```diff
@@ -579,7 +578,52 @@ public extension UIView {
...
+    ///
+    /// - Parameter view: the view to find
+    /// - Parameter attribute: the attribute to find
+    /// - Returns: matching constraint
+    func findConstraint(view: UIView, attribute: NSLayoutConstraint.Attribute) -> NSLayoutConstraint? {
```
LucianoPAlmeida, line 590: What about changing this signature to `findConstraint(attribute:  NSLayoutConstraint.Attribute, at: UIView)` ?

Sources/SwifterSwift/UIKit/UIViewExtensions.swift
```diff
@@ -579,7 +578,52 @@ public extension UIView {
     func ancestorView<T: UIView>(withClass name: T.Type) -> T? {
         return ancestorView(where: { $0 is T }) as? T
     }
+}
+
+// MARK: - findConstraint and friends
```
LucianoPAlmeida, line 583:
```suggestion
// MARK: - Constraints
```

Sources/SwifterSwift/UIKit/UIViewExtensions.swift
gurgeous: Certainly! I was copying the style from elsewhere in the project, though. Another example where swiftformat would help.

Tests/UIKitTests/UIViewExtensionsTests.swift
gurgeous: sure

Sources/SwifterSwift/UIKit/UIViewExtensions.swift
gurgeous: `at` feels a little funny to me because it sort of implies that you are only searching "at" that view, when you are actually searching the entire ancestor chain. `for` or `forView` would probably be my preference. I don't have a strong opinion, though. I'll change it.

LucianoPAlmeida: Nice, 'for' would be great :)

LucianoPAlmeida: As the last thing could you just this? Your suggestion is indeed better

PR #889, opened 2020-07-28
Add XCTest extension for comparing Color objects

guykogus:
🚀
As this is our first XCTest extension it's a bit tricky to work out where everything goes, so this is mainly starting as a discussion.

## Checklist

### Targets
Source files that include `XCTest` can only be added to test targets.
This means that they can't be included in the frameworks in the Xcode project, which means that they can't be included in Carthage.
If a non-test target makes any reference to `XCTest`, even if it's protected by
```Swift
#if canImport(XCTest)
import XCTest
```
it will throw a build-time error saying **Failed to load module 'XCTest'**, so it will probably cause projects to fail building if we add it via CocoaPods and, probably, SPM.

### Location
Where do the source files belong? If we place them in `Sources/SwifterSwift/XCTest` then they will be included in the Swift package, unless we add that folder to the `excludes` list, which is easy enough.
However, if we add it to `Tests/XCTest` then people may miss those extensions because they never open that folder. Maybe not 🤷‍♂️

### Imports
In order to use `Color` in this example I need to call `@testable import SwifterSwift`, which may cause problems if we do find a way to make these extensions part of the framework. A solution would be to copy the `typealias`es.
Regardless, it's not really an issue unless we can actually include it in the library along with the other files anyway.

## New project
An alternative is to actually have a separate project, let's say called SwifterTests, which has SwifterSwift as a dependency. It would allow anyone who's mad enough to include the entire project as a dependency to include SwifterSwift for their project and SwifterTests to their test targets. E.g.

```ruby
target 'MyApp' do
   pod 'SwifterSwift'

   target 'MyAppTests' do
      pod 'SwifterTests'
   end
end
```
Obviously there are a bunch of issues here, the main one being visibility, but I think it's the cleanest option.

### DIFF
--- a/Tests/XCTest/XCTestExtensions.swift
+++ b/Tests/XCTest/XCTestExtensions.swift
@@ -10,7 +10,7 @@
 import XCTest
 @testable import SwifterSwift
 
-#if canImport(CoreGraphics)
+#if canImport(AppKit) || canImport(UIKit)
 import CoreGraphics
 
 public func XCTAssertEqual(_ expression1: @autoclosure () throws -> Color,
--- a/Tests/XCTest/XCTestExtensions.swift
+++ b/Tests/XCTest/XCTestExtensions.swift
@@ -22,7 +22,6 @@ public typealias Color = NSColor
 #endif
 
 #if canImport(AppKit) || canImport(UIKit)
-
 /// SwifterSwift: Asserts that two `Color`s are equal within a certain accuracy.
 /// - Parameters:
 ///   - expression1: A `Color`.
--- a/Tests/XCTest/XCTestExtensions.swift
+++ b/Tests/XCTest/XCTestExtensions.swift
@@ -8,11 +8,29 @@
 
 #if canImport(XCTest)
 import XCTest
-@testable import SwifterSwift
+
+#if canImport(UIKit)
+import UIKit
+/// SwifterSwift: Color
+public typealias Color = UIColor
+#endif
+
+#if canImport(AppKit) && !targetEnvironment(macCatalyst)
+import AppKit
+/// SwifterSwift: Color
+public typealias Color = NSColor
+#endif
 
 #if canImport(AppKit) || canImport(UIKit)
-import CoreGraphics
 
+/// SwifterSwift: Asserts that two `Color`s are equal within a certain accuracy.
+/// - Parameters:
+///   - expression1: A `Color`.
+///   - expression2: A `Color`.
+///   - accuracy: Describes the maximum difference between `expression1` and `expression2` for these values to be considered equal.
+///   - message: An optional description of the failure.
+///   - file: The file in which failure occurred. Defaults to the file name of the test case in which this function was called.
+///   - line: The line number on which failure occurred. Defaults to the line number on which this function was called.
 public func XCTAssertEqual(_ expression1: @autoclosure () throws -> Color,
                            _ expression2: @autoclosure () throws -> Color,
                            accuracy: CGFloat,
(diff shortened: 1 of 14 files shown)


gurgeous:
Just reading over your comment now... Since this is our first test extension, I'd probably vote for something cheap and cheerful:

- put the file in `Tests/XCTest`
- copy the `typealiases` so the file is standalone
- encourage people to paste this file into their project
- documentation

This is something we encourage with our extensions in general, right? Seems like a logical direction here, and sidesteps the build issues outlined above. If adoption is high and users are clamoring for a better solution, we can also investigate more complicated distribution mechanisms (Cocoapods, Carthage, SPM, etc)

LucianoPAlmeida:
@guykogus This still WIP? Let us know when it's ready 
Ahh also, we might need to add a new subspec on pods file :)

guykogus:
> @guykogus This still WIP? Let us know when it's ready
> Ahh also, we might need to add a new subspec on pods file :)

I was waiting to hear feedback from yourself and @omaralbeik. It's ready based on the current limitations.
Also, if we add the new subspec`pod lib lint` breaks because of the error I mentioned: **Failed to load module 'XCTest'**

guykogus:
BTW I'm looking forward to Xcode 12 going live so that we can have a `throw`ing version of this function.

LucianoPAlmeida:
> I was waiting to hear feedback from yourself and @omaralbeik. It's ready based on the current limitations.

Ah sorry, I saw the WIP so was actually waiting too ... haha

> Also, if we add the new subspec`pod lib lint` breaks because of the error I mentioned: **Failed to load module 'XCTest'**

Uhum, I've never run into this, but  my guess is that we have to tell the subspec that it does need the `XCTest` das dependency framework like this
https://github.com/Quick/Nimble/blob/2c6978e887d7caa8acc2cab7b38e89d4d63eafb6/Nimble.podspec#L40 
Have you test this?

Tests/XCTest/XCTestExtensions.swift
```diff
@@ -0,0 +1,54 @@
...
+#endif
+
+#if canImport(AppKit) && !targetEnvironment(macCatalyst)
+import AppKit
+/// SwifterSwift: Color
+public typealias Color = NSColor
+#endif
+
+#if canImport(AppKit) || canImport(UIKit)
+/// SwifterSwift: Asserts that two `Color`s are equal within a certain accuracy.
```
LucianoPAlmeida, line 25: I think we should document better what exactly the accuracy means, like what is the measurement used for determining a color value, in this case, the accuracy is applied on RGB values... because a Float accuracy in color does not make clear the behavior of the function, so it is important that is documented.

Tests/XCTest/XCTestExtensions.swift
```diff
@@ -0,0 +1,54 @@
...
+                           line: UInt = #line) {
+    var color1: Color!
+    XCTAssertNoThrow(color1 = try expression1(), message(), file: file, line: line)
+    var color2: Color!
+    XCTAssertNoThrow(color2 = try expression2(), message(), file: file, line: line)
+    var red1: CGFloat = 0, green1: CGFloat = 0, blue1: CGFloat = 0, alpha1: CGFloat = 0
+    var red2: CGFloat = 0, green2: CGFloat = 0, blue2: CGFloat = 0, alpha2: CGFloat = 0
+    color1.getRed(&red1, green: &green1, blue: &blue1, alpha: &alpha1)
+    color2.getRed(&red2, green: &green2, blue: &blue2, alpha: &alpha2)
+    XCTAssertEqual(red1, red2, accuracy: accuracy, message(), file: file, line: line)
```
LucianoPAlmeida, line 47: `message` in the `XCTAssertEqual` is also an autoclosure(I think) so we can just forward it instead of calling `message()`?

guykogus: The compiler throws an error if I send `message`, because the parameter is actually expecting `String`, not `() -> String`. Sending `message()` won't actually call the closure, it'll synthesise to passing `{ message() }` because of the `@autoclosure`.

PR #890, opened 2020-08-06
Added note for UIView.addShadow()

qchenqizhi:
Added a note that I often forget to UIView addShadow(ofColor: radius: offset: opacity:)

## Checklist

### DIFF
--- a/Sources/SwifterSwift/UIKit/UIViewExtensions.swift
+++ b/Sources/SwifterSwift/UIKit/UIViewExtensions.swift
@@ -270,7 +270,7 @@ public extension UIView {
 
     /// SwifterSwift: Add shadow to view.
     ///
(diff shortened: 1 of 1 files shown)


Sources/SwifterSwift/UIKit/UIViewExtensions.swift
```diff
@@ -270,11 +270,14 @@ public extension UIView {
 
     /// SwifterSwift: Add shadow to view.
     ///
+    /// - Note: This method only works with non-clear background color, or set a `shadowPath`.
```
LucianoPAlmeida, line 273: What does "or set a shadowPath" means, if I understood correctly this is "or if the view has a shadowPath set"? If is that what you mean can you update to make it clear? :)

qchenqizhi: Updated, thanks

PR #892, opened 2020-08-09
SwiftFormat

### LINKED ISSUE #887: swiftformat?
As long as I have some PRs hanging out there, here's another topic for discussion. Would you consider adopting [swiftformat](https://github.com/nicklockwood/SwiftFormat)? We use this quite happily in my current project alongside similar tools in (prettier, rubocop, etc).

Swiftformat would make it easier for newcomers (like me) to make contributions and not have to worry about whitespace/formatting issues. I spent a fair bit of time ironing out formatting issues for those PRs, unfortunately. I'm not familiar with Danger, but I imagine you could enforce via the Dangerfile and the PR checklist.

gurgeous:
Adds [SwiftFormat](https://github.com/nicklockwood/SwiftFormat), fixes #887. Here are some specific things to look at in the PR:

1. Initial set of .swiftformat rules. The presence of SwiftFormat is more important than the actual rules IMO.
1. Changes to .swiftlint.yml to ignore transient warnings that are fixed by running SwiftFormat.
1. New file `scripts/pre-commit`, a pre-commit hook that runs SwiftFormat. It includes installation instructions.
1. Modifications to swiftlint:disable statements to work around wrapping issues.

Future work to consider:

1. Document swiftformat for core devs.
1. Add a Brewfile. Swiftlint should go in there too.
1. Perhaps we should add `tx, ty, tz, sx, sy, sz, & t2` to the identifier_name exclusions.


gurgeous:
CI was complaining about the `.swiftversion` file, so I switched to `--swiftversion 5.1` in `.swiftformat`. This matches the podspec. Moving swiftformat to 5.1 also enabled it to ditch all those unnecessary return keywords.

Sources/SwifterSwift/AppKit/NSImageExtensions.swift
```diff
@@ -58,10 +52,10 @@ public extension NSImage {
         guard let data = tiffRepresentation else { return }
         guard let imageRep = NSBitmapImageRep(data: data) else { return }
 
-        guard let imageData = imageRep.representation(using: type, properties: [.compressionFactor: compressionFactor]) else { return }
+        guard let imageData = imageRep.representation(using: type, properties: [.compressionFactor: compressionFactor])
```
LucianoPAlmeida, line 55: humm... this formating style felt really weird to me, I would expect something like breaking at the begining

Sources/SwifterSwift/AppKit/NSViewExtensions.swift
```diff
@@ -29,7 +23,7 @@ public extension NSView {
     @IBInspectable
     var borderWidth: CGFloat {
         get {
-            return layer?.borderWidth ?? 0
```
LucianoPAlmeida, line 32: Personally, I'm not a big fan of this enforcement ... I think although the language allows this, I think we should use in very specific case(personally I would use it more in cases where we could do one-liner e.g. var x { a + b } ). But I'll let you guys decide on that :) To me if we would enforce this I think we should enforce at least to try let the single expression in the same line e.g. `get { layer?.borderWidth ?? 0 }` I think it would be a good way to go

Sources/SwifterSwift/Dispatch/DispatchQueueExtensions.swift
```diff
@@ -52,7 +45,8 @@ public extension DispatchQueue {
     func asyncAfter(delay: Double,
                     qos: DispatchQoS = .unspecified,
                     flags: DispatchWorkItemFlags = [],
-                    execute work: @escaping () -> Void) {
+                    execute work: @escaping () -> Void)
+    {
```
LucianoPAlmeida, line 49: Is this a bug, maybe line width maybe conflicting with the same-line bracet rule ... ?

Sources/SwifterSwift/Foundation/NSRegularExpressionExtensions.swift
```diff
@@ -29,30 +22,37 @@ public extension NSRegularExpression {
     func enumerateMatches(in string: String,
                           options: MatchingOptions = [],
                           range: Range<String.Index>,
-                          using block: @escaping (_ result: NSTextCheckingResult?, _ flags: MatchingFlags, _ stop: inout Bool) -> Void) {
+                          using block: @escaping (
```
LucianoPAlmeida, line 25: I like this very much! Line width is nice

Sources/SwifterSwift/AppKit/NSImageExtensions.swift
gurgeous: See: https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#elseonsameline

I personally like `--guardelse same-line`, though I also don't mind sticking to defaults.

image

Sources/SwifterSwift/AppKit/NSViewExtensions.swift
gurgeous: See: https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#redundantreturn

I think this is idiomatic Swift now and thus the default with SwiftFormat. Once the redundant returns are gone the remaining returns are clear indicators of early exit. I personally prefer this style, especially after working in other languages with implicit return (like Ruby).

Not a strong opinion, though. We can disable the `redundantReturn` rule if others agree. @guykogus? @omaralbeik? Chime in if you have opinions...

image

Sources/SwifterSwift/Dispatch/DispatchQueueExtensions.swift
gurgeous: I suspect it always uses a newline when parameter are wrapped. Probably to give a cleaner indent after the bracket. Not sure, though.

Sources/SwifterSwift/AppKit/NSViewExtensions.swift
LucianoPAlmeida: Yeah, I like it very much but in a case by case basis, as I mentioned we should have a convention to use as we see fit, but not enforce as a rule everywhere :)

gurgeous:
> Here are my 2 cents overall
> 
> 1. In general looks nice, I just have a minor concern about enforcing omitting the return. See comments bellow for details but TBH I would prefer not to enforce this.

This is worth a bit of discussion IMO. Let's give it a day and see if anyone else chimes in, otherwise we can disable it. It would probably be some work to restore the missing returns since running SwiftFormat again won't magically bring them back in the PR :)


> 2. There are some places where maxwidth restrictions force it to put the bracket in the line bellow which conflicts with `closingparen same-line` is that a bug?

I don't think so - see above. Let me know if you think this merits further investigation.


> 3. I'm not a big fan of those big refactors because I think this messes up with git history ... at least makes it a bit harder to git blame and track down PRs/commits where things were really implemented. I struggled a bit with this in the past. But I think there is no way around this right? any alternatives?

The perfectionist in me finds them frustrating as well, but the pragmatist just pushes on through. :) Despite my misgivings it's usually not much of a problem for projects like this. Especially when all the methods are nicely self contained and git blame is rarely required.

 
> But overall it looks great! Thank you for taking this on @gurgeous!

Thank you, my pleasure!! This is a fun and important project and I look forward to doing more.

gurgeous:
No further comments on this one, so let's finish it up. @LucianoPAlmeida let me know if you want me to disable `redundantReturn`. That will require redoing the PR (to restore the missing `returns`) but I'd be happy to tackle it if you think worthwhile.

Should we add a line to the changelog? It's mostly internal, not sure anyone will care apart from core devs.

LucianoPAlmeida:
> No further comments on this one, so let's finish it up. @LucianoPAlmeida let me know if you want me to disable `redundantReturn`. That will require redoing the PR (to restore the missing `returns`) but I'd be happy to tackle it if you think worthwhile.
> 

I was actually waiting on thoughts from the others @omaralbeik @guykogus, I'm +1 to disable it, but we normally like to have more opinions on stuff like this :)

> Should we add a line to the changelog? It's mostly internal, not sure anyone will care apart from core devs.

IMH no need for changelog here since doesn't affect users.

Sources/SwifterSwift/Foundation/NSRegularExpressionExtensions.swift
```diff
@@ -81,10 +82,11 @@ public extension NSRegularExpression {
...
                          options: MatchingOptions = [],
-                         range: Range<String.Index>) -> Int {
-        return numberOfMatches(in: string,
-                               options: options,
-                               range: NSRange(range, in: string))
+                         range: Range<String.Index>) -> Int
+    {
+        numberOfMatches(in: string,
+                        options: options,
+                        range: NSRange(range, in: string))
```
guykogus, line 89: This is what I meant before about redundant returns for functions. I find it very strange.

.swiftformat
```diff
@@ -0,0 +1,8 @@
+--closingparen same-line
+--commas inline
+--exclude Examples
+--header "{file} - Copyright {year} SwifterSwift"
+--ifdef no-indent
+--importgrouping testable-top
+--maxwidth 120
+--swiftversion 5.1
```
guykogus, line 8: Aren't we on 5.2?

guykogus (commented):
Good stuff! I don't love all the formatting but it's great to have something.
Also, is there some way to set up GitHub Actions to run swiftformat and ensure that there are no changes when calling `git status`? I did something like that at the last company I worked for and my old colleagues still hate me for it 😅

Sources/SwifterSwift/AppKit/NSImageExtensions.swift
gurgeous: Let's try `--guardelse same-line`

Sources/SwifterSwift/AppKit/NSViewExtensions.swift
gurgeous: I don't think we can customize it to that extent. Personally this doesn't bother me (because Ruby) but I don't have a strong opinion. Our convention is good even without autoformatting, I think. Let's turn off `redundantReturn`

Sources/SwifterSwift/Dispatch/DispatchQueueExtensions.swift
gurgeous: Maybe we can disable `wrapMultilineStatementBraces`

Sources/SwifterSwift/CoreLocation/CLLocationArrayExtensions.swift
```diff
@@ -22,12 +16,11 @@ public extension Array where Element: CLLocation {
             return Measurement(value: 0.0, unit: unit)
         }
         var distance: CLLocationDistance = 0.0
-        for idx in 0..<count-1 {
+        for idx in 0 ..< count - 1 {
```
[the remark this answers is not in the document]
gurgeous: Might be able to customize this with judicious use of `--nospaceoperators`. I will experiment and if it looks promising add to config.

Sources/SwifterSwift/Foundation/DateExtensions.swift
```diff
@@ -51,25 +42,24 @@ public extension Date {
...
 
 // MARK: - Properties
-public extension Date {
 
+public extension Date {
     /// SwifterSwift: User’s current calendar.
     var calendar: Calendar {
-        return Calendar(identifier: Calendar.current.identifier) // Workaround to segfault on corelibs foundation https://bugs.swift.org/browse/SR-10147
+        Calendar(identifier: Calendar.current
+            .identifier) // Workaround to segfault on corelibs foundation https://bugs.swift.org/browse/SR-10147
```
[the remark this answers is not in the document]
gurgeous: 👍

Sources/SwifterSwift/CoreAnimation/CATransform3DExtensions.swift
```diff
@@ -170,7 +160,8 @@ public extension CATransform3D {
     ///   - sz: z-axis scale
     /// - Returns: The scaled matrix.
     @inlinable
-    func scaledBy(x sx: CGFloat, y sy: CGFloat, z sz: CGFloat) -> CATransform3D { // swiftlint:disable:this identifier_name
+    // swiftlint:disable:next identifier_name
```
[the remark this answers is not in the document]
gurgeous: Here are a few ideas, not in order:

1. Move the comment up
1. Add sx/sy/sz/tx/ty/tz/t2... to our list of exclusions in the config.
1. Disable rule for the entire file.

Pick one!

.swiftformat
gurgeous: The podspec said 5.1, but I think we are building 5.2. I'll change it.

gurgeous:
The branch has been updated with swiftformat config changes in response to the PR feedback. I used the following:

```sh
$ git checkout upstream/master -- Sources Tests
$ swiftformat .
```

There are additional fixes that need to happen once we're happy with the formatting:

1. fix `swiftlint:disable identifier_name` after we pick a fix
1. the ugly wrapping in `DateExtensions`

I'll tackle those once we are done with the formatting. @LucianoPAlmeida & @guykogus can you take another look? I think this is getting close. Thanks!

guykogus (changes requested):
For CATransform3DExtensions let's just disable `identifier_name` so that we don't have to mark it for each instance.

Sources/SwifterSwift/CoreAnimation/CATransform3DExtensions.swift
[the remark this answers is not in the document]
guykogus: 3

gurgeous:
If there are no further comments I will apply those last two changes and prep for merge. @guykogus not sure about CI but I'd suggest we save that for a subsequent effort. I haven't used Github Actions yet... Hopefully we can start with the pre-commit hook. :) I also think we should add a Brewfile, but that can wait as well.

Brewfile
```diff
@@ -1,2 +1,5 @@
-brew "xctool"
+# install with `brew bundle`
+
+brew "swiftformat"
```
guykogus, line 3: No need for this comment
```suggestion
brew "swiftformat"
```

scripts/pre-commit
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+#
+# Run swiftformat on git commit
+# see: https://github.com/nicklockwood/SwiftFormat#git-pre-commit-hook
+#
+# Install with:
+#
+# $ brew install swiftformat
+# $ ln -sf ../../scripts/pre-commit .git/hooks/pre-commit
```
guykogus, line 10: I guess for us it's from the home folder, no?
```suggestion
# $ ln -sf scripts/pre-commit .git/hooks/pre-commit
```

guykogus (changes requested):
Looking great. Just a couple of little comments

scripts/pre-commit
gurgeous: I think it has to be relative to the location of `pre-commit`:

```sh
$ cd SwifterSwift
$ ln -sf scripts/pre-commit .git/hooks/pre-commit
$ .git/hooks/pre-commit
command not found: .git/hooks/pre-commit

$ cd SwifterSwift
$ ln -sf ../../scripts/pre-commit .git/hooks/pre-commit
$ .git/hooks/pre-commit
... runs properly ...
```

I find this counterintuitive as well and I messed it up the first time we tried this with our app 😄

guykogus: Ah I see, in the first instance you're creating a link that's trying to find `.git/hooks/scripts/pre-commit`, instead of `.git/hooks/../../scripts/pre-commit`. Makes sense now.

guykogus (approved):

LucianoPAlmeida:
cc @omaralbeik

omaralbeik (approved):
Nice work 🚀

PR #897, opened 2020-08-30
Rename conflicting UIView shadowExtensions

### LINKED ISSUE #622: shadowOffset for UILabel is conflict with UILabel's property.
[v] I've read, understood, and done my best to follow the [Contributing guidelines](https://github.com/SwifterSwift/SwifterSwift/tree/master/CONTRIBUTING.md) before opening this issue.

### What did you do?
Set shadowOffset, shadowColor, shadowRadius inspectable value on UILabel.

### What did you expect to happen?
UILabel with proper shadow

### What happened instead?
`shadowOffset` is not applied, because UILabel has `shadowOffset` property.
So shadow is always above text, because `layer.shadowOffset` default value is (0, -3)
(Also UILabel has `shadowColor` property.)

### SwifterSwift Environment

- SwifterSwift version: 4.6.0
- Xcode version: 10
- macOS version running Xcode: 10.14.2
- Swift version: 4.2
- Platform(s) running SwifterSwift: iOS

LucianoPAlmeida:
Fixes #622 

* This may cause issues in users who use those properties in storyboard as `@IBInspectable`.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,9 @@
 The changelog for **SwifterSwift**. Also see the [releases](https://github.com/SwifterSwift/SwifterSwift/releases) on GitHub.
 
 ## Upcoming Release
-
+### Breaking Change
+- **UIView**
+  - Rename `shadowColor`, `shadowOffset`, `shadowOpacity` and `shadowRadius` to `layerShadowColor`, `layerShadowOffset`, `layerShadowOpacity` and `layerShadowRadius` to avoid naming colisions with subclasses properties defined in other modules e.g. UIKit. [#897](https://github.com/SwifterSwift/SwifterSwift/pull/897) by [LucianoPAlmeida](https://github.com/LucianoPAlmeida)
 ### Added
 - **UIScrollView**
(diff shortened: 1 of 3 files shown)


LucianoPAlmeida:
cc @Jayxiang @multinerd

guykogus (commented):
Should we increment the major version here too?

LucianoPAlmeida:
> Should we increment the major version here too?

Not sure, since we are following up the swift versions the next one would be `5.3`, but anyways we should do that in the bump version/release PR

guykogus (approved):

Jayxiang:
Would it be better if renamed to `shadowColorOfLayer` or `shadowColorForLayer`, because it will be more clear on the Storyboard.
Storyboard

LucianoPAlmeida:
> Would it be better if renamed to `shadowColorOfLayer` or `shadowColorForLayer`, because it will be more clear on the Storyboard.

Hi @Jayxiang :)
Personally I prefer the original one as I think is less verbose on reading `view's layer shadow color`. But we can open a discussion since the release is not out yet.

Jayxiang:
Yes, I just found that it will not be displayed completely on the storyboard.
If we use `shadowColorOfLayer` or `shadowColorForLayer`, it can eliminate the ambiguity, and it will not be too difficult to distinguish on the storyboard. But it may be a bit verbose..
sb

PR #900, opened 2020-09-02
Fix imports for NSAttributedString extensions

guykogus:
🚀

## Checklist


LucianoPAlmeida (dismissed):

gurgeous:
This is an improvement. Minor quibble - `bolded` and `italicized` only work when you are using the system font, which is often not the case. For example, the last three apps that I've built all used custom fonts to avoid looking generic. Novice users who are new to attributed string will be particularly befuddled. Might be worth a disclaimer.

Sources/SwifterSwift/Foundation/NSAttributedStringExtensions.swift
```diff
@@ -14,10 +14,29 @@ import AppKit
...
+    #if !os(Linux)
     var bolded: NSAttributedString {
-        return applying(attributes: [.font: UIFont.boldSystemFont(ofSize: UIFont.systemFontSize)])
+        let pointSize: CGFloat
+        if !string.isEmpty,
+            let font = attribute(.font, at: 0, effectiveRange: nil) as? Font {
+            pointSize = font.pointSize
+        } else {
+            #if os(tvOS) || os(watchOS)
+            pointSize = Font.preferredFont(forTextStyle: .headline).pointSize
```
gurgeous, line 34: very minor, but should this be .body? same in `italicized`. It may not actually matter:

https://gist.github.com/zacwest/916d31da5d03405809c4#file-results-markdown

gurgeous (commented):
looks good!

Sources/SwifterSwift/Foundation/NSAttributedStringExtensions.swift
guykogus: This is the default in `tvOS`. In `watchOS` you can't really investigate, but it looks right.

Sources/SwifterSwift/Foundation/NSAttributedStringExtensions.swift
```diff
@@ -14,10 +14,29 @@ import AppKit
...
 public extension NSAttributedString {
-    #if os(iOS)
-    /// SwifterSwift: Bolded string.
+    #if canImport(UIKit)
+    private typealias Font = UIFont
+    #endif
+
+    #if canImport(AppKit) && !targetEnvironment(macCatalyst)
+    private typealias Font = NSFont
+    #endif
```
gurgeous, line 23: As a convention we might want to put these at the top of the file and make them `public`. I believe the compiler complains if you try to add a `Font` parameter to a func with a private typealias. I ran into this while experimenting with `applying`.

Sources/SwifterSwift/Foundation/NSAttributedStringExtensions.swift
```diff
@@ -28,45 +47,50 @@ public extension NSAttributedString {
...
-    #if os(iOS)
-    /// SwifterSwift: Italicized string.
+    #if canImport(UIKit)
+    /// SwifterSwift: Italicized string using the system font.
     var italicized: NSAttributedString {
-        return applying(attributes: [.font: UIFont.italicSystemFont(ofSize: UIFont.systemFontSize)])
+        let pointSize: CGFloat
+        if !string.isEmpty,
+            let font = attribute(.font, at: 0, effectiveRange: nil) as? UIFont {
+            pointSize = font.pointSize
```
gurgeous, line 56: This is fine. A `guard !isEmpty else { return self }` might be an improvement.

gurgeous (approved):
A few minor suggestions, but LGTM

Sources/SwifterSwift/Foundation/NSAttributedStringExtensions.swift
guykogus: I'd rather avoid doing that as it would conflict with [SwiftGen](https://github.com/SwiftGen/SwiftGen). But then again our `Color` definition also clashes with it, so... 🤷‍♂️ The SwiftGen configuration allows for changing the generated enum names, so it should be fine.

guykogus:
@LucianoPAlmeida @omaralbeik I made a tiny change here to make the unit tests ran in randomised order and parallel. The first for reliability that there's no interdependency between tests, and the second for performance (in fairness there isn't much - on my 8-core MBP there's like 0.5-1.0 a second improvement, on average).

Sources/SwifterSwift/Foundation/NSAttributedStringExtensions.swift
gurgeous: Hm. Good point. I wonder if the typealiases are namedspaced to the module. I have no clue, honestly, but I don't recall any conflicts when we pulled SwifterSwift into my app, which uses SwiftGen. If we think it's an issue maybe we should rename those to SColor/SFont? Just an idea.

LucianoPAlmeida (approved):
LGTM!

LucianoPAlmeida (approved):

Issue #896, opened 2020-08-30
proposed NSRegularExpression extensions


gurgeous:
I figured I'd open this up for discussion before putting together a PR. These are from my current project. What do you guys think?

### NSRegularExpression

```swift
// fatalError when bad pattern, useful for statics and other hardcoded regex
init(_ pattern: String, options: NSRegularExpression.Options = [])

// return true if the string matches
hasMatch(_ string: String) -> Bool

// run on entire string, annoying NSRange no longer required
enumerateMatches
firstMatch
matches
numberOfMatches
rangeOfFirstMatch
replaceMatches
stringByReplacingMatches
```

### String

```swift
// like matches(pattern)
matches(regex: NSRegularExpression) -> Bool

// regex.stringByReplacingMatches(self)
stringByReplacingMatches(regex: NSRegularExpression, withTemplate template: String) -> String
stringByReplacingMatches(pattern: String, withTemplate template: String) -> String
```

guykogus:
The initialiser isn't necessary - you can just use `try!` with the standard init.

I already created versions of each `NSRegularExpression` function that uses `Range<String.Index>` instead of `NSRange`. It would be nice to have no range value at all. As an option, instead of creating duplicates of each function, we can change the `range` parameters to be optional and use the string's full range in case they're `nil`.

For `String`  we already have `matches(pattern:)` and `~=`. You could add new `replacingOccurrences` functions (to maintain naming) such as
```Swift
func replacingOccurrences(of regex: NSRegularExpression, with template: String, range searchRange: Range<String.Index>) -> String
```

gurgeous:
> The initialiser isn't necessary - you can just use `try!` with the standard init.

This is annoying with swiftlint, but I don't feel strongly about it.

> I already created versions of each `NSRegularExpression` function that uses `Range<String.Index>` instead of `NSRange`. It would be nice to have no range value at all. As an option, instead of creating duplicates of each function, we can change the `range` parameters to be optional and use the string's full range in case they're `nil`.

👍 


> For `String` we already have `matches(pattern:)` and `~=`.

Hm. It's nice when these methods work with both NSRegularExpression and string pattern. The patterns are handy, but I often use the regex flavors for performance. That would be four methods total:

```swift
matches(regex)
matches(pattern) (we already have this)
~= regex
~= pattern (we already have this)
```


> You could add new `replacingOccurrences` functions (to maintain naming) such as
> 
> ```swift
[...]

PR #901, opened 2020-09-03
extensions & tests

gurgeous:
New String regex extensions as discussed in #896:

```swift
func matches(regex: NSRegularExpression) -> Bool
func replacingOccurrences(of pattern: String, with template: String, range searchRange: Range<String.Index>? = nil) -> String
func replacingOccurrences(of regex: NSRegularExpression, with template: String, range searchRange: Range<String.Index>? = nil) -> String
static func ~= (lhs: String, rhs: NSRegularExpression) -> Bool
```

Once I get the thumbs up I'll add the changelog.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 - **String**:
   - Overloaded Swift's 'contains' operator (`~=`) for `String` to check regex matching. [#858](https://github.com/SwifterSwift/SwifterSwift/pull/858) by [VatoKo](https://github.com/VatoKo)
   - `regexEscaped`, which returns an escaped string for inclusion in a regex pattern
+  - Added `matches(regex:options:)` and the `~= regex` to check directly against NSRegularExpression. Also added `replacingOccurrences(pattern:template:searchRange:)` and `replacingOccurrences(regex:template:searchRange:)` as a more convenient way to replace patterns and NSRegularExpressions. 
+[#901](https://github.com/SwifterSwift/SwifterSwift/pull/901) by [gurgeous](https://github.com/gurgeous)
 - **DispatchQueue**:
   - Added `asyncAfter(delay:qos:flags:execute:)` method to use it more conveniently without typing unnecessary `.now() + delay`. [#859](https://github.com/SwifterSwift/SwifterSwift/pull/859) by [VatoKo](https://github.com/VatoKo)
   - Re-added `debounce(delay:action:)` for only executing a closure once using a throttle delay. [#869](https://github.com/SwifterSwift/SwifterSwift/pull/869) by [guykogus](https://github.com/guykogus)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,7 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 - **String**:
   - Overloaded Swift's 'contains' operator (`~=`) for `String` to check regex matching. [#858](https://github.com/SwifterSwift/SwifterSwift/pull/858) by [VatoKo](https://github.com/VatoKo)
   - `regexEscaped`, which returns an escaped string for inclusion in a regex pattern
-  - Added `matches(regex:options:)` and the `~= regex` to check directly against NSRegularExpression. Also added `replacingOccurrences(regex:template:searchRange:)` as a more convenient way to replace NSRegularExpressions. [#901](https://github.com/SwifterSwift/SwifterSwift/pull/901) by [gurgeous](https://github.com/gurgeous)
+  - Added `matches(regex:options:)` and the `~= regex` to check directly against NSRegularExpression. Also added `replacingOccurrences(regex:template:options:searchRange:)` as a more convenient way to replace NSRegularExpressions. [#901](https://github.com/SwifterSwift/SwifterSwift/pull/901) by [gurgeous](https://github.com/gurgeous)
 - **DispatchQueue**:
   - Added `asyncAfter(delay:qos:flags:execute:)` method to use it more conveniently without typing unnecessary `.now() + delay`. [#859](https://github.com/SwifterSwift/SwifterSwift/pull/859) by [VatoKo](https://github.com/VatoKo)
   - Re-added `debounce(delay:action:)` for only executing a closure once using a throttle delay. [#869](https://github.com/SwifterSwift/SwifterSwift/pull/869) by [guykogus](https://github.com/guykogus)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,8 +20,7 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 - **String**:
   - Overloaded Swift's 'contains' operator (`~=`) for `String` to check regex matching. [#858](https://github.com/SwifterSwift/SwifterSwift/pull/858) by [VatoKo](https://github.com/VatoKo)
   - `regexEscaped`, which returns an escaped string for inclusion in a regex pattern
-  - Added `matches(regex:options:)` and the `~= regex` to check directly against NSRegularExpression. Also added `replacingOccurrences(pattern:template:searchRange:)` and `replacingOccurrences(regex:template:searchRange:)` as a more convenient way to replace patterns and NSRegularExpressions. 
-[#901](https://github.com/SwifterSwift/SwifterSwift/pull/901) by [gurgeous](https://github.com/gurgeous)
+  - Added `matches(regex:options:)` and the `~= regex` to check directly against NSRegularExpression. Also added `replacingOccurrences(regex:template:searchRange:)` as a more convenient way to replace NSRegularExpressions. [#901](https://github.com/SwifterSwift/SwifterSwift/pull/901) by [gurgeous](https://github.com/gurgeous)
 - **DispatchQueue**:
   - Added `asyncAfter(delay:qos:flags:execute:)` method to use it more conveniently without typing unnecessary `.now() + delay`. [#859](https://github.com/SwifterSwift/SwifterSwift/pull/859) by [VatoKo](https://github.com/VatoKo)
   - Re-added `debounce(delay:action:)` for only executing a closure once using a throttle delay. [#869](https://github.com/SwifterSwift/SwifterSwift/pull/869) by [guykogus](https://github.com/guykogus)
@@ -45,6 +44,8 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
   - Added `loadFromNib(withClass:)`, which loads a UIView of a particular type from a nib file. [#885](https://github.com/SwifterSwift/SwifterSwift/pull/885) by [gurgeous](https://github.com/gurgeous)
   - Added `findConstraint` for finding an existing constraint. [#886](https://github.com/SwifterSwift/SwifterSwift/pull/886) by [gurgeous]
   - Added `widthConstraint`, `heightConstraint`, `leadingConstraint`, `trailingConstraint`, `topConstraint`, and `bottomConstraint` for finding specific constraints. [#886](https://github.com/SwifterSwift/SwifterSwift/pull/886) by [gurgeous]
+- **StringProtocol**
+  - Added `replacingOccurrences(ofPattern:withTemplate:options:searchRange:)` as a more convenient way to replace patterns. [#901](https://github.com/SwifterSwift/SwifterSwift/pull/901) by [gurgeous](https://github.com/gurgeous)
 
 ### Changed
 - **NSAttributedString**:
--- a/Sources/SwifterSwift/SwiftStdlib/StringExtensions.swift
+++ b/Sources/SwifterSwift/SwiftStdlib/StringExtensions.swift
@@ -968,6 +968,17 @@ public extension String {
     #endif
 
     #if canImport(Foundation)
+    /// SwifterSwift: Verify if string matches the regex.
+    ///
+    /// - Parameter regex: Regex to verify.
+    /// - Returns: true if string matches the regex.
+    func matches(regex: NSRegularExpression) -> Bool {
+        let range = NSRange(location: 0, length: utf16.count)
+        return regex.firstMatch(in: self, options: [], range: range) != nil
+    }
+    #endif
+
+    #if canImport(Foundation)
     /// SwifterSwift: Overload Swift's 'contains' operator for matching regex pattern
     ///
     /// - Parameter lhs: String to check on regex pattern.
@@ -975,6 +986,50 @@ public extension String {
     /// - Returns: true if string matches the pattern.
     static func ~= (lhs: String, rhs: String) -> Bool {
         return lhs.range(of: rhs, options: .regularExpression) != nil
+    }
+    #endif
+
+    #if canImport(Foundation)
+    /// SwifterSwift: Overload Swift's 'contains' operator for matching regex
+    ///
+    /// - Parameter lhs: String to check on regex.
+    /// - Parameter rhs: Regex to match against.
+    /// - Returns: true if string matches the regex.
+    static func ~= (lhs: String, rhs: NSRegularExpression) -> Bool {
+        let range = NSRange(location: 0, length: lhs.utf16.count)
+        return rhs.firstMatch(in: lhs, options: [], range: range) != nil
+    }
+    #endif
+
+    #if canImport(Foundation)
+    /// SwifterSwift: Returns a new string in which all occurrences of a regex pattern in a specified range of the receiver are replaced by the template.
+    /// - Parameter pattern: Regex pattern to replace.
+    /// - Parameter template: The regex template to replace the pattern.
+    /// - Parameter searchRange: The range in the receiver in which to search.
+    /// - Returns: A new string in which all occurrences of regex pattern in searchRange of the receiver are replaced by template.
+    func replacingOccurrences(of pattern: String, with template: String,
+                              range searchRange: Range<String.Index>? = nil) -> String {
+        return replacingOccurrences(of: pattern, with: template, options: .regularExpression, range: searchRange)
+    }
+    #endif
+
+    #if canImport(Foundation)
+    /// SwifterSwift: Returns a new string in which all occurrences of a regex in a specified range of the receiver are replaced by the template.
+    /// - Parameter regex: Regex to replace.
+    /// - Parameter template: The template to replace the regex.
+    /// - Parameter searchRange: The range in the receiver in which to search.
+    /// - Returns: A new string in which all occurrences of regex in searchRange of the receiver are replaced by template.
+    func replacingOccurrences(
+        of regex: NSRegularExpression,
+        with template: String,
+        range searchRange: Range<String.Index>? = nil) -> String {
+        let range: NSRange
+        if let searchRange = searchRange {
+            range = NSRange(searchRange, in: self)
+        } else {
+            range = NSRange(location: 0, length: utf16.count)
+        }
+        return regex.stringByReplacingMatches(in: self, options: [], range: range, withTemplate: template)
     }
     #endif
 
--- a/Sources/SwifterSwift/SwiftStdlib/StringExtensions.swift
+++ b/Sources/SwifterSwift/SwiftStdlib/StringExtensions.swift
@@ -974,7 +974,7 @@ public extension String {
     /// - Parameter options: The matching options to use.
     /// - Returns: true if string matches the regex.
     func matches(regex: NSRegularExpression, options: NSRegularExpression.MatchingOptions = []) -> Bool {
-        let range = NSRange(location: 0, length: utf16.count)
+        let range = NSRange(startIndex..<endIndex, in: self)
         return regex.firstMatch(in: self, options: options, range: range) != nil
     }
     #endif
--- a/Sources/SwifterSwift/SwiftStdlib/StringExtensions.swift
+++ b/Sources/SwifterSwift/SwiftStdlib/StringExtensions.swift
@@ -961,7 +961,7 @@ public extension String {
     /// SwifterSwift: Verify if string matches the regex pattern.
     ///
     /// - Parameter pattern: Pattern to verify.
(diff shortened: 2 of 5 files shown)


Sources/SwifterSwift/SwiftStdlib/StringExtensions.swift
```diff
@@ -967,6 +967,17 @@ public extension String {
     }
     #endif
 
+    #if canImport(Foundation)
+    /// SwifterSwift: Verify if string matches the regex.
+    ///
+    /// - Parameter regex: Regex to verify.
+    /// - Returns: true if string matches the regex.
+    func matches(regex: NSRegularExpression) -> Bool {
+        let range = NSRange(location: 0, length: utf16.count)
```
LucianoPAlmeida, line 976: Why utf16?

Sources/SwifterSwift/SwiftStdlib/StringExtensions.swift
```diff
@@ -967,6 +967,17 @@ public extension String {
...
+        return regex.firstMatch(in: self, options: [], range: range) != nil
```
LucianoPAlmeida, line 977: Could options be an extra optional parameter with a default `[]`? That would be useful?

Tests/SwiftStdlibTests/StringExtensionsTests.swift
```diff
@@ -565,26 +565,44 @@ final class StringExtensionsTests: XCTestCase {
...
+        XCTAssert("email@mail.com" ~= emailPattern)
         XCTAssert("hat" ~= "[a-z]at")
         XCTAssertFalse("" ~= "[a-z]at")
         XCTAssert("" ~= "[a-z]*")
         XCTAssertFalse("" ~= "[0-9]+")
     }
     #endif
 
+    func testRegexMatchOperator() {
+        // swiftlint:disable:next force_try
```
LucianoPAlmeida, line 600: Could we just make the test function throws so we don't need to do any of this force `try!` and lint exclusion...?

Sources/SwifterSwift/SwiftStdlib/StringExtensions.swift
```diff
@@ -978,6 +989,50 @@ public extension String {
     }
     #endif
 
+    #if canImport(Foundation)
+    /// SwifterSwift: Overload Swift's 'contains' operator for matching regex
+    ///
+    /// - Parameter lhs: String to check on regex.
+    /// - Parameter rhs: Regex to match against.
+    /// - Returns: true if string matches the regex.
```
LucianoPAlmeida, line 997: I think this makes it clear 
```suggestion
    /// - Returns: true if there is at least one match for the regex in the string.
```

Sources/SwifterSwift/SwiftStdlib/StringExtensions.swift
```diff
@@ -978,6 +989,50 @@ public extension String {
...
+    /// - Parameter regex: Regex to replace.
+    /// - Parameter template: The template to replace the regex.
+    /// - Parameter searchRange: The range in the receiver in which to search.
+    /// - Returns: A new string in which all occurrences of regex in searchRange of the receiver are replaced by template.
+    func replacingOccurrences(
+        of regex: NSRegularExpression,
+        with template: String,
+        range searchRange: Range<String.Index>? = nil) -> String {
+        let range: NSRange
+        if let searchRange = searchRange {
```
LucianoPAlmeida, line 1027: Should we bound check the range? Or this is done by the `stringByReplacingMatches`?

Tests/SwiftStdlibTests/StringExtensionsTests.swift
```diff
@@ -565,26 +565,44 @@ final class StringExtensionsTests: XCTestCase {
...
         XCTAssertFalse("dasda".matches(pattern: "\\d{3}"))
-        XCTAssertFalse("notanemail.com".matches(pattern: "[A-Z0-9a-z._%+-]+@[A-Za-z0-9.-]+\\.[A-Za-z]{2,}"))
-        XCTAssert("email@mail.com".matches(pattern: "[A-Z0-9a-z._%+-]+@[A-Za-z0-9.-]+\\.[A-Za-z]{2,}"))
+        XCTAssertFalse("notanemail.com".matches(pattern: emailPattern))
+        XCTAssert("email@mail.com".matches(pattern: emailPattern))
+    }
+
+    func testRegexMatches() {
+        // swiftlint:disable force_try
+        XCTAssert("123".matches(regex: try! NSRegularExpression(pattern: "\\d{3}")))
```
LucianoPAlmeida, line 579: I know it was like this before, but I think will be an improvement to use `XCTAssertTrue` :)

Tests/SwiftStdlibTests/StringExtensionsTests.swift
```diff
@@ -860,4 +878,31 @@ final class StringExtensionsTests: XCTestCase {
...
     }
+
+    func testReplacingOccurrencesPattern() {
+        let string = "hello"
+        XCTAssertEqual("hello", string.replacingOccurrences(of: "not", with: "found"))
+        XCTAssertEqual("hexo", string.replacingOccurrences(of: "l+", with: "x"))
+        XCTAssertEqual("hellxo", string.replacingOccurrences(of: "(ll)", with: "$1x"))
+
+        let range = string.startIndex..<string.index(string.startIndex, offsetBy: 3)
+        XCTAssertEqual("hexlo", string.replacingOccurrences(of: "l", with: "x", range: range))
```
LucianoPAlmeida, line 889: Could you add some bound off and empty string test cases?

Sources/SwifterSwift/SwiftStdlib/StringExtensions.swift
gurgeous: I was under the impression that this was the right choice for converting String length to NSRange. I have seen this in loads of places, including SwifterSwift (see `isSpelledCorrectly`). I use this in my app, though we only support a small set of locales.

Is that incorrect?

Sources/SwifterSwift/SwiftStdlib/StringExtensions.swift
gurgeous: Pretty sure that's handled internally

Tests/SwiftStdlibTests/StringExtensionsTests.swift
gurgeous: I'll add empty tests. Invalid range is hard to test because you can't even create them with `offsetBy:' - it throws an unchecked exception if you try `offsetBy: 99`. Not sure how to test that one, but I'm open to suggestions... Maybe we don't need it.

Sources/SwifterSwift/SwiftStdlib/StringExtensions.swift
gurgeous: 👍

Tests/SwiftStdlibTests/StringExtensionsTests.swift
gurgeous: 👍

gurgeous: I had no idea you could do that, btw 😄

Sources/SwifterSwift/SwiftStdlib/StringExtensions.swift
LucianoPAlmeida: This is more a question, I'm not familiar enough with string encoding/representation, but to me just `count` would always return the correct length of the string and as far as I know strings are backed by a contiguous storage UTF-8 buffer... not really sure if this would make a difference... but again I'm not too familiar with string representation.

Tests/SwiftStdlibTests/StringExtensionsTests.swift
LucianoPAlmeida: Sure! I just suggested because normally in unit test we tend to think in add the most variable possibilities to make sure the expected behavior in most case we can :)

Sources/SwifterSwift/SwiftStdlib/StringExtensions.swift
gurgeous: I've never researched the topic in depth, but now I know the answer! Apple docs to the rescue. When converting to an NSString/NSRange API, utf16.count should be used. See:

https://developer.apple.com/documentation/swift/string/utf16view

I verified this in a playground:

image

Tests/SwiftStdlibTests/StringExtensionsTests.swift
gurgeous: No worries, I agree it's good practice and sometimes I need a nudge to avoid being lazy. 😄

Sources/SwifterSwift/SwiftStdlib/StringExtensions.swift
LucianoPAlmeida: Ah that makes sense! Thanks

guykogus:
```suggestion
        let range = NSRange(startIndex..<endIndex, in: self)
```
I really need to add these core `String` extensions, which are the correct way to convert between `Range<String.Index>` and `NSRange`.
```Swift
extension String {
    func range(nsRange: NSRange) -> Range<Index> {
        return Range(nsRange, in: self)!
    }

    func nsRange(range: Range<Index>) -> NSRange {
        return NSRange(range, in: self)
    }

    var fullRange: Range<Index> {
        return startIndex..<endIndex
    }

    var fullNSRange: NSRange {
        return nsRange(range: fullRange)
    }
}
```
The issue with using the `utf16` view is that it used to be correct to use the `utf8` view, but then that was changed (I believe when Swift 4 was released). Don't assume the underlying implementation, even though it's highly unlikely to change.

Sources/SwifterSwift/SwiftStdlib/StringExtensions.swift
```diff
@@ -967,6 +967,18 @@ public extension String {
     }
     #endif
 
+    #if canImport(Foundation)
+    /// SwifterSwift: Verify if string matches the regex.
+    ///
+    /// - Parameter regex: Regex to verify.
+    /// - Parameter options: The matching options to use.
+    /// - Returns: true if string matches the regex.
```
guykogus, line 975:
```suggestion
    /// - Returns: `true` if string matches the regex.
```

Sources/SwifterSwift/SwiftStdlib/StringExtensions.swift
```diff
@@ -978,6 +990,50 @@ public extension String {
...
 
+    #if canImport(Foundation)
+    /// SwifterSwift: Overload Swift's 'contains' operator for matching regex
+    ///
+    /// - Parameter lhs: String to check on regex.
+    /// - Parameter rhs: Regex to match against.
+    /// - Returns: true if there is at least one match for the regex in the string.
+    static func ~= (lhs: String, rhs: NSRegularExpression) -> Bool {
+        let range = NSRange(location: 0, length: lhs.utf16.count)
+        return rhs.firstMatch(in: lhs, options: [], range: range) != nil
```
guykogus, line 1001:
```suggestion
        return rhs.firstMatch(in: lhs, range: range) != nil
```
You can omit the default value.

Sources/SwifterSwift/SwiftStdlib/StringExtensions.swift
```diff
@@ -978,6 +990,50 @@ public extension String {
     }
     #endif
 
+    #if canImport(Foundation)
+    /// SwifterSwift: Overload Swift's 'contains' operator for matching regex
+    ///
+    /// - Parameter lhs: String to check on regex.
+    /// - Parameter rhs: Regex to match against.
+    /// - Returns: true if there is at least one match for the regex in the string.
```
guykogus, line 998:
```suggestion
    /// - Returns: `true` if there is at least one match for the regex in the string.
```

Sources/SwifterSwift/SwiftStdlib/StringExtensions.swift
```diff
@@ -978,6 +990,50 @@ public extension String {
...
+    static func ~= (lhs: String, rhs: NSRegularExpression) -> Bool {
+        let range = NSRange(location: 0, length: lhs.utf16.count)
```
guykogus, line 1000:
```suggestion
        let range = NSRange(startIndex..<endIndex, in: self)
```

Sources/SwifterSwift/SwiftStdlib/StringExtensions.swift
```diff
@@ -978,6 +990,50 @@ public extension String {
...
+    func replacingOccurrences(
+        of regex: NSRegularExpression,
+        with template: String,
+        range searchRange: Range<String.Index>? = nil) -> String {
+        let range: NSRange
+        if let searchRange = searchRange {
+            range = NSRange(searchRange, in: self)
+        } else {
+            range = NSRange(location: 0, length: utf16.count)
+        }
```
guykogus, line 1032:
```suggestion
        let range =NSRange(searchRange ?? startIndex..<endIndex, in: self)
```

Sources/SwifterSwift/SwiftStdlib/StringExtensions.swift
```diff
@@ -978,6 +990,50 @@ public extension String {
...
+    #endif
+
+    #if canImport(Foundation)
+    /// SwifterSwift: Returns a new string in which all occurrences of a regex pattern in a specified range of the receiver are replaced by the template.
+    /// - Parameter pattern: Regex pattern to replace.
+    /// - Parameter template: The regex template to replace the pattern.
+    /// - Parameter searchRange: The range in the receiver in which to search.
+    /// - Returns: A new string in which all occurrences of regex pattern in searchRange of the receiver are replaced by template.
+    func replacingOccurrences(of pattern: String, with template: String,
+                              range searchRange: Range<String.Index>? = nil) -> String {
```
guykogus, line 1012:
```suggestion
    func replacingOccurrences(of pattern: String,
                              with template: String,
                              range searchRange: Range<String.Index>? = nil) -> String {
```
Did you run `swiftformat`? 😛

guykogus (changes requested):

Sources/SwifterSwift/SwiftStdlib/StringExtensions.swift
gurgeous: That would be great! I have something similar in my app, though it uses utf16. As you correctly pointed out, these are more robust. Just wish we could use them in the extensions 😄

Sources/SwifterSwift/SwiftStdlib/StringExtensions.swift
gurgeous: Indeed, pre-commit hook installed! 🎉

guykogus: And it doesn't put each parameter on a new line? That's weird, no?

Sources/SwifterSwift/SwiftStdlib/StringExtensions.swift
```diff
@@ -978,6 +990,45 @@ public extension String {
...
+    }
+    #endif
+
+    #if canImport(Foundation)
+    /// SwifterSwift: Returns a new string in which all occurrences of a regex pattern in a specified range of the receiver are replaced by the template.
+    /// - Parameter pattern: Regex pattern to replace.
+    /// - Parameter template: The regex template to replace the pattern.
+    /// - Parameter searchRange: The range in the receiver in which to search.
+    /// - Returns: A new string in which all occurrences of regex pattern in searchRange of the receiver are replaced by template.
+    func replacingOccurrences(of pattern: String, with template: String,
```
guykogus, line 1011: 1. This can be generalised to `StringProtocol`.
2. Users should be able to pass `CompareOptions`. I've added that `assert` because the documentation says that only those options can be used with it.

```Swift
extension StringProtocol {
    #if canImport(Foundation)
    func replacingOccurrences<Target, Replacement>(
        ofPattern pattern: Target,
        withTemplate template: Replacement,
        options: String.CompareOptions = [.regularExpression],
        range searchRange: Range<Self.Index>? = nil) -> String where Target: StringProtocol, Replacement: StringProtocol {
        assert(options.isStrictSubset(of: [.regularExpression, .anchored, .caseInsensitive]), "Invalid options for regular expression replacement")
        return replacingOccurrences(of: pattern, with: template, options: options.union(.regularExpression), range: searchRange)
    }
    #endif
}
```

CHANGELOG.md
```diff
@@ -20,6 +20,8 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 - **String**:
   - Overloaded Swift's 'contains' operator (`~=`) for `String` to check regex matching. [#858](https://github.com/SwifterSwift/SwifterSwift/pull/858) by [VatoKo](https://github.com/VatoKo)
   - `regexEscaped`, which returns an escaped string for inclusion in a regex pattern
+  - Added `matches(regex:options:)` and the `~= regex` to check directly against NSRegularExpression. Also added `replacingOccurrences(pattern:template:searchRange:)` and `replacingOccurrences(regex:template:searchRange:)` as a more convenient way to replace patterns and NSRegularExpressions.
```
guykogus, line 23: If you take my suggested change, you'll also need to move this to `StringProtocol.
```suggestion
  - Added `matches(regex:options:)` and the `~= regex` to check directly against NSRegularExpression. Also added `replacingOccurrences(ofPattern:withTemplate:options:searchRange:)` and `replacingOccurrences(regex:template:searchRange:)` as a more convenient way to replace patterns and NSRegularExpressions. 
```

CHANGELOG.md
```diff
@@ -20,6 +20,8 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 - **String**:
   - Overloaded Swift's 'contains' operator (`~=`) for `String` to check regex matching. [#858](https://github.com/SwifterSwift/SwifterSwift/pull/858) by [VatoKo](https://github.com/VatoKo)
   - `regexEscaped`, which returns an escaped string for inclusion in a regex pattern
+  - Added `matches(regex:options:)` and the `~= regex` to check directly against NSRegularExpression. Also added `replacingOccurrences(pattern:template:searchRange:)` and `replacingOccurrences(regex:template:searchRange:)` as a more convenient way to replace patterns and NSRegularExpressions. 
+[#901](https://github.com/SwifterSwift/SwifterSwift/pull/901) by [gurgeous](https://github.com/gurgeous)
```
guykogus, line 24: This should not be on a separate line.

guykogus (changes requested):
Sorry, found a few more suggestions to make.

Sources/SwifterSwift/SwiftStdlib/StringExtensions.swift
gurgeous: It seems to take a simple approach that wraps if lines are too long. Maybe it can be configured. If it has to wrap I'd prefer one-per-line. Good news is that it will preserve newlines if you do this manually.

Sources/SwifterSwift/SwiftStdlib/StringExtensions.swift
gurgeous: Fine by me, thanks

gurgeous:
OK, I moved replacingOccurrences w/ pattern to StringProtocol. Tests and changelog were updated too.

Sources/SwifterSwift/SwiftStdlib/StringExtensions.swift
```diff
@@ -978,6 +990,33 @@ public extension String {
...
+    #if canImport(Foundation)
+    /// SwifterSwift: Returns a new string in which all occurrences of a regex in a specified range of the receiver are replaced by the template.
+    /// - Parameter regex: Regex to replace.
+    /// - Parameter template: The template to replace the regex.
+    /// - Parameter searchRange: The range in the receiver in which to search.
+    /// - Returns: A new string in which all occurrences of regex in searchRange of the receiver are replaced by template.
+    func replacingOccurrences(
+        of regex: NSRegularExpression,
+        with template: String,
+        range searchRange: Range<String.Index>? = nil) -> String {
```
guykogus, line 1014: You don't want `options` as a parameter of the function?
```suggestion
        with template: String,
        options: NSRegularExpression.MatchingOptions = [],
        range searchRange: Range<String.Index>? = nil) -> String {
```
If you use it make sure to update the CHANGELOG too.

Sources/SwifterSwift/SwiftStdlib/StringProtocolExtensions.swift
```diff
@@ -20,4 +20,29 @@ public extension StringProtocol {
...
+    func replacingOccurrences<Target, Replacement>(
+        ofPattern pattern: Target,
+        withTemplate template: Replacement,
+        options: String.CompareOptions = [.regularExpression],
+        range searchRange: Range<Self.Index>? = nil) -> String where Target: StringProtocol,
+        Replacement: StringProtocol {
+        assert(
+            options.isStrictSubset(of: [.regularExpression, .anchored, .caseInsensitive]),
+            "Invalid options for regular expression replacement")
+        print(options.union(.regularExpression))
```
guykogus, line 40: Remove print statement

guykogus (changes requested):

Sources/SwifterSwift/SwiftStdlib/StringExtensions.swift
gurgeous: Good idea, I'll toss it in there. Sometimes it's easy to get stuck in a rut applying specific changes without looking at the big picture. 😄

Sources/SwifterSwift/SwiftStdlib/StringProtocolExtensions.swift
gurgeous: Whoops, and that's why we love code reviews

guykogus: Indeed. Speaking of, do you mind checking mine? :)

guykogus (approved):
💯

guykogus (approved):

LucianoPAlmeida (approved):
Nice!

PR #911, opened 2020-10-17
Added/dictionary filter using multiple keys

MaratIbragimov:
## Checklist

- [] I have added a [changelog](https://github.com/SwifterSwift/SwifterSwift/blob/master/CHANGELOG_GUIDELINES.md) entry describing my changes.

### DIFF
--- a/Sources/SwifterSwift/SwiftStdlib/DictionaryExtensions.swift
+++ b/Sources/SwifterSwift/SwiftStdlib/DictionaryExtensions.swift
@@ -189,6 +189,33 @@ public extension Dictionary {
             }
         }
     }
+    
+    /// SwifterSwift: Filters out a new dictionary based on spericied `keys`
+    ///
+    ///        var dict =  ["key1": 1, "key2": 2, "key3": 3, "key4": 4]
+    ///        dict["key1", "key3", "key4"] -> ["key1": 1, "key3": 3, "key4": 4]
+    ///        dict["key2"] -> ["key2": 2]
+    ///
+    /// - Note: A new copy of a dictionary is created. if none of the keys are found an empty dictionary will be returned
+    ///
+    /// - Complexity: O(K), _K  is the length of the keys array and O(K) space.
+    ///
+    /// - Parameter keys: An array of keys to the desired value.
+    ///
+    /// - Returns: A new dictionary that contains specified  keys only . if none of the keys exists an empty dictioanry will be returned.
+    subscript(withKeys keys: Key...) ->  [Key: Value]{
+        
+        let uniqueKeys = keys.withoutDuplicates()
+        let newDict = uniqueKeys.reduce(into: [Key: Value]()) { res, item in
+            if let value = self[item] {
+                res[item] = value
+            }
+        }
+        
+        return newDict
+        
+        
+    }
 }
 
 // MARK: - Operators
--- a/Sources/SwifterSwift/SwiftStdlib/DictionaryExtensions.swift
+++ b/Sources/SwifterSwift/SwiftStdlib/DictionaryExtensions.swift
@@ -204,9 +204,7 @@ public extension Dictionary {
     ///
     /// - Returns: A new dictionary that contains specified  keys only . if none of the keys exists an empty dictioanry will be returned.
     subscript(withKeys keys: Key...) ->  [Key: Value]{
-        
-        let uniqueKeys = keys.withoutDuplicates()
-        let newDict = uniqueKeys.reduce(into: [Key: Value]()) { res, item in
+        let newDict = keys.reduce(into: [Key: Value]()) { res, item in
             if let value = self[item] {
                 res[item] = value
             }
--- a/Tests/SwiftStdlibTests/DictionaryExtensionsTests.swift
+++ b/Tests/SwiftStdlibTests/DictionaryExtensionsTests.swift
@@ -163,4 +163,20 @@ final class DictionaryExtensionsTests: XCTestCase {
             [6: ["Bryant"], 5: ["James"], 4: ["Wade", "John"], 0: ["", ""]])
         XCTAssertEqual(Dictionary(grouping: array3, by: \String.count), [:])
     }
+    
+    func testGetByKeys() {
+        let dict = ["James": 100,
+                    "Wade": 200,
+                    "Bryant": 500,
+                    "John": 600,
+                    "Jack": 1000]
+        
+        let res = dict[withKeys: "James", "Wade", "Jack"]
+        XCTAssertEqual(res, ["James": 100, "Wade": 200, "Jack": 1000])
+        
+        
+        let noResults = dict[withKeys: "Michael", "Joe", "Christian"]
+        XCTAssertTrue(noResults.isEmpty)
+    
+    }
 }


Sources/SwifterSwift/SwiftStdlib/DictionaryExtensions.swift
```diff
@@ -189,6 +189,28 @@ public extension Dictionary {
...
+    ///        dict["key2"] -> ["key2": 2]
+    ///
+    /// - Note: A new copy of a dictionary is created. if none of the keys are found an empty dictionary will be returned
+    ///
+    /// - Complexity: O(K), _K  is the length of the keys array and O(K) space.
+    ///
+    /// - Parameter keys: An array of keys to the desired value.
+    ///
+    /// - Returns: A new dictionary that contains specified  keys only . if none of the keys exists an empty dictioanry will be returned.
+    subscript(withKeys keys: Key...) -> [Key: Value] {
```
LucianoPAlmeida, line 206: Not sure having this as a subscript is an ideal API... This may probably be a better suit to go for a `filter(byKeys: Key...)`.
But at the same time, I'm not sure if such a version of `filter` is worth adding since we can achieve the same with very simple code 
```swift 
let dict = ["a": 1, "b": 2]
let keys = ["a"]
let filtered = dict.filter { keys.contains($0.key) }
print(filtered) // ["a": 1]
```
So this would feel like we are abstracting something that is very simple already...

MaratIbragimov: Hi @LucianoPAlmeida  
The main idea is to filter out a certain "slice" of a Dictionary based on a given array of keys only, `filter` is a good candidate, but in this case, it might not be an efficient solution, as we need to iterate over all of the `(key, value)` tuples, which takes `O(N)`, then for each iteration, we need to look up for the current `$0.key` in `keys` which takes `O(K)` where K is the number of keys, what gives in a total running time of `O(N*K)`.

The name and subscript can be changed to a function in case we will proceed with this PR :)

LucianoPAlmeida: I see that this makes sense, but still not sure about it being a subscript. IMO it should be a function...

gurgeous: This could be useful. I agree with Luciano - subscript is not a good fit. When in doubt, I look to lodash. Perhaps a function called `pick`? See https://lodash.com/docs/4.17.15#pick.

MaratIbragimov: @gurgeous 
Nice idea, I'll change it to be a function and name it as `pick`

guykogus: @MaratIbragimov to fix the `O(N*K)` issue you just need to create a `Set` so that it becomes `O(N+K)`.
```Swift
let keys = Set(["a"])
```

Tests/SwiftStdlibTests/DictionaryExtensionsTests.swift
```diff
@@ -163,4 +163,20 @@ final class DictionaryExtensionsTests: XCTestCase {
...
+        let dict = ["James": 100,
+                    "Wade": 200,
+                    "Bryant": 500,
+                    "John": 600,
+                    "Jack": 1000]
+        
+        let res = dict[withKeys: "James", "Wade", "Jack"]
+        XCTAssertEqual(res, ["James": 100, "Wade": 200, "Jack": 1000])
+        
+
```
guykogus, line 177: You have too many white spaces here. Can you please run `swiftformat`? There's a pre-commit hook you can install in `/scripts/pre-commit`.

Sources/SwifterSwift/SwiftStdlib/DictionaryExtensions.swift
MaratIbragimov: @guykogus indeed, that's what I did at the beginning, but then I changed it, because I actually have an O(1) access on the dictionary without additional allocation.
The only case I can think of using `Set` is when there are duplicated keys passed to the` function` / `subscript`

guykogus: @MaratIbragimov you're right. If you just traverse through the keys you can check the dictionary directly.
```Swift
func pick(keys: [Key]) -> Dictionary {
    return keys.reduce(into: Dictionary()) { result, key in
        result[key] = self[key]
    }
}
```
Otherwise you have to first go through all the keys and then through the entire dictionary.
```Swift
func pick(keys: [Key]) -> Dictionary {
    let keysSet = Set(keys)
    return filter { key, _ in keysSet.contains(key) }
}
```
If you want to get really fancy, you can make this a one-liner. It's a bit ugly though 😬
```Swift
func pick(keys: [Key]) -> Dictionary {
    return keys.reduce(into: Dictionary()) { $0[$1] = self[$1] }
}
```

guykogus:
Don't forget the CHANGELOG

Sources/SwifterSwift/SwiftStdlib/DictionaryExtensions.swift
MaratIbragimov: Thanks :), @guykogus I guess I'll go with your first suggestion, as it reads better in my opinion :)

LucianoPAlmeida (approved):
LGTM!

Sources/SwifterSwift/SwiftStdlib/DictionaryExtensions.swift
```diff
@@ -189,6 +189,25 @@ public extension Dictionary {
...
+    ///        dict.pick(keys: ["key1", "key3", "key4"]) -> ["key1": 1, "key3": 3, "key4": 4]
+    ///        dict.pick(keys: ["key2"]) -> ["key2": 2]
+    ///
+    /// - Note: Creats a new instance of Dictionary. If none of the keys was found in the original dictionary, an empty dictionary will be returned.
+    ///
+    /// - Complexity: O(K), _K  is the length of the keys array and O(K) space.
+    ///
+    /// - Parameter keys: An array of keys to the desired value.
+    ///
+    /// - Returns: A new dictionary that contains specified  keys only . if none of the keys exists an empty dictioanry will be returned.
```
gurgeous, line 205: Quick proofread:

```swift
    /// - Parameter keys: Array of keys to be picked
    ///
    /// - Returns: A new dictionary that contains only the specified keys.
```

You could probably remove the `Note`.

Sources/SwifterSwift/SwiftStdlib/DictionaryExtensions.swift
```diff
@@ -189,6 +189,25 @@ public extension Dictionary {
...
+    func pick(keys: [Key]) -> [Key: Value] {
+        keys.reduce(into: [Key: Value]()) { result, item in
+            result[item] = self[item]
+        }
+    }
```
gurgeous, line 210: I know how we got here, but this probably should be moved about of the subscript extension

Tests/SwiftStdlibTests/DictionaryExtensionsTests.swift
```diff
@@ -163,4 +163,16 @@ final class DictionaryExtensionsTests: XCTestCase {
...
+    func testGetByKeys() {
+        let dict = ["James": 100,
+                    "Wade": 200,
+                    "Bryant": 500,
+                    "John": 600,
+                    "Jack": 1000]
+        let res = dict.pick(keys: ["James", "Wade", "Jack"])
+        XCTAssertEqual(res, ["James": 100, "Wade": 200, "Jack": 1000])
+        let noResults = dict.pick(keys: ["Michael", "Joe", "Christian"])
+        XCTAssertTrue(noResults.isEmpty)
```
gurgeous, line 176: Can you add a `nil` test like:

```swift
["James": 100, "Wade": nil].pick(keys: ["James", "Wade"])
```

Pretty sure it'll work properly, but I can't say for certain without a test

gurgeous (changes requested):
Thanks Marat, and welcome to the project! All contributions are appreciated. We can be a little picky, so thank you for your patience :)

CHANGELOG.md
```diff
@@ -8,6 +8,9 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
   - Rename `shadowColor`, `shadowOffset`, `shadowOpacity` and `shadowRadius` to `layerShadowColor`, `layerShadowOffset`, `layerShadowOpacity` and `layerShadowRadius` to avoid naming colisions with subclasses properties defined in other modules e.g. UIKit. [#897](https://github.com/SwifterSwift/SwifterSwift/pull/897) by [LucianoPAlmeida](https://github.com/LucianoPAlmeida)
 
 ### Added
+- **Dictionary**
+    - Added `pick(keys:)`
+    [#911](https://github.com/SwifterSwift/SwifterSwift/pull/911) by [MaratIbragimov](https://github.com/MaratIbragimov)
```
guykogus, line 13: Add a short description and put it all on one line.
```suggestion
    - Added `pick(keys:)` <reason>. [#911](https://github.com/SwifterSwift/SwifterSwift/pull/911) by [MaratIbragimov](https://github.com/MaratIbragimov)
```

Sources/SwifterSwift/SwiftStdlib/DictionaryExtensions.swift
```diff
@@ -189,6 +189,25 @@ public extension Dictionary {
...
+
+    /// SwifterSwift: Creates a new dictionary using specified keys
+    ///
+    ///        var dict =  ["key1": 1, "key2": 2, "key3": 3, "key4": 4]
+    ///        dict.pick(keys: ["key1", "key3", "key4"]) -> ["key1": 1, "key3": 3, "key4": 4]
+    ///        dict.pick(keys: ["key2"]) -> ["key2": 2]
+    ///
+    /// - Note: Creats a new instance of Dictionary. If none of the keys was found in the original dictionary, an empty dictionary will be returned.
+    ///
+    /// - Complexity: O(K), _K  is the length of the keys array and O(K) space.
```
guykogus, line 201:
```suggestion
    /// - Complexity: O(K) where _K_  is the length of the `keys` array.
```

Tests/SwiftStdlibTests/DictionaryExtensionsTests.swift
guykogus: In order for that to work you'll probably need to declare `dict` as `[String: Int?]`

guykogus (changes requested):
A few little notes ;)

Tests/SwiftStdlibTests/DictionaryExtensionsTests.swift
MaratIbragimov: Yea, true, and then in this case we will have a results like `["James":Optional(100), "Wade": nil ]`

Sources/SwifterSwift/SwiftStdlib/DictionaryExtensions.swift
```diff
@@ -110,6 +110,23 @@ public extension Dictionary {
...
+    ///
+    ///        var dict =  ["key1": 1, "key2": 2, "key3": 3, "key4": 4]
+    ///        dict.pick(keys: ["key1", "key3", "key4"]) -> ["key1": 1, "key3": 3, "key4": 4]
+    ///        dict.pick(keys: ["key2"]) -> ["key2": 2]
+    ///
+    /// - Complexity: O(K), where _K_  is the length of the keys array and O(K) space.
+    ///
+    /// - Parameter keys: An array of keys to the desired value.
+    ///
+    /// - Returns: A new dictionary that contains specified  keys only . if none of the keys exists an empty dictioanry will be returned.
```
LucianoPAlmeida, line 124:
```suggestion
    /// - Returns: A new dictionary that contains specified keys only. If none of the keys exists an empty dictionary will be returned.
```

Sources/SwifterSwift/SwiftStdlib/DictionaryExtensions.swift
```diff
@@ -110,6 +110,23 @@ public extension Dictionary {
...
+
+    /// SwifterSwift: Creates a new dictionary using specified keys
+    ///
+    ///        var dict =  ["key1": 1, "key2": 2, "key3": 3, "key4": 4]
+    ///        dict.pick(keys: ["key1", "key3", "key4"]) -> ["key1": 1, "key3": 3, "key4": 4]
+    ///        dict.pick(keys: ["key2"]) -> ["key2": 2]
+    ///
+    /// - Complexity: O(K), where _K_  is the length of the keys array and O(K) space.
+    ///
+    /// - Parameter keys: An array of keys to the desired value.
```
LucianoPAlmeida, line 122:
```suggestion
    /// - Parameter keys: An array of keys of the entries that will be in the resulting dictionary.
```

Tests/SwiftStdlibTests/DictionaryExtensionsTests.swift
```diff
@@ -163,4 +163,29 @@ final class DictionaryExtensionsTests: XCTestCase {
...
+
+        let optionalValuesDict = ["James": 100,
+                                  "Wade": nil,
+                                  "Bryant": 500,
+                                  "John": nil,
+                                  "Jack": 1000]
+
+        let pickedWithOptioanls = optionalValuesDict.pick(keys: ["James", "Bryant", "John"])
```
LucianoPAlmeida, line 184: Can you just add an additional test with a key that doesn't exist in the original dictionary?

Sources/SwifterSwift/SwiftStdlib/DictionaryExtensions.swift
```diff
@@ -110,6 +110,23 @@ public extension Dictionary {
...
+
+    /// SwifterSwift: Creates a new dictionary using specified keys
+    ///
+    ///        var dict =  ["key1": 1, "key2": 2, "key3": 3, "key4": 4]
+    ///        dict.pick(keys: ["key1", "key3", "key4"]) -> ["key1": 1, "key3": 3, "key4": 4]
+    ///        dict.pick(keys: ["key2"]) -> ["key2": 2]
+    ///
+    /// - Complexity: O(K), where _K_  is the length of the keys array and O(K) space.
+    ///
+    /// - Parameter keys: Parameter keys: An array of keys of the entries that will be in the resulting dictionary.
```
LucianoPAlmeida, line 122:
```suggestion
    /// - Parameter keys: An array of keys of the entries that will be in the resulting dictionary.
```

guykogus (changes requested):
Please use `swiftformat` to fix the tests coding style.

LucianoPAlmeida (approved):

LucianoPAlmeida (dismissed):

MaratIbragimov:
@guykogus  can you please review it, I pushed the latest changes with `swiftformatter`

Sources/SwifterSwift/SwiftStdlib/DictionaryExtensions.swift
```diff
@@ -110,6 +110,23 @@ public extension Dictionary {
...
         return [K: V](uniqueKeysWithValues: try compactMap(transform))
     }
+
+    /// SwifterSwift: Creates a new dictionary using specified keys
+    ///
+    ///        var dict =  ["key1": 1, "key2": 2, "key3": 3, "key4": 4]
+    ///        dict.pick(keys: ["key1", "key3", "key4"]) -> ["key1": 1, "key3": 3, "key4": 4]
+    ///        dict.pick(keys: ["key2"]) -> ["key2": 2]
+    ///
+    /// - Complexity: O(K), where _K_  is the length of the keys array and O(K) space.
```
guykogus, line 120:
```suggestion
    /// - Complexity: O(K), where _K_ is the length of the keys array.
```

Sources/SwifterSwift/SwiftStdlib/DictionaryExtensions.swift
```diff
@@ -110,6 +110,23 @@ public extension Dictionary {
...
+    ///
+    /// - Parameter keys: An array of keys of the entries that will be in the resulting dictionary.
```
guykogus, line 122: This reads better.
```suggestion
    /// - Parameter keys: An array of keys that will be the entries in the resulting dictionary.
```

Sources/SwifterSwift/SwiftStdlib/DictionaryExtensions.swift
```diff
@@ -110,6 +110,23 @@ public extension Dictionary {
...
+    ///
+    /// - Returns: A new dictionary that contains specified keys only. If none of the keys exists an empty dictionary will be returned.
```
guykogus, line 124:
```suggestion
    /// - Returns: A new dictionary that contains the specified keys only. If none of the keys exist, an empty dictionary will be returned.
```

Tests/SwiftStdlibTests/DictionaryExtensionsTests.swift
```diff
@@ -163,4 +163,31 @@ final class DictionaryExtensionsTests: XCTestCase {
...
+
+    func testGetByKeys() {
+        let dict = ["James": 100,
+                    "Wade": 200,
+                    "Bryant": 500,
+                    "John": 600,
+                    "Jack": 1000]
+        let picked = dict.pick(keys: ["James", "Wade", "Jack"])
+        let empty1 = dict.pick(keys: ["Pippen", "Rodman"])
+        let empty2 = dict.pick(keys: ["Michael", "Joe", "Christian"])
```
guykogus, line 175: No need for `empty2`, you've already tested the scenario with `empty1`. What maybe you should have is a mixed one, where some kys are found and others are not.

guykogus (changes requested):

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
   - Rename `shadowColor`, `shadowOffset`, `shadowOpacity` and `shadowRadius` to `layerShadowColor`, `layerShadowOffset`, `layerShadowOpacity` and `layerShadowRadius` to avoid naming colisions with subclasses properties defined in other modules e.g. UIKit. [#897](https://github.com/SwifterSwift/SwifterSwift/pull/897) by [LucianoPAlmeida](https://github.com/LucianoPAlmeida)
 
 ### Added
+- **Dictionary**
+    - Added `pick(keys:)` to pick part of a dictioanry with specified keys. [#911](https://github.com/SwifterSwift/SwifterSwift/pull/911) by [MaratIbragimov](https://github.com/MaratIbragimov)
```
guykogus, line 12:
```suggestion
  - Added `pick(keys:)` to pick part of a dictioanry with specified keys. [#911](https://github.com/SwifterSwift/SwifterSwift/pull/911) by [MaratIbragimov](https://github.com/MaratIbragimov)
```

guykogus (changes requested):

guykogus (approved):

PR #912, opened 2020-10-20
Improve `Sequence` extensions with cleaner code and performance improvements

guykogus:
Also added `contains(_:)` for `Hashable` elements

## Checklist

### DIFF
--- a/Sources/SwifterSwift/SwiftStdlib/SequenceExtensions.swift
+++ b/Sources/SwifterSwift/SwiftStdlib/SequenceExtensions.swift
@@ -38,10 +38,16 @@ public extension Sequence {
     ///
     ///        [2, 2, 4, 7].last(where: {$0 % 2 == 0}) -> 4
     ///
-    /// - Parameter condition: condition to evaluate each element against.
+    /// - Parameter predicate: condition to evaluate each element against.
     /// - Returns: the last element in the array matching the specified condition. (optional)
-    func last(where condition: (Element) throws -> Bool) rethrows -> Element? {
-        return try reversed().first(where: condition)
+    func last(where predicate: (Element) throws -> Bool) rethrows -> Element? {
+        var lastElement: Element?
+        for element in self {
+            if try predicate(element) {
+                lastElement = element
+            }
+        }
+        return lastElement
     }
 
     /// SwifterSwift: Filter elements based on a rejection condition.
--- a/Sources/SwifterSwift/SwiftStdlib/SequenceExtensions.swift
+++ b/Sources/SwifterSwift/SwiftStdlib/SequenceExtensions.swift
@@ -41,10 +41,7 @@ public extension Sequence {
     /// - Parameter condition: condition to evaluate each element against.
     /// - Returns: the last element in the array matching the specified condition. (optional)
     func last(where condition: (Element) throws -> Bool) rethrows -> Element? {
-        for element in reversed() {
-            if try condition(element) { return element }
-        }
-        return nil
+        return try reversed().first(where: condition)
     }
 
     /// SwifterSwift: Filter elements based on a rejection condition.
@@ -88,9 +85,7 @@ public extension Sequence {
     ///   - condition: condition to evaluate each element against.
     ///   - body: a closure that takes an element of the array as a parameter.
     func forEach(where condition: (Element) throws -> Bool, body: (Element) throws -> Void) rethrows {
-        for element in self where try condition(element) {
-            try body(element)
-        }
+        try lazy.filter(condition).forEach(body)
     }
 
     /// SwifterSwift: Reduces an array while returning each interim combination.
@@ -118,12 +113,7 @@ public extension Sequence {
     ///   - transform: transform element function to evaluate every element.
     /// - Returns: Return an filtered and mapped array.
     func filtered<T>(_ isIncluded: (Element) throws -> Bool, map transform: (Element) throws -> T) rethrows -> [T] {
-        return try compactMap {
-            if try isIncluded($0) {
-                return try transform($0)
-            }
-            return nil
-        }
+        return try lazy.filter(isIncluded).map(transform)
     }
 
     /// SwifterSwift: Get the only element based on a condition.
@@ -169,14 +159,13 @@ public extension Sequence {
     /// - Returns: A tuple of matched and non-matched items
     func divided(by condition: (Element) throws -> Bool) rethrows -> (matching: [Element], nonMatching: [Element]) {
         // Inspired by: http://ruby-doc.org/core-2.5.0/Enumerable.html#method-i-partition
-        var matching = ContiguousArray<Element>()
-        var nonMatching = ContiguousArray<Element>()
+        var matching = [Element]()
+        var nonMatching = [Element]()
 
-        var iterator = makeIterator()
-        while let element = iterator.next() {
+        for element in self {
             try condition(element) ? matching.append(element) : nonMatching.append(element)
         }
-        return (Array(matching), Array(nonMatching))
+        return (matching, nonMatching)
     }
 
     /// SwifterSwift: Return a sorted array based on a key path and a compare function.
@@ -259,7 +248,7 @@ public extension Sequence {
     ///   - value: The value to compare with `Element` property
     /// - Returns: The last element of the collection that has property by given key path equals to given `value` or `nil` if there is no such element.
     func last<T: Equatable>(where keyPath: KeyPath<Element, T>, equals value: T) -> Element? {
-        return last { $0[keyPath: keyPath] == value }
+        return reversed().first { $0[keyPath: keyPath] == value }
     }
 }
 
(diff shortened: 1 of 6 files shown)


Sources/SwifterSwift/SwiftStdlib/SequenceExtensions.swift
```diff
@@ -259,7 +248,7 @@ public extension Sequence {
     ///   - value: The value to compare with `Element` property
     /// - Returns: The last element of the collection that has property by given key path equals to given `value` or `nil` if there is no such element.
     func last<T: Equatable>(where keyPath: KeyPath<Element, T>, equals value: T) -> Element? {
-        return last { $0[keyPath: keyPath] == value }
```
guykogus, line 262: This was referencing an internal function, which is against our guidelines.

Sources/SwifterSwift/SwiftStdlib/SequenceExtensions.swift
```diff
@@ -41,10 +41,7 @@ public extension Sequence {
     /// - Parameter condition: condition to evaluate each element against.
     /// - Returns: the last element in the array matching the specified condition. (optional)
     func last(where condition: (Element) throws -> Bool) rethrows -> Element? {
-        for element in reversed() {
-            if try condition(element) { return element }
-        }
-        return nil
+        return try reversed().first(where: condition)
```
LucianoPAlmeida, line 44: Seems like there's no much we could do here to avoid the copy of `reversed()` the only other implementation I could think of would be iterate forward over all items ... so we would basically trade avoid copy for a slower version ...

Sources/SwifterSwift/SwiftStdlib/SequenceExtensions.swift
```diff
@@ -118,12 +113,7 @@ public extension Sequence {
     ///   - transform: transform element function to evaluate every element.
     /// - Returns: Return an filtered and mapped array.
     func filtered<T>(_ isIncluded: (Element) throws -> Bool, map transform: (Element) throws -> T) rethrows -> [T] {
-        return try compactMap {
-            if try isIncluded($0) {
-                return try transform($0)
-            }
-            return nil
-        }
+        return try lazy.filter(isIncluded).map(transform)
```
LucianoPAlmeida, line 116: Nice!

Sources/SwifterSwift/SwiftStdlib/SequenceExtensions.swift
```diff
@@ -316,15 +314,15 @@ public extension Sequence where Element: Hashable {
...
-public extension Sequence where Element: Numeric {
+public extension Sequence where Element: AdditiveArithmetic {
     /// SwifterSwift: Sum of all elements in array.
     ///
     ///        [1, 2, 3, 4, 5].sum() -> 15
     ///
     /// - Returns: sum of the array's elements.
     func sum() -> Element {
-        return reduce(into: 0, +=)
+        return reduce(.zero, +)
```
LucianoPAlmeida, line 326: Why this? To me, this seems like a regression in performance, minor but still ...

LucianoPAlmeida (approved):

Sources/SwifterSwift/SwiftStdlib/SequenceExtensions.swift
guykogus: Some sequences have a lazy-esque version of `reversed`. I'd like to think that Swift will call them, instead of creating the copy.

Sources/SwifterSwift/SwiftStdlib/SequenceExtensions.swift
guykogus: `0` is an `Int`, `.zero` is the actual element type.
`reduce(into:_:)` passes an `inout` variable, which means that there's added indirection via a reference. Since it's (probably) just a simple value like an `Int` it should be faster to copy the value instead of needing to dereference it.

LucianoPAlmeida: Yeah, it shouldn't be a huge problem in this case I think

Sources/SwifterSwift/SwiftStdlib/SequenceExtensions.swift
LucianoPAlmeida: Yeah, but I think the only other version we have of `reversed` that is a lazy view is down to `BidirectionalCollection` ... in the case of this function the selected overload will always be this copy one of `Sequence`... but for this case is fine since `BidirectionalCollection` has it's own version of `last(where:)` which is index based and can iterate backwards.

LucianoPAlmeida:
We have to check what is going on with the CI, seems like all PRs are failing

Sources/SwifterSwift/SwiftStdlib/SequenceExtensions.swift
guykogus: Makes sense that it can only from bidirectional it can be lazified. I guess better to improve performance than write shorter code.

guykogus:
@LucianoPAlmeida it turns out that `last(where:)` wasn't actually being tested because it was running on an `Array`, which already has a built-in definition (thanks for pointing that out). I created a `LinkedList`, which is ordered and only conforms to `Sequence`, so that we could test it. A little overkill, I know... 😅

LucianoPAlmeida:
> @LucianoPAlmeida it turns out that `last(where:)` wasn't actually being tested because it was running on an `Array`, which already has a built-in definition (thanks for pointing that out). 

Nice! 

> I created a `LinkedList`, which is ordered and only conforms to `Sequence`, so that we could test it. A little overkill, I know... 😅

I believe we don't need this, we can test it with a `Set<>` because it is a `Sequence` but not a collection, so a believe it will be able to handle this :)

guykogus:
> I believe we don't need this, we can test it with a `Set<>` because it is a `Sequence` but not a collection, so a believe it will be able to handle this :) Although is not ordered we can still test using the right elements and predicates right?

No, the order when iterating isn't fixed because the hash values are randomised. Maybe we could have a fixed hashed value object...

LucianoPAlmeida:
> No, the order when iterating isn't fixed because the hash values are randomised. Maybe we could have a fixed hashed value object...

I think even then ordering is not guaranteed ... so perhaps we have no choice with this case then

Tests/SwiftStdlibTests/SequenceExtensionsTests.swift
```diff
@@ -97,7 +155,22 @@ final class SequenceExtensionsTests: XCTestCase {
         XCTAssertEqual(tuple.1, [1, 3, 5])
     }
 
-    func testContains() {
+    func testContainsEquatable() {
+        struct Foo: Equatable, ExpressibleByIntegerLiteral {
```
LucianoPAlmeida, line 159: Seems swiftlint is complaining about nesting here ...

LucianoPAlmeida:
Better thinking about it today, a `last(where:)` for `Sequence` kinda don't make sense in semantic terms, because the protocol as in general is a protocol that doesn't guarantee Ordering so `last` will not produce an "expected" result. Point is that we asking the last of a Sequence ... but what is exactly the last in the sense of the protocol? 
The best example of this is the `Set` one, it feels a little weird to call `set.last { ... }`

Maybe we should think about deprecating this one.

guykogus:
> Better thinking about it today, a `last(where:)` for `Sequence` kinda don't make sense in semantic terms, because the protocol as in general is a protocol that doesn't guarantee Ordering so `last` will not produce an "expected" result. Point is that we asking the last of a Sequence ... but what is exactly the last in the sense of the protocol?
> The best example of this is the `Set` one, it feels a little weird to call `set.last { ... }`
> 
> Maybe we should think about deprecating this one.

I think I agree, actually. `Sequence` has a `first(where:)`, but that would be better named `any(where:)`. I'm happy to deprecate this.

LucianoPAlmeida:
> I think I agree, actually. `Sequence` has a `first(where:)`, but that would be better named `any(where:)`. I'm happy to deprecate this.

Great, let's do this \o/
About `first(where:)` do we have that? I thought it was from stdlib ... although to me it will make more sense a `find(where:)`

LucianoPAlmeida (approved):

LucianoPAlmeida:
@guykogus What is left on this one? I think we could revert the test `last` changes if we decide to move on deprecating it... but besides that there anything else?

guykogus:
@LucianoPAlmeida I haven't forgotten about this, I've just been crazy busy and distracted with my day job. I'll finish this soon, promise!

LucianoPAlmeida:
> @LucianoPAlmeida I haven't forgotten about this, I've just been crazy busy and distracted with my day job. I'll finish this soon, promise!

No problem :) 
I was just checking all the open PRs, so on worries!

LucianoPAlmeida (approved):
Awesome, Thanks @guykogus!

Issue #916, opened 2020-11-03
shorthand_operator is disabled


sk-:
> Please fill out this template when filing an issue.
>
> All comment lines beginning with an **>** instruct you with what info we expect. You can delete those lines once you've filled in the info.
>

[x] I've read, understood, and done my best to follow the [Contributing guidelines](https://github.com/SwifterSwift/SwifterSwift/tree/master/CONTRIBUTING.md) before opening this issue.

### What did you do?
I'm evaluating replacing some of our extensions with SwifterSwift, which overall looks really nice and complete, however I found that `shorthand_operator` is disabled.

The rationale for that check is to prevent creating temporary objects when using the inline operators.

So instead of implementing `+=` based on `+`, one should prefer to implement `+` based on `+=`, like:

```
public func + (left: A, right: A) -> A {
  var result = left
  result += right
  return result
}
```

Is there a technical rationale for the current implementation, if not, would you open to accept a PR changing the implementation? There are only 3 files affected, see https://github.com/SwifterSwift/SwifterSwift/search?q=shorthand_operator.

### What did you expect to happen?
> Please replace this with what you expected to happen.

### What happened instead?
> Please replace this with of what happened instead.

### SwifterSwift Environment

- SwifterSwift version:
- Xcode version:
- macOS version running Xcode:
- Swift version:
- Platform(s) running SwifterSwift:

### Demo Project
>Please link to or upload a project we can download that reproduces the issue.

LucianoPAlmeida:
Hi @sk- , can you elaborate on why this is better? 

> The rationale for that check is to prevent creating temporary objects when using the inline operators.

What is exactly the "temporary object" in this situation? Can you elaborate on why this is specific for inline?

sk-:
@LucianoPAlmeida in the current inline operators implementations the following is happening:

```swift
static func += (lhs: inout CGPoint, rhs: CGPoint) {
    // swiftlint:disable:next shorthand_operator
    lhs = lhs + rhs
}
```

1) Allocates and initializes a new instance (inside `+`)
2) Then performs some mutation to prepare that new instance (inside `+`)
3) Sets `lhs` to point to the new instance
4) Releases the old `lhs`, possibly deallocating it

See https://github.com/realm/SwiftLint/issues/1828#issuecomment-328230004 for more details.

Whereas, if you implement it the other way around, you only have step 2 for inline operations.

LucianoPAlmeida:
[...]

PR #919, opened 2020-11-05
refactor: base operators on in-place ones

sk-:
Operators should be implemented based on their in-place variants, not the other way around. In that way in-place operations won't allocate any new objects.

See #916 for more details.

This PR allows us to reenable the swiftlint rule shorthand_operator. It also refactors other operators to use their in-place variant to reduce code duplication.

## Checklist

### DIFF
--- a/Sources/SwifterSwift/CoreGraphics/CGPointExtensions.swift
+++ b/Sources/SwifterSwift/CoreGraphics/CGPointExtensions.swift
@@ -51,7 +51,9 @@ public extension CGPoint {
     ///   - rhs: CGPoint to add.
     /// - Returns: result of addition of the two given CGPoints.
     static func + (lhs: CGPoint, rhs: CGPoint) -> CGPoint {
-        return CGPoint(x: lhs.x + rhs.x, y: lhs.y + rhs.y)
+        var result = lhs
+        result += rhs
+        return result
     }
 
     /// SwifterSwift: Add a CGPoints to self.
@@ -65,8 +67,8 @@ public extension CGPoint {
     ///   - lhs: self
     ///   - rhs: CGPoint to add.
     static func += (lhs: inout CGPoint, rhs: CGPoint) {
-        // swiftlint:disable:next shorthand_operator
-        lhs = lhs + rhs
+        lhs.x += rhs.x
+        lhs.y += rhs.y
     }
 
     /// SwifterSwift: Subtract two CGPoints.
@@ -81,7 +83,9 @@ public extension CGPoint {
     ///   - rhs: CGPoint to subtract.
     /// - Returns: result of subtract of the two given CGPoints.
     static func - (lhs: CGPoint, rhs: CGPoint) -> CGPoint {
-        return CGPoint(x: lhs.x - rhs.x, y: lhs.y - rhs.y)
+        var result = lhs
+        result -= rhs
+        return result
     }
 
     /// SwifterSwift: Subtract a CGPoints from self.
@@ -95,8 +99,8 @@ public extension CGPoint {
     ///   - lhs: self
     ///   - rhs: CGPoint to subtract.
     static func -= (lhs: inout CGPoint, rhs: CGPoint) {
-        // swiftlint:disable:next shorthand_operator
-        lhs = lhs - rhs
+        lhs.x -= rhs.x
+        lhs.y -= rhs.y
     }
 
     /// SwifterSwift: Multiply a CGPoint with a scalar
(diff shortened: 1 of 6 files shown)


sk-:
I changed all the other operators (except for the ones in `NSAttributedString`) to be based on the in-place operators. Let me know what you think.

Sources/SwifterSwift/CoreGraphics/CGPointExtensions.swift
```diff
@@ -81,7 +83,9 @@ public extension CGPoint {
     ///   - rhs: CGPoint to subtract.
     /// - Returns: result of subtract of the two given CGPoints.
     static func - (lhs: CGPoint, rhs: CGPoint) -> CGPoint {
-        return CGPoint(x: lhs.x - rhs.x, y: lhs.y - rhs.y)
```
LucianoPAlmeida, line 84: TBH I think this is better the way it was ... to me it is simpler :)

Sources/SwifterSwift/CoreGraphics/CGPointExtensions.swift
```diff
@@ -51,7 +51,9 @@ public extension CGPoint {
     ///   - rhs: CGPoint to add.
     /// - Returns: result of addition of the two given CGPoints.
     static func + (lhs: CGPoint, rhs: CGPoint) -> CGPoint {
-        return CGPoint(x: lhs.x + rhs.x, y: lhs.y + rhs.y)
```
LucianoPAlmeida, line 54: This one too ...

Sources/SwifterSwift/CoreGraphics/CGPointExtensions.swift
```diff
@@ -65,8 +67,8 @@ public extension CGPoint {
     ///   - lhs: self
     ///   - rhs: CGPoint to add.
     static func += (lhs: inout CGPoint, rhs: CGPoint) {
-        // swiftlint:disable:next shorthand_operator
-        lhs = lhs + rhs
```
LucianoPAlmeida, line 69: Nice!

Sources/SwifterSwift/CoreGraphics/CGPointExtensions.swift
```diff
@@ -110,7 +114,9 @@ public extension CGPoint {
     ///   - scalar: scalar value.
     /// - Returns: result of multiplication of the given CGPoint with the scalar.
     static func * (point: CGPoint, scalar: CGFloat) -> CGPoint {
-        return CGPoint(x: point.x * scalar, y: point.y * scalar)
+        var result = point
```
LucianoPAlmeida, line 117: Here too ...

LucianoPAlmeida (commented):
Let's just change the places that use `-`, `+`, `*`, `/`... inside an `...=`, which I believe are the ones which the rule applies. The other let's leave as it was before :)

LucianoPAlmeida:
Also, don't forget the changelog entry :)

guykogus (changes requested):
Unfortunately code duplication is inevitable in this project. One of the conditions is that any function should be copyable as-is, without having to copy any other functions*. This means that you can't use the in-place variants.

In some places it's also a lot more inefficient. E.g. in `CGPoint.+(lhs:rhs:)` instead of creating the new object directly with the added values, you're creating a copy of an object, jumping to another function in order to add the values via a reference and then jump back to return the updated object. It's longer, less efficient code that violates a primary directive of the project.

On the other hand, the updates to the in-place functions, e.g. `CGPoint.+=(lhs:rhs:)`, are much better! I agree that they're more efficient and they also fix the copying issue mentioned.

---

\* In the past this wasn't enforced so it exists in some places, but we don't want it to happen any more**.

** Even though I'm personally not in favour of it.

LucianoPAlmeida:
> One of the conditions is that any function should be copyable as-is, without having to copy any other functions*. This means that you can't use the in-place variants.

IMO, non-complex arithmetic operators such `+` and `+=` can be an exception to this rule because one is always a mutating complement of the other, and people likely copying both anyways. I think that is the reason we allowed this... 

> In some places it's also a lot more inefficient. E.g. in CGPoint.+(lhs:rhs:) 

+ 1

guykogus:
> IMO, non-complex arithmetic operators such `+` and `+=` can be an exception to this rule because one is always a mutating complement of the other, and people likely copying both anyways. I think that is the reason we allowed this...

I'm more than happy to make an exception here. My other points still stand though :)

sk-:
Closing as I don't have the time to address the requirements. Feel free to reuse portions of this PR.

LucianoPAlmeida:
> Closing as I don't have the time to address the requirements. Feel free to reuse portions of this PR.

Hey @sk- , no worries!
But feel free to get back to it when possible in the future :)

PR #927, opened 2020-12-19
Modify Namespaces in ColorExtensions

ShivaHuang:
Use `enum` to declare a namespace, instead of `struct`.

The difference of using enum and struct is,  if struct is used, we need to make `init` private to prevent user create an instance of this namespace.

Like the following shows: https://github.com/SwifterSwift/SwifterSwift/blob/61187db25876a8bb3cb49548478d072bdefb52a5/Sources/SwifterSwift/Shared/ColorExtensions.swift#L365-L368
https://github.com/SwifterSwift/SwifterSwift/blob/61187db25876a8bb3cb49548478d072bdefb52a5/Sources/SwifterSwift/Shared/ColorExtensions.swift#L458-L461
https://github.com/SwifterSwift/SwifterSwift/blob/61187db25876a8bb3cb49548478d072bdefb52a5/Sources/SwifterSwift/Shared/ColorExtensions.swift#L1294-L1297

And it's easy to forget to do the protection like:
https://github.com/SwifterSwift/SwifterSwift/blob/61187db25876a8bb3cb49548478d072bdefb52a5/Sources/SwifterSwift/Shared/ColorExtensions.swift#L1749-L1751

Using `enum` can avoid these issues by natural.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,7 +67,7 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 - **UIAlertController**:
   - Mark `show` method as unavailable for `iOSAppExtension` targets. [#918](https://github.com/SwifterSwift/SwifterSwift/pull/918) by [LucianoPAlmeida](https://github.com/LucianoPAlmeida)
 - **ColorExtension**:
-  - Use `enum` to declare namespace instead of using `struct`. Thus private initailizer is no longer needed. by [Shiva Huang](https://github.com/ShivaHuang)
+  - Use `enum` to declare namespace instead of using `struct`. Thus private initailizer is no longer needed. [#927] by [Shiva Huang](https://github.com/ShivaHuang)
 
 ### Deprecated
 - **Sequence**:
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -66,6 +66,8 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
   - Added `addPaddingRight`,`addPaddingRightIcon`extension,[#878](https://github.com/SwifterSwift/SwifterSwift/pull/878) by [Jayxiang](https://github.com/Jayxiang)
 - **UIAlertController**:
   - Mark `show` method as unavailable for `iOSAppExtension` targets. [#918](https://github.com/SwifterSwift/SwifterSwift/pull/918) by [LucianoPAlmeida](https://github.com/LucianoPAlmeida)
+- **ColorExtension**:
+  - Use `enum` to declare namespace instead of using `struct`. Thus private initailizer is no longer needed. by [Shiva Huang](https://github.com/ShivaHuang)
```
thisIsTheFoxe, line 70:
```suggestion
  - Use `enum` to declare namespace instead of using `struct`. Thus private initailizer is no longer needed. [#927] by [Shiva Huang](https://github.com/ShivaHuang)
```

thisIsTheFoxe (commented):
Not sure if the *fixed* entry is needed, since it's not a know bug.

Other than that lgtm..

LucianoPAlmeida (approved):
Although I think is a bit weird to use enum that way ... I think we can take this

Issue #928, opened 2020-12-19
Add Enum Angle to UIViewExtension


ShivaHuang:
- [x] I've read, understood, and done my best to follow the [Contributing guidelines](https://github.com/SwifterSwift/SwifterSwift/tree/master/CONTRIBUTING.md) before opening this issue.

### What did you do?

I tried to use function `rotate(byAngle:, ofType:, animated:, duration:, completion:)` from `UIViewExtension` but the signature is not that swifty.

### What did you expect to happen?

We had defined an enum `AngleUnit` for angle unit, why not put an associated value for the angle?

We can define an enum like this:

```swift
enum Angle {
    /// SwifterSwift: degrees.
    case degree(CGFloat)

    /// SwifterSwift: radians.
    case radians(CGFloat)
}
```

So we can invoke the rotate function like this:

```swift
view.rotate(byAngle: .degree(90.0))
```

The same technique is also used by Apple, please reference [DispatchTimeInterval](https://developer.apple.com/documentation/dispatch/dispatchtimeinterval).

Further more, enum `Angle` can be improved by adding some computed properties like:

```swift
extension Angle {
    // Get the angle value in degree.
    var degreeValue: CGFloat { ... }

    // Get the angle value in radians.
    var radiansValue: CGFloat { ... }
```

and maybe conform to protocols like `equatable`, `comparable`, `AdditiveArithmetic`.

For the `rotate` functions, I think `rotate(by angle: Angle, ...)` is more Swift API naming convention upheld.
The same applies to `rotate(to angle: Angle, ...)`.

### What happened instead?
Currently we need to invoke this function this way:

```swift
view.rotate(byAngle: 90, ofType: .degree)
```

### SwifterSwift Environment

- SwifterSwift version: 5.2
- Xcode version: 12.3
- macOS version running Xcode: macOS Big Sur 11.1
- Swift version: 5.3.2
- Platform(s) running SwifterSwift: iOS

### Demo Project

The sample implementation has already listed above. I'll happy to open a PR for this change.

LucianoPAlmeida:
Ah, I see the point, but unfortunately, we avoid creating new types inside the library... but a solution I can see to this problem is having an API that takes a `Measurement` like `func rotate(angle: Measurement<UnitAngle>)` and then we could extend `Measurement` class like 
```swift 
extension Measurement where UnitType == UnitAngle {
   static func degrees(_ value: Double) { Measurement(value: value, unit: .degrees) }
[...]

PR #940, opened 2021-01-30
UIView Extension Rotate

ShivaHuang:
Currently there are two functions to rotate a view.
* `rotate(byAngle:ofType:animated:duration:completion:)`
* `rotate(toAngle:ofType:animated:duration:completion:)`

However, there are several problems here.

1. Though these two functions have different names and different implementations, the result is identical. You can find the test cases of them are exactly the same.
2. They introduced a new type `UIView.AngleUnit` to library.
3. The naming is not fully following the naming convention of Swift.

To solve these problems, we can use `Measurement<UnitAngle>` to presenting the angle to rotate and change the function signature to `rotate(by:animated:duration:completion:)`, just like existed extension `UIImage.rotated(by:)`.

Ref: https://github.com/SwifterSwift/SwifterSwift/issues/928

## Checklist

### DIFF
--- a/Sources/SwifterSwift/UIKit/UIViewExtensions.swift
+++ b/Sources/SwifterSwift/UIKit/UIViewExtensions.swift
@@ -18,18 +18,6 @@ public extension UIView {
         case vertical
     }
 
-    /// SwifterSwift: Angle units.
-    ///
-    /// - degrees: degrees.
-    /// - radians: radians.
-    enum AngleUnit {
-        /// SwifterSwift: degrees.
-        case degrees
-
-        /// SwifterSwift: radians.
-        case radians
-    }
-
     /// SwifterSwift: Shake animations types.
     ///
     /// - linear: linear animation.
@@ -387,48 +375,6 @@ public extension UIView {
         let radians = CGFloat(angle.converted(to: .radians).value)
         UIView.animate(withDuration: aDuration, delay: 0, options: .curveLinear, animations: { () -> Void in
             self.transform = self.transform.rotated(by: radians)
-        }, completion: completion)
-    }
-
-    /// SwifterSwift: Rotate view by angle on relative axis.
-    ///
-    /// - Parameters:
-    ///   - angle: angle to rotate view by.
-    ///   - type: type of the rotation angle.
-    ///   - animated: set true to animate rotation (default is true).
-    ///   - duration: animation duration in seconds (default is 1 second).
-    ///   - completion: optional completion handler to run with animation finishes (default is nil).
-    func rotate(
-        byAngle angle: CGFloat,
-        ofType type: AngleUnit,
-        animated: Bool = false,
-        duration: TimeInterval = 1,
-        completion: ((Bool) -> Void)? = nil) {
-        let angleWithType = (type == .degrees) ? .pi * angle / 180.0 : angle
-        let aDuration = animated ? duration : 0
-        UIView.animate(withDuration: aDuration, delay: 0, options: .curveLinear, animations: { () -> Void in
-            self.transform = self.transform.rotated(by: angleWithType)
-        }, completion: completion)
-    }
-
-    /// SwifterSwift: Rotate view to angle on fixed axis.
-    ///
-    /// - Parameters:
-    ///   - angle: angle to rotate view to.
-    ///   - type: type of the rotation angle.
-    ///   - animated: set true to animate rotation (default is false).
-    ///   - duration: animation duration in seconds (default is 1 second).
-    ///   - completion: optional completion handler to run with animation finishes (default is nil).
-    func rotate(
-        toAngle angle: CGFloat,
-        ofType type: AngleUnit,
-        animated: Bool = false,
-        duration: TimeInterval = 1,
-        completion: ((Bool) -> Void)? = nil) {
-        let angleWithType = (type == .degrees) ? .pi * angle / 180.0 : angle
-        let aDuration = animated ? duration : 0
-        UIView.animate(withDuration: aDuration, animations: {
-            self.transform = self.transform.concatenating(CGAffineTransform(rotationAngle: angleWithType))
         }, completion: completion)
     }
 
--- a/Sources/SwifterSwift/UIKit/UIViewExtensions.swift
+++ b/Sources/SwifterSwift/UIKit/UIViewExtensions.swift
@@ -362,9 +362,9 @@ public extension UIView {
     ///
     /// - Parameters:
     ///   - angle: Angle to rotate view by.
-    ///   - animated: Set `true` to animate rotation (default is `true`).
-    ///   - duration: Animation duration in seconds (default is 1 second).
-    ///   - completion: Optional completion handler to run with animation finishes (default is nil).
+    ///   - animated: Set `true` to animate rotation (default is `false`).
+    ///   - duration: Animation duration in seconds (default is `1` second).
+    ///   - completion: Optional completion handler to run with animation finishes (default is `nil`).
     @available(watchOS 3.0, tvOS 10.0, *)
     func rotate(
         by angle: Measurement<UnitAngle>,
--- a/Sources/SwifterSwift/UIKit/UIViewExtensions.swift
+++ b/Sources/SwifterSwift/UIKit/UIViewExtensions.swift
@@ -373,6 +373,26 @@ public extension UIView {
     /// SwifterSwift: Rotate view by angle on relative axis.
     ///
     /// - Parameters:
+    ///   - angle: Angle to rotate view by.
+    ///   - animated: Set `true` to animate rotation (default is `true`).
+    ///   - duration: Animation duration in seconds (default is 1 second).
+    ///   - completion: Optional completion handler to run with animation finishes (default is nil).
+    @available(watchOS 3.0, tvOS 10.0, *)
+    func rotate(
+        by angle: Measurement<UnitAngle>,
+        animated: Bool = false,
+        duration: TimeInterval = 1,
+        completion: ((Bool) -> Void)? = nil) {
+        let aDuration = animated ? duration : 0
+        let radians = CGFloat(angle.converted(to: .radians).value)
+        UIView.animate(withDuration: aDuration, delay: 0, options: .curveLinear, animations: { () -> Void in
+            self.transform = self.transform.rotated(by: radians)
+        }, completion: completion)
+    }
+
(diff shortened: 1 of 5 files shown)


Sources/SwifterSwift/UIKit/UIViewExtensions.swift
```diff
@@ -373,42 +361,20 @@ public extension UIView {
     /// SwifterSwift: Rotate view by angle on relative axis.
     ///
     /// - Parameters:
-    ///   - angle: angle to rotate view by.
-    ///   - type: type of the rotation angle.
-    ///   - animated: set true to animate rotation (default is true).
-    ///   - duration: animation duration in seconds (default is 1 second).
-    ///   - completion: optional completion handler to run with animation finishes (default is nil).
+    ///   - angle: Angle to rotate view by.
+    ///   - animated: Set `true` to animate rotation (default is `true`).
```
guykogus, line 365: Is this the default? 😂

Sources/SwifterSwift/UIKit/UIViewExtensions.swift
```diff
@@ -373,42 +361,20 @@ public extension UIView {
...
-        byAngle angle: CGFloat,
-        ofType type: AngleUnit,
+        by angle: Measurement<UnitAngle>,
         animated: Bool = false,
         duration: TimeInterval = 1,
         completion: ((Bool) -> Void)? = nil) {
-        let angleWithType = (type == .degrees) ? .pi * angle / 180.0 : angle
         let aDuration = animated ? duration : 0
+        let radians = CGFloat(angle.converted(to: .radians).value)
         UIView.animate(withDuration: aDuration, delay: 0, options: .curveLinear, animations: { () -> Void in
```
guykogus, line 376: Personally I think it's really bad that this is wrapped inside an animation block. It's so unnecessary. Let whoever wants to rotate the view wrap it themselves.

Tests/UIKitTests/UIViewExtensionsTests.swift
```diff
@@ -252,46 +252,23 @@ final class UIViewExtensionsTests: XCTestCase {
         waitForExpectations(timeout: 0.5)
     }
 
-    func testRotateByAngle() {
+    func testRotateBy() {
         let view1 = UIView()
         let transform1 = CGAffineTransform(rotationAngle: 2)
-        view1.rotate(byAngle: 2, ofType: .radians, animated: false, duration: 0, completion: nil)
+        view1.rotate(by: .init(value: 2, unit: .radians), animated: false, duration: 0, completion: nil)
```
guykogus, line 258: I think we never use `.init` in the code. It doesn't look nice.
Also, since this is a test, and not the public library itself, you can use any SwifterSwift extension, like the ones you recently added 😉 
```suggestion
        view1.rotate(by: .radians(2), animated: false, duration: 0, completion: nil)
```

Tests/UIKitTests/UIViewExtensionsTests.swift
```diff
@@ -252,46 +252,23 @@ final class UIViewExtensionsTests: XCTestCase {
...
         XCTAssertEqual(view1.transform, transform1)
 
         let view2 = UIView()
         let transform2 = CGAffineTransform(rotationAngle: .pi * 90.0 / 180.0)
-        view2.rotate(byAngle: 90, ofType: .degrees, animated: false, duration: 0, completion: nil)
+        view2.rotate(by: .init(value: 90, unit: .degrees), animated: false, duration: 0, completion: nil)
```
guykogus, line 263: Same here.
```suggestion
        view2.rotate(by: .degrees(90), animated: false, duration: 0, completion: nil)
```

Tests/UIKitTests/UIViewExtensionsTests.swift
```diff
@@ -252,46 +252,23 @@ final class UIViewExtensionsTests: XCTestCase {
...
-        view2.rotate(toAngle: 90, ofType: .degrees, animated: false, duration: 0, completion: nil)
-        XCTAssertEqual(view2.transform, transform2)
-
-        let rotateExpectation = expectation(description: "view rotated")
-
-        let view3 = UIView()
-        let transform3 = CGAffineTransform(rotationAngle: 2)
-
-        view3.rotate(toAngle: 2, ofType: .radians, animated: true, duration: 0.5) { _ in
+        view3.rotate(by: .init(value: 2, unit: .radians), animated: true, duration: 0.5) { _ in
```
guykogus, line 271:
```suggestion
        view3.rotate(by: .radians(2), animated: true, duration: 0.5) { _ in
```

SwifterSwift.xcodeproj/project.pbxproj
```diff
@@ -1309,6 +1314,14 @@
 			path = HealthKitTests;
 			sourceTree = "<group>";
 		};
+		587AC78525C1D4C7009977B5 /* Deprecated */ = {
+			isa = PBXGroup;
+			children = (
+				587AC78625C1D4F0009977B5 /* UIKitDeprecated.swift */,
```
guykogus, line 1320: We put a `Deprecated` folder inside the parent folder, in this case `Sources/SwifterSwift/UIKit`. You can see that one already exists in `SwiftStdlib`. This is required because if somebody decides to use only a subspec then the deprecated folder will need to be included there.

guykogus (changes requested):

Sources/SwifterSwift/UIKit/UIViewExtensions.swift
LucianoPAlmeida: +1

Tests/UIKitTests/UIViewExtensionsTests.swift
ShivaHuang: Cool, that's the final format I want for these PRs.

SwifterSwift.xcodeproj/project.pbxproj
ShivaHuang: Yeah, this `Deprecated` folder located at `Sources/SwifterSwift/UIKit`, and I moved the deprecated functions and enum in a file named `UIKitDeprecated.swift` in this folder.

Did I miss something?

Sources/SwifterSwift/UIKit/UIViewExtensions.swift
ShivaHuang: Ok, you got me coping and pasting the comment from the deprecated function 😂

It looks like this typo has been a long time. I'll fix them. 😁

Sources/SwifterSwift/UIKit/UIViewExtensions.swift
ShivaHuang: Hmmm...I didn't think about it before.

My initial intention was, the way using a enum to tell the angle value is in degrees or radians doesn't look the swift way.

I got hint from @LucianoPAlmeida that `Measurement<UnitAngle>` can be a good solution, so I just modified the parameter type and name from the old function and didn't think much about the function itself.

After you pointed it out, I reviewed the functions in `UIViewExtensions`, found that there are four functions in this kind. Two for rotating, which are merged into one in this PR, one for scaling, and one for shaking.

* `rotate(byAngle:ofType:animated:duration:completion:)`
* `rotate(toAngle:ofType:animated:duration:completion:)`
* `scale(by:animated:duration:completion:)`
* `shake(direction:duration:animationType:completion:)`

I believe the original author for these functions is trying to add extensions for simple view animations, that explains why these function signatures look like the one from `UIKit`, for example `UIScrollView.zoom(to:animated:)` or `UISwitch.setOn(_:animated:)`.

I think these functions should be re-considered the functionality they provide.

Do you think we should modify them in this PR, or we can create an issue and make some discussion there?

guykogus: I'm happy to leave this PR for just removing the old `AngleUnit` enum. However, notice that you've removed 2 functions and only added 1.

ShivaHuang: Sorry I didn't get your point here.

The removed functions are identical in their functionality, so I think we should only keep one of them.
In the meantime, `Measurement` is used to replace the parameter in the remaining function and the argument label is also modified.

Or do you suggest, we should only remove `AngleUnit`, keeping the 2 functions in this PR, and then in another PR, deprecating the duplicated function?

I think there are two more modifications can be done to these functions. I'm also considering if they should be split into multiple PRs or put all of them in one single PR.

guykogus: Originally there was a `rotate(byAngle:` function and `rotate(toAngle:`. You've replaced the first but just deleted the second.

SwifterSwift.xcodeproj/project.pbxproj
guykogus: Ah, you did it correctly. I was just reading from the PR and it looked like you'd placed the folder at the top level. I can see that it's correct :D

Tests/UIKitTests/UIViewExtensionsTests.swift
guykogus: That's great, but you still haven't used it here 😛

Sources/SwifterSwift/UIKit/UIViewExtensions.swift
ShivaHuang: Yeah, because they are doing the same thing.

Maybe we should re-organize these modifications so the history would be more clear?
Something like:

1. Deprecate `rotate(toAngle:` because duplicated functionality to `rotate(byAngle:`
2. Add new function `rotate(by:)`, which takes a `Measurement` and not wrapped in an animation block.
3. Deprecate `rotate(byAngle:`.
4. Deprecate `AngleUnit` since no function use it.

What do you think?

Tests/UIKitTests/UIViewExtensionsTests.swift
ShivaHuang: Oh, my bad🙈

Sources/SwifterSwift/UIKit/UIViewExtensions.swift
guykogus: Oh damn, you're right, the original `rotate(toAngle:)` was broken. I don't think it intentionally was meant to do that.
Do you know how to fix this? It's not straightforward, so if not we can just deprecate it.

PR #943, opened 2021-02-04
Increase the filter operation of the Array by type

macoooo:
## Checklist

### DIFF
--- a/Sources/SwifterSwift/SwiftStdlib/ArrayExtensions.swift
+++ b/Sources/SwifterSwift/SwiftStdlib/ArrayExtensions.swift
@@ -131,3 +131,50 @@ public extension Array where Element: Equatable {
         return filter { set.insert($0[keyPath: path]).inserted }
     }
 }
+
+public extension Array {
+    /// The inheritance system is UIButton -> UIControl -> UIView
+    /// Both U type and U subclasses will be matched
+    ///
+    ///     [UIButton(), UIControl(),UIView()]
+    ///         .filter(kindOf: UIControl.self) ->
+    ///     [<UIButton: 0x7fa710cf54c0>, <UIControl: 0x7fa671429950>]
+    /// - Parameter : A data type
+    /// - Returns: A set of the elements that isIncluded allows.
+    func filter<U>(kindOf _: U.Type) -> [U] {
+        var result = [U]()
+        for case let e as U in self {
+            result.append(e)
+        }
+        return result
+    }
+
+    /// Only match U type, subclass will not match
+    ///
+    ///     [UIButton(), UIControl(), UIView()]
+    ///         .filter(memberOf: UIControl.self)) ->
+    ///     [<UIControl: 0x7fba09816c90>]
+    /// - Parameter t: A data type
+    /// - Returns: A set of the elements that isIncluded allows.
+    func filter<U>(memberOf t: U.Type) -> [U] {
+        var result = [U]()
+        for e in self where String(describing: type(of: e)) == String(describing: t) {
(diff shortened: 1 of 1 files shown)


Sources/SwifterSwift/SwiftStdlib/ArrayExtensions.swift
```diff
@@ -131,3 +131,50 @@ public extension Array where Element: Equatable {
...
+public extension Array {
+    /// The inheritance system is UIButton -> UIControl -> UIView
+    /// Both U type and U subclasses will be matched
+    ///
+    ///     [UIButton(), UIControl(),UIView()]
+    ///         .filter(kindOf: UIControl.self) ->
+    ///     [<UIButton: 0x7fa710cf54c0>, <UIControl: 0x7fa671429950>]
+    /// - Parameter : A data type
+    /// - Returns: A set of the elements that isIncluded allows.
+    func filter<U>(kindOf _: U.Type) -> [U] {
```
LucianoPAlmeida, line 144: Can you elaborate on why this wrapper would be useful? Because to me this would be just abstract some operation that is already trivial e.g.
```
[UIButton(), UIControl(), UIView()].compactMap { $0 as? UIControl }  //   [UIControl]
```
So this may look a bit over abstract something that is already simple.

Sources/SwifterSwift/SwiftStdlib/ArrayExtensions.swift
```diff
@@ -131,3 +131,50 @@ public extension Array where Element: Equatable {
...
+    /// Only match U type, subclass will not match
+    ///
+    ///     [UIButton(), UIControl(), UIView()]
+    ///         .filter(memberOf: UIControl.self)) ->
+    ///     [<UIControl: 0x7fba09816c90>]
+    /// - Parameter t: A data type
+    /// - Returns: A set of the elements that isIncluded allows.
+    func filter<U>(memberOf t: U.Type) -> [U] {
+        var result = [U]()
+        for e in self where String(describing: type(of: e)) == String(describing: t) {
```
LucianoPAlmeida, line 161: You can simplify this comparing meta types directly 
```suggestion
        for e in self where type(of: e) ==  t  {
```

LucianoPAlmeida: But again, can you provide an useful use case for that because that is also could be trivially achievable with
```swift
[UIButton(), UIControl(), UIView()].filter { type(of: $0) == UIControl.self }
```

Sources/SwifterSwift/SwiftStdlib/ArrayExtensions.swift
```diff
@@ -131,3 +131,50 @@ public extension Array where Element: Equatable {
...
+        return result
+    }
+
+    /// Inferred from the return value type, it has the same function as filter<U>(kindOf type: U.Type)
+    ///
+    ///      let array: [UIView] = [UIButton(), UIControl(), UIView()]
+    ///      let res: [UIControl] = array.filterRetType()
+    ///      res  ->   [<UIButton: 0x7fc0a3c27820>, <UIControl: 0x7fc0a3c27ee0>]
+    /// - Returns: A set of the elements that isIncluded allows.
+    func filterRetType<U>() -> [U] {
```
LucianoPAlmeida, line 173: Not really sure about this, the only difference from `filter<U>(kindOf type: U.Type)` is the source of the inference ...

LucianoPAlmeida (commented):
Thank you for your contribution! 
I left some comments inline

PR #972, opened 2021-07-06
Rename conflicting UIView extensions

### LINKED ISSUE #909: Ambiguous use of 'cornerRadius'
1. Found this candidate (AuthenticationServices.ASAuthorizationAppleIDButton)
2. Found this candidate (SwifterSwift.UIView)

Jayxiang:
Fixes #909 
## Checklist


LucianoPAlmeida (commented):
Code changes looks good, only missing changelog entry. Can you please add it under breaking changes? =]

LucianoPAlmeida (approved):
Thanks!

guykogus:
I'm not really sure that I'm on board with changing `borderColor` and `borderWidth`, since according to the issue they come from other libraries. Do we really want to rename every name that clashes with another open source library? I get it if it clashes with one of Apple's own frameworks.

LucianoPAlmeida:
> I'm not really sure that I'm on board with changing `borderColor` and `borderWidth`, since according to the issue they come from other libraries. Do we really want to rename every name that clashes with another open source library? I get it if it clashes with one of Apple's own frameworks.

The way I see is more a way to improve the expressivity of the names(since all are abstractions to change layer properties, I like having `layer` prefix to make that clear) than fix collision issues since, the new name may colide as well... So at the I'm good with this change because of that, fixing collision is just a bonus =]

guykogus (approved):

PR #980, opened 2021-08-18
Added SKSpriteNode extension

erikdrobne:
🚀 
Added `aspectFill` method to dynamically size sprite node with respect to aspect ratio.

## Checklist

### DIFF
--- a/Sources/SwifterSwift/SpriteKit/SKSpriteNodeExtensions.swift
+++ b/Sources/SwifterSwift/SpriteKit/SKSpriteNodeExtensions.swift
@@ -1,0 +1,25 @@
+// SKSpriteNodeExtensions.swift - Copyright 2021 SwifterSwift
+
+#if canImport(SpriteKit)
+import SpriteKit
+
+// MARK: - Methods
+
+public extension SKSpriteNode {
+    /// SwifterSwift: SKSpriteNode sized with respect to aspect ratio.
+    ///
+    ///        node.aspectFill(to: CGSize(width: 300, height: 300)
+    ///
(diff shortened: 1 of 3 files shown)


Sources/SwifterSwift/SpriteKit/SKSpriteNodeExtensions.swift
```diff
@@ -0,0 +1,25 @@
...
+
+public extension SKSpriteNode {
+    /// SwifterSwift: SKSpriteNode sized with respect to aspect ratio.
+    ///
+    ///        node.aspectFill(to: CGSize(width: 300, height: 300)
+    ///
+    /// - Parameter fillSize: fill size to use for aspect ratio calculation.
+    func aspectFill(to fillSize: CGSize) {
+        if let texture = self.texture {
+            let horizontalRatio = fillSize.width / texture.size().width
```
LucianoPAlmeida, line 16: Let's be safe and check for `texture.size().width` being `0` to avoid division by zero crash, I think this would be unlikely but still nice to be safe =]

LucianoPAlmeida (commented):
Thanks for the contribution @erikdrobne!
Seems a reasonable, can you also add a changelog entry?

Sources/SwifterSwift/SpriteKit/SKSpriteNodeExtensions.swift
erikdrobne: I agree, will add. :)

LucianoPAlmeida (approved):
LGTM!

erikdrobne:
@SD10 Thanks for the invitation! 😊

PR #982, opened 2021-08-31
Adding daysRemaining(in:) to DateExtensions.

pbeyer08:
Addresses #552.

This PR adds a `Date` extension which returns the number of days remaining in various `Calendar.Components` such as `.month` or `.year`.

## Checklist

### DIFF
--- a/Sources/SwifterSwift/Foundation/DateExtensions.swift
+++ b/Sources/SwifterSwift/Foundation/DateExtensions.swift
@@ -900,6 +900,24 @@ public extension Date {
         let componentValue = components.value(for: component)!
         return abs(componentValue) <= value
     }
+		
+	/// SwifterSwift: calculate number of full days remaining in various supported Calendar.Components.
+	///
+	/// - Parameters:
+	///   - component: Calendar.Component to use.
+	/// - Returns: number of days remaining in that Calendar.Component, nil if unsupported.
+	func daysRemaining(in component: Calendar.Component) -> Int? {
+		switch component {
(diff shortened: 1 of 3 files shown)


pbeyer08:
NOTE:  This extension uses `end(of:)` and `daysSince(:)` extensions, which in turn depend on various other `Date` extensions.  I assume the following contribution rule does not apply here:

> Do not use an existing SwifterSwift extension inside another SwifterSwift extension. All extensions should be able to be copied and pasted and work immediately without having to copy another extension.

Sources/SwifterSwift/Foundation/DateExtensions.swift
```diff
@@ -900,6 +900,24 @@ public extension Date {
...
+	/// - Returns: number of days remaining in that Calendar.Component, nil if unsupported.
+	func daysRemaining(in component: Calendar.Component) -> Int? {
+		switch component {
+			case .month, .year, .weekOfMonth, .weekOfYear:
+				if let endOfComponent = self.end(of: component) {
+					let daysRemaining = endOfComponent.daysSince(self)
+					return Int(daysRemaining)
+				}
+			default:
+				()
```
erikdrobne, line 917: You could also return `nil` in the `default` case.

guykogus (changes requested):
Nice. Just a few small suggestions.
N.B. Agree that it makes sense here to use another `end(of:)`, more because it's a pretty complicated function than because that one already uses others.

PR #997, opened 2021-11-22
Optimize URLExtensions

honbolau:
Replace `for` with `reduce` in the `queryParameters`


RomanPodymov (changes requested):
Hello @honbolau 
Please check [CONTRIBUTING.md](https://github.com/SwifterSwift/SwifterSwift/blob/master/CONTRIBUTING.md).

LucianoPAlmeida:
Hi @honbolau, thanks for your PR :) 
But this seems to me just a syntax preference and I see not much difference from the original version, so we are going to keep it as it is.

PR #1002, opened 2022-02-21
Add ability to add actions for `UIControl` with closures.

denandreychuk:
It's nice to have extension to add `UIControl` events inline.

## Checklist

### DIFF
--- a/Sources/SwifterSwift/UIKit/UIControlExtensions.swift
+++ b/Sources/SwifterSwift/UIKit/UIControlExtensions.swift
@@ -13,6 +13,7 @@ public extension UIControl {
     ///   - event: Event to perform action for.
     ///   - action: Action to perform.
     func on(_ event: UIControl.Event, perform action: @escaping () -> Void) {
+        // https://stackoverflow.com/a/41438789/9014720
         @objc final class ClosureSleeve: NSObject {
             private let closure: () -> Void
             init (_ closure: @escaping () -> Void) {
--- a/Sources/SwifterSwift/UIKit/UIControlExtensions.swift
+++ b/Sources/SwifterSwift/UIKit/UIControlExtensions.swift
@@ -1,0 +1,30 @@
+// UIControlExtensions.swift - Copyright 2022 SwifterSwift
+
+#if canImport(UIKit)
+import UIKit
+
+// MARK: - Methods
+
+public extension UIControl {
+    
+    /// SwifterSwift: Perform action on specified event.
+    ///
+    /// - Parameters:
+    ///   - event: Event to perform action for.
+    ///   - action: Action to perform.
+    func on(_ event: UIControl.Event, perform action: @escaping () -> Void) {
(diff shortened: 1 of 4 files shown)


Sources/SwifterSwift/UIKit/UIControlExtensions.swift
```diff
@@ -0,0 +1,31 @@
...
+// MARK: - Methods
+
+public extension UIControl {
+    
+    /// SwifterSwift: Perform action on specified event.
+    ///
+    /// - Parameters:
+    ///   - event: Event to perform action for.
+    ///   - action: Action to perform.
+    func on(_ event: UIControl.Event, perform action: @escaping () -> Void) {
```
denandreychuk, line 15: I chose `on(event:perform)` instead of something like `addAction`, cause I want to add similar extensions to `UIView` `onTap(perform:)`. It's also provide ability to extend it even further to something like `onTouchUpInside`.

Sources/SwifterSwift/UIKit/UIControlExtensions.swift
```diff
@@ -0,0 +1,31 @@
...
+            private let closure: () -> Void
+            init (_ closure: @escaping () -> Void) {
+                self.closure = closure
+            }
+            @objc func invoke () {
+                closure()
+            }
+        }
+        let sleeve = ClosureSleeve(action)
+        addTarget(sleeve, action: #selector(ClosureSleeve.invoke), for: event)
```
LucianoPAlmeida, line 27: The problem with this approach is how you remove this `sleeve` object from the UIControl object? Because as I see this will be there for ever ...

denandreychuk: You're right. It will be there for ever. I have 2 prob solutions:
1. Return selector from sleeve with discardable result to add ability to use removeTarget, ClosureSleeve will still be in memory tho till self do.
2. Remove support for <iOS13, use UIAction, and return action with discardable result.

LucianoPAlmeida: I think the ideal solution would be to return an object that holds both UIControl instance and sleeve and remove target at its destruction. But at the same time it would require callers to know that they should hold this reference in order to use it. So I'm not sure if it worth it having this API.
As far as your suggestions: 1 is actually not that better either because callers would be required to know they should remove the selector from target which is not a intuitive API IMO and also don't fit the library. 2 is really not possible, at least not only for implementing this extension. For sure this project can revise the minimum target deployment in general but that is a separated thing. 
Anyways @denandreychuk thank you for taking the time to work on this, but IMHO I think this extension is not a good fit for this library. 

@guykogus thoughts?

denandreychuk: For 2, I mean to use available check to be able to get access to UIAction API. There is no need to change minimum target for the whole project.

Issue #967, opened 2021-06-27
Future of SwifterSwift


omaralbeik:
# Future of SwifterSwift

First of all I want to thank you all for making SwifterSwift what it is today 🚀 

Here are some insights from Github

- 132 contributors!
- 4000+ monthly repo visitors
- 4500+ monthly clones
- 500+ monthly visitors to swifterswift.com

For many of the 132 contributors (including me), SwifterSwift was the first open source contribution, how cool is that!

I wanted to share my vision for the project going forward and gather your feedback. Lets shape the future of SwifterSwift together!

---

### Create a safe and welcoming environment for others to learn and grow through Open Source.

Unlike other open source projects, contributing to SwifterSwift does not require a deep knowledge of the code base, nor the history behind it, all you need is a Swift extension that you want to share with the community!

After your first pull request is merged, you will automatically receive an invitation to join the organization on Github, so you can show that nice "SwifterSwift" badge in your Github profile, but more importantly you're now one of us, you can review pull requests and drive the future of the project!

We want to make it even easier for everyone to contribute to the project by automating some of the confusing parts for new joiners, like adding a change log entry. 

SwifterSwift was how I got to love open source, I personally learned tons and made many friends!

### Experiment with Github Actions

Utilize Github Actions to automate monthly releases without having one of the team to trigger the process manually.

A few years ago -frankly when Swift 3 was released- we thought it is a good idea to sync SwifterSwift's version with Swift's versions

> - SwifterSwift v3 supports Swift 3
> - SwifterSwift v4 supports Swift 4
> - SwifterSwift v5 supports Swift 5
> ...

Now that Swift is reaching a level where the language is mature enough and its features are evolving slowly, we need a versioning system that gives us more flexibility releasing more versions.

[CalVer](https://calver.org/) is a promising option that will enable us to have fully automated monthly releases.

> SwifterSwift v2021.06.01 the first release of June 2021
> SwifterSwift v2021.06.02 the second release of June 2021
> SwifterSwift v2021.10.01 the first release of October 2021

We can use Github Actions' [scheduled events](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#scheduled-events) to release a new version every month without the need for any of the maintainer time to trigger the process or sign off the release.

### swifterswift.com

Create a website where people can search for extensions and copy-paste them to their projects with ease.

An automated script will make sure the website is always up to date whenever a change is made is the main branch.

To make generating the website easier, some changes will be made to the current structure

[...]

PR #1006, opened 2022-03-28
Updated `init(color: size:)` to add a default value for the size parameter.

Andy0570:
Reference from <https://stackoverflow.com/questions/26542035/create-uiimage-with-solid-color-in-swift>

## Checklist

### DIFF
--- a/Sources/SwifterSwift/UIKit/UIImageExtensions.swift
+++ b/Sources/SwifterSwift/UIKit/UIImageExtensions.swift
@@ -336,18 +336,23 @@ public extension UIImage {
 public extension UIImage {
     /// SwifterSwift: Create UIImage from color and size.
     ///
+    ///     let redImage = UIImage(color: .red)
+    ///     let blueImage = UIImage(color: .blue, size: CGSize(width: 20, height: 20))
+    ///
     /// - Parameters:
     ///   - color: image fill color.
     ///   - size: image size.
-    convenience init(color: UIColor, size: CGSize) {
-        UIGraphicsBeginImageContextWithOptions(size, false, 1)
+    convenience init(color: UIColor, size: CGSize = CGSize(width: 1, height: 1)) {
+        // https://stackoverflow.com/questions/26542035/create-uiimage-with-solid-color-in-swift
+        let rect = CGRect(origin: .zero, size: size)
+        UIGraphicsBeginImageContextWithOptions(rect.size, false, 1)
 
         defer {
             UIGraphicsEndImageContext()
         }
 
         color.setFill()
-        UIRectFill(CGRect(origin: .zero, size: size))
+        UIRectFill(rect)
 
         guard let aCgImage = UIGraphicsGetImageFromCurrentImageContext()?.cgImage else {
             self.init()
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -95,6 +95,9 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
   - Add `beginRefreshing(animated:sendAction:)` that works inside any `UIScrollView` and not only `UITableView`. [#949](https://github.com/SwifterSwift/SwifterSwift/pull/949) by [yonat](https://github.com/yonat)
 - **SKSpriteNode**:
     - Added `aspectFill(to:)` to size SKSpriteNode with respect to aspect ratio. [#490](https://github.com/SwifterSwift/SwifterSwift/pull/490) by [erikdrobne](https://github.com/erikdrobne).
+- **UIImage**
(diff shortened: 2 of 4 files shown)


Andy0570:
English is not my native language; please excuse typing errors.

---

There is one thing that confuses me when optimizing that method.

```Swift
UIGraphicsBeginImageContextWithOptions(_ size: CGSize, _ opaque: Bool, _ scale: CGFloat)
```

The third `scale` parameter of this function is passed a value of `0` in stackoverflow or other publicly available examples, whereas our previous implementation passed a value of `1`.

From Apple's [documentation](https://developer.apple.com/documentation/uikit/1623912-uigraphicsbeginimagecontextwitho).

> The scale factor to apply to the bitmap. If you specify a value of 0.0, the scale factor is set to the scale factor of the device's main screen.

If you set this parameter to `0`, the resulting image size can be automatically scaled to fit the screen. We should also set it to `0`, or leave the choice to the developer, like this.

```swift
init(color: size: scale:)
```

I am concerned that if the generated images are not automatically adapted to the screen size, the images will cause off-screen rendering problems when displayed due to scaling.

Sources/SwifterSwift/UIKit/UIImageExtensions.swift
```diff
@@ -336,18 +336,23 @@ public extension UIImage {
...
     ///
+    ///     let redImage = UIImage(color: .red)
+    ///     let blueImage = UIImage(color: .blue, size: CGSize(width: 20, height: 20))
+    ///
     /// - Parameters:
     ///   - color: image fill color.
     ///   - size: image size.
-    convenience init(color: UIColor, size: CGSize) {
-        UIGraphicsBeginImageContextWithOptions(size, false, 1)
+    convenience init(color: UIColor, size: CGSize = CGSize(width: 1, height: 1)) {
```
LucianoPAlmeida, line 345: Can you provide an use case where is useful to have 1x1 default size? Personally I would prefer not having a default size value.

LucianoPAlmeida:
> The third `scale` parameter of this function is passed a value of `0` in [stackoverflow](https://stackoverflow.com/questions/26542035/create-uiimage-with-solid-color-in-swift) or other publicly available examples, whereas our previous implementation passed a value of `1`.

This is a very old extension in this repo, so I don't know the reason for that. Not sure, but I guess in this case scale wouldn't make any difference because the whole image is a single color.

Sources/SwifterSwift/UIKit/UIImageExtensions.swift
Andy0570: In some scenarios, there is no difference between providing a 10x10 size and a 1x1 size：
```swift
func fetchImageFromUrl(url: URL) -> UIImage {
    // ...

    // If download image failure
    return UIImage(color: .gray, size: CGSize(width: 10, height: 10))
}
```

Or just set the background color for the different states of the button：
```swift
let button = UIButton()
button.setImage(UIImage.init(color: .blue), for: .normal)
button.setImage(UIImage.init(color: .gray), for: .disabled)
```

Also, adding a default parameter to a method doesn't mean you can't actively provide that parameter, it's just one more option. You can still use your preferred method:
```Swift
let redImage = UIImage(color: .red)
let blueImage = UIImage(color: .blue, size: CGSize(width: 20, height: 20))
```

LucianoPAlmeida: > Also, adding a default parameter to a method doesn't mean you can't actively provide that parameter, it's just one more option.

Sure, my question is more in which case a user would want to use a default value that would return a 1x1 image? To me doesn't make much sense a default so specific. Normally APIs provide size values as default it has to be something more general that would apply to most default cases which to me it is not the case for a 1x1 default.

LucianoPAlmeida: From a user perspective imagine calling a `UIImage(color: .red)` and trying to reason why the API provider decided a 1x1 image should be default. Default values should feel naturally expected when defining an API in my opinion. 
That said, I see the example of the button background, but still think a 1x1 default is not the right path.
But is only my opinion, let's wait for others thoughts as well to see what they think =]

Andy0570:
The discussion seems to have come to a standstill and I am considering closing the pull request.

LucianoPAlmeida:
> The discussion seems to have come to a standstill and I am considering closing the pull request.

Sorry for that @Andy0570, the team is busy with other things and we are actually thinking about the future of this project in a more general sense. See https://github.com/SwifterSwift/SwifterSwift/issues/967 
Thanks for the PR, but I guess we can close this for now =]

PR #1026, opened 2022-07-09
hexString computed property for CryptoKit.Digest

MarcoEidinger:
## Checklist

### DIFF
--- a/Sources/SwifterSwift/CryptoKit/DigestExtensions.swift
+++ b/Sources/SwifterSwift/CryptoKit/DigestExtensions.swift
@@ -1,6 +1,6 @@
 // DigestExtensions.swift - Copyright 2022 SwifterSwift
 
-#if canImport(CoreLocation)
+#if canImport(CryptoKit)
 import CryptoKit
 
 @available(iOS 13.0, macOS 10.15, watchOS 6.0, tvOS 13.0, *)
--- a/Sources/SwifterSwift/CryptoKit/DigestExtensions.swift
+++ b/Sources/SwifterSwift/CryptoKit/DigestExtensions.swift
@@ -1,0 +1,17 @@
+// DigestExtensions.swift - Copyright 2022 SwifterSwift
+
+#if canImport(CoreLocation)
+import CryptoKit
+
+@available(iOS 13.0, macOS 10.15, watchOS 6.0, tvOS 13.0, *)
+public extension Digest {
+
+    // MARK: - Properties
+
+    /// SwifterSwift: Hexadecimal value string (read-only)
+    var hexString: String {
+        let bytes: [UInt8] = Array(makeIterator())
+        return Data(bytes).map { String(format: "%02X", $0) }.joined()
+    }
+}
+#endif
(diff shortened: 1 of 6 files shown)


Sources/SwifterSwift/CryptoKit/DigestExtensions.swift
```diff
@@ -0,0 +1,17 @@
+// DigestExtensions.swift - Copyright 2022 SwifterSwift
+
+#if canImport(CoreLocation)
```
LucianoPAlmeida, line 3: Wouldn't that be CriptoKit?
```suggestion
#if canImport(CryptoKit)
```

MarcoEidinger: Thanks @LucianoPAlmeida , good catch!

Sources/SwifterSwift/CryptoKit/DigestExtensions.swift
```diff
@@ -0,0 +1,17 @@
...
+import CryptoKit
+
+@available(iOS 13.0, macOS 10.15, watchOS 6.0, tvOS 13.0, *)
+public extension Digest {
+
+    // MARK: - Properties
+
+    /// SwifterSwift: Hexadecimal value string (read-only)
+    var hexString: String {
+        let bytes: [UInt8] = Array(makeIterator())
```
LucianoPAlmeida, line 13: This implementation does one loop in `self` iterator to construct the array, another to construct data with bytes(which could be dropped since we already have the bytes why construct a data with bytes to iterate over those same underlying bytes?), another loop for map and last one for joined. I would suggest we go with a simpler single loop approach. WDYT?
```swift
extension Digest {
  var hexString: String {
    var result = ""
    for byte in self {
      result += String(format: "%02X", byte)
    }
    return result
 }
}
```

Sources/SwifterSwift/CryptoKit/DigestExtensions.swift
```diff
@@ -0,0 +1,17 @@
...
+
+#if canImport(CryptoKit)
+import CryptoKit
+
+@available(iOS 13.0, macOS 10.15, watchOS 6.0, tvOS 13.0, *)
+public extension Digest {
+
+    // MARK: - Properties
+
+    /// SwifterSwift: Hexadecimal value string (read-only)
```
LucianoPAlmeida, line 11: Can we have a Complexity doc as well, to make clear for people that we have a linear `O(n)` at least in this computed property?

LucianoPAlmeida (commented):
@MarcoEidinger Thanks for the PR! Added some minor comments there =]

MarcoEidinger:
@LucianoPAlmeida thanks for the very valuable feedback. I'll push another commit addressing your feedback items

Sources/SwifterSwift/CryptoKit/DigestExtensions.swift
MarcoEidinger: Changed as suggested with 42e2959a4b2b11b05b26f18ad12b10da4a734faa

Sources/SwifterSwift/CryptoKit/DigestExtensions.swift
MarcoEidinger: Added with 42e2959a4b2b11b05b26f18ad12b10da4a734faa

LucianoPAlmeida (approved):
Thank you!

SD10:

PR #1034, opened 2022-08-09
fix nearestTenMinutes behavior

mmdock:
fix nearestTenMinutes rounding behavior to behavior like nearestHalfHour

Screen Shot 2022-08-09 at 12 01 57 PM

As can be seen in the screenshot, nearestTenMinutes, if the minute is halfway of the target interval period (aka multiple of 5 that ends in 5), rounds down.  nearestHalfHour, rounds up when the minute is halfway between a period range

This PR aligns behavior expectations on similarly named computed variables

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 
 ### Fixed
 - **Date**
-  - fixed incorrect calculation in `nearestTenMinutes` , to align with other `nearest*` date calculations.  [#1034](https://github.com/SwifterSwift/SwifterSwift/pull/1034) by [mmdock](https://github.com/mmdock)
+  - Fixed incorrect calculation in `nearestTenMinutes` to align with other `nearest*` date calculations. [#1034](https://github.com/SwifterSwift/SwifterSwift/pull/1034) by [mmdock](https://github.com/mmdock)
 
 ## [v5.3.0](https://github.com/SwifterSwift/SwifterSwift/releases/tag/5.3.0)
 ### Breaking Change
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 
 ## Upcoming Release
 
+### Fixed
+- **Date**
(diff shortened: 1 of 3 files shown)


LucianoPAlmeida:
CI is failing due to a tool issue, it was a tool we really weren't using so it was removed. Just update with master =]

LucianoPAlmeida:
Regarding the PR itself, I really don't have an opinion about that, so I'll let others take on it. But some thoughts on top of my head:
* Changing this modifies behavior of the method in which someone may rely on, so that will be breaking (not sure how much breaking)
* If we are moving forward with this, can we add some testing cases to exercise that flow?

mmdock:
> Regarding the PR itself, I really don't have an opinion about that, so I'll let others take on it. But some thoughts on top of my head:
> 
> * Changing this modifies behavior of the method in which someone may rely on, so that will be breaking (not sure how much breaking)
> * If we are moving forward with this, can we add some testing cases to exercise that flow?

I can make some updates in a few days with testing cases, etc

mmdock:
> * Changing this modifies behavior of the method in which someone may rely on, so that will be breaking (not sure how much breaking)

tbh, I would regard this as a bug in the initial behavior expectations.  

I would go so far as to say the `nearestQuarterHour` and `nearestFiveMinutes` are also inaccurate based on the name descriptor. 

So in nearestFive minutes, if I have "5:52:50 PM" it would round to "5:50:00 PM".  To me that is inaccurate, since the date is actually closer to "5:55:00 PM"

LucianoPAlmeida (commented):
Can you add a change log entry? I think this looks fine =]

mmdock:
> Can you add a change log entry? I think this looks fine =]

sure! was backpacking, will wrap this up tomorrow :)

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 
 ## Upcoming Release
 
+### Fixed
+- **Date**
+  - fixed incorrect calculation in `nearestTenMinutes` , to align with other `nearest*` date calculations.  [#1034](https://github.com/SwifterSwift/SwifterSwift/pull/1034) by [mmdock](https://github.com/mmdock)
```
guykogus, line 9:
```suggestion
  - Fixed incorrect calculation in `nearestTenMinutes` to align with other `nearest*` date calculations. [#1034](https://github.com/SwifterSwift/SwifterSwift/pull/1034) by [mmdock](https://github.com/mmdock)
```

guykogus (changes requested):
Nitpick

LucianoPAlmeida (approved):

guykogus (approved):

PR #1039, opened 2022-08-25
Add `addGradient(colors:locations:direction:)` extension

Andy0570:
Add `addGradient(colors:locations:direction:)` extension to add gradient color.

## Checklist


Andy0570:
As a reminder, this extension reference from <https://github.com/swiftdevcenter/GradientColorExample>, And already marked in the code.

mustafagunes:
@Andy0570 you need to add the tests of the pull request you opened. You should also describe your improvement to the changelog.md file.

Sources/SwifterSwift/UIKit/UIViewExtensions.swift
```diff
@@ -472,6 +480,45 @@ public extension UIView {
...
+    ///         direction: .topToBottom
+    ///     )
+    ///
+    /// - Parameters:
+    ///   - colors: An array of colors defining the color of each gradient stop.
+    ///   - locations: An array of NSNumber defining the location of each
+    ///                gradient stop as a value in the range [0,1]. The values must be
+    ///                monotonically increasing.
+    ///   - direction: Enumeration type describing the direction of the gradient.
+    func applyGradient(colors: [Any]?, locations: [NSNumber]? = [0.0, 1.0], direction: GradientDirection = .topToBottom) {
```
guykogus, line 497: Just because the old CoreAnimation code doesn't have a great API doesn't mean that we have to do the same.
```suggestion
    func applyGradient(colors: [Color]?, locations: [CGFloat]? = [0.0, 1.0], direction: GradientDirection = .topToBottom) {
```
See suggestions below to conform to this new interface.

Explanation: Using `Color` instead of `Any` for `colors` should be pretty self-explanatory. It makes calls to the function cleaner and easier, and removes chances of user error. Using `CGFloat` instead of `NSNumber`, however, is less important because calling the function stays the same, however it's more UIKit-esque, in my opinion, and removes strange calls such as `locations: [NSNumber(value: false)]`, which would be allowed but conceptually illogical.
It would also make it consistent with the style used in `CAGradientLayerExtensions.swift`.

Sources/SwifterSwift/UIKit/UIViewExtensions.swift
```diff
@@ -472,6 +480,45 @@ public extension UIView {
...
+        // <https://github.com/swiftdevcenter/GradientColorExample>
+        let gradientLayer = CAGradientLayer()
+        gradientLayer.frame = self.bounds
+        gradientLayer.colors = colors
+        gradientLayer.locations = locations
```
guykogus, line 502:
```suggestion
        gradientLayer.colors = colors?.map { $0.cgColor }
        gradientLayer.locations = locations?.map { NSNumber(value: $0) }
```

Sources/SwifterSwift/UIKit/UIViewExtensions.swift
```diff
@@ -472,6 +480,45 @@ public extension UIView {
...
+    /// - Parameters:
+    ///   - colors: An array of colors defining the color of each gradient stop.
+    ///   - locations: An array of NSNumber defining the location of each
+    ///                gradient stop as a value in the range [0,1]. The values must be
+    ///                monotonically increasing.
+    ///   - direction: Enumeration type describing the direction of the gradient.
+    func applyGradient(colors: [Any]?, locations: [NSNumber]? = [0.0, 1.0], direction: GradientDirection = .topToBottom) {
+        // <https://github.com/swiftdevcenter/GradientColorExample>
+        let gradientLayer = CAGradientLayer()
+        gradientLayer.frame = self.bounds
```
guykogus, line 500:
```suggestion
        gradientLayer.frame = bounds
```

Sources/SwifterSwift/UIKit/UIViewExtensions.swift
```diff
@@ -472,6 +480,45 @@ public extension UIView {
         CATransaction.commit()
     }
 
+    /// SwifterSwift: Apply Gradient Colors.
+    ///
+    ///     view.applyGradient(
+    ///         colors: [UIColor.red.cgColor, UIColor.blue.cgColor],
```
guykogus, line 486: If you use my suggestions, this can be simplified
```suggestion
    ///         colors: [.red, .blue],
```

Andy0570:
I've added a CHANGELOG entry, but I'm not very good at adding tests for UIView Extension. Request for help.

Sources/SwifterSwift/UIKit/UIViewExtensions.swift
```diff
@@ -472,6 +480,45 @@ public extension UIView {
...
+    ///
+    ///     view.applyGradient(
+    ///         colors: [.red, .blue],
+    ///         locations: [0.0, 1.0],
+    ///         direction: .topToBottom
+    ///     )
+    ///
+    /// - Parameters:
+    ///   - colors: An array of colors defining the color of each gradient stop.
+    ///   - locations: An array of NSNumber defining the location of each
```
guykogus, line 493:
```suggestion
    ///   - locations: An array of `CGFloat` defining the location of each
```

Sources/SwifterSwift/UIKit/UIViewExtensions.swift
```diff
@@ -472,6 +480,45 @@ public extension UIView {
...
+    ///                gradient stop as a value in the range [0,1]. The values must be
+    ///                monotonically increasing.
+    ///   - direction: Enumeration type describing the direction of the gradient.
+    func applyGradient(colors: [Color]?, locations: [CGFloat]? = [0.0, 1.0], direction: GradientDirection = .topToBottom) {
```
guykogus, line 497: I don't think these should be optional, since a gradient doesn't make sense without them.
```suggestion
    func applyGradient(colors: [Color], locations: [CGFloat] = [0.0, 1.0], direction: GradientDirection = .topToBottom) {
```

Sources/SwifterSwift/UIKit/UIViewExtensions.swift
```diff
@@ -49,6 +49,14 @@ public extension UIView {
         /// SwifterSwift: easeInOut animation.
         case easeInOut
     }
+
+    /// SwifterSwift: Apply gradient directions
+    enum GradientDirection {
```
guykogus, line 54: I'm wondering if this should be a struct, e.g.
```swift
struct GradientDirection {
    static let topToBottom = GradientDirection(startPoint: CGPoint(x: 0.5, y: 0.0),
                                               endPoint: CGPoint(x: 0.5, y: 1.0))
    static let bottomToTop = GradientDirection(startPoint: CGPoint(x: 0.5, y: 1.0),
                                               endPoint: CGPoint(x: 0.5, y: 0.0))
    static let leftToRight = GradientDirection(startPoint: CGPoint(x: 0.0, y: 0.5),
                                               endPoint: CGPoint(x: 1.0, y: 0.5))
    static let rightToLeft = GradientDirection(startPoint: CGPoint(x: 1.0, y: 0.5),
                                               endPoint: CGPoint(x: 0.0, y: 0.5))

    let startPoint: CGPoint
    let endPoint: CGPoint
}
```
This way it can be customised to exact values.

Sources/SwifterSwift/UIKit/UIViewExtensions.swift
guykogus, line 497: You're not applying the gradient, you're actually adding it, since multiple calls to this function would add multiple gradient layers. So I'm wondering if the function name should be changed to `addGradient`?

guykogus (changes requested):
Sorry for adding more comments so late into this...

Sources/SwifterSwift/UIKit/UIViewExtensions.swift
Andy0570: I added a computed property in `GradientDirection` enum to make it more 'exact', so there is no need to add an extra struct to achieve this.

```swift
enum GradientDirection {
    case topToBottom
    case bottomToTop
    case leftToRight
    case rightToLeft

    var gradientEndpoint: (CGPoint, CGPoint) {
        switch self {
        case .topToBottom:
            return (CGPoint(x: 0.5, y: 0.0), CGPoint(x: 0.5, y: 1.0))
        case .bottomToTop:
            return (CGPoint(x: 0.5, y: 1.0), CGPoint(x: 0.5, y: 0.0))
        case .leftToRight:
            return (CGPoint(x: 0.0, y: 0.5), CGPoint(x: 1.0, y: 0.5))
        case .rightToLeft:
            return (CGPoint(x: 1.0, y: 0.5), CGPoint(x: 0.0, y: 0.5))
        }
    }
}
```

CHANGELOG.md
```diff
@@ -468,8 +469,8 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
   - Added scalar multiplication of CGFloat and CGVector via standard multiplication operator (\*). [#527](https://github.com/SwifterSwift/SwifterSwift/pull/527) by [moyerr](https://github.com/moyerr)
   - Added negation of vectors via prefix (-) operator. [#527](https://github.com/SwifterSwift/SwifterSwift/pull/527) by [moyerr](https://github.com/moyerr)
   - Added `init(angle:magnitude:)` to create vectors based on their angle and magnitude. [#527](https://github.com/SwifterSwift/SwifterSwift/pull/527) by [moyerr](https://github.com/moyerr)
--**UIRefreshControl**:
-  - `beginRefresh(in tableView:, animated:, sendAction:)` UIRefreshControl extension to begin refresh programmatically. [#525](https://github.com/SwifterSwift/SwifterSwift/pull/525) by [ratulSharker](https://github.com/ratulSharker)
+- **UIRefreshControl**:
+  - `beginRefresh(in tableView:, animated:, sendAction:)` UIRefreshControl extension to begin refresh programatically. [#525](https://github.com/SwifterSwift/SwifterSwift/pull/525) by [ratulSharker](https://github.com/ratulSharker)
```
guykogus, line 473:
```suggestion
  - `beginRefresh(in tableView:, animated:, sendAction:)` UIRefreshControl extension to begin refresh programmatically. [#525](https://github.com/SwifterSwift/SwifterSwift/pull/525) by [ratulSharker](https://github.com/ratulSharker)
```

Sources/SwifterSwift/UIKit/UIViewExtensions.swift
guykogus: But what if somebody wants to customise the direction? This is limiting.
Also, you've used a tuple instead of the struct, which is harder to read when using and is doing the same thing.

guykogus (changes requested):

Andy0570 (commented):
Fix spelling mistakes.

PR #1049, opened 2022-10-01
Update DateExtensions.swift

crpatil1901:
## Checklist

- [`x` ] I checked the [**Contributing Guidelines**](https://github.com/SwifterSwift/SwifterSwift/blob/master/CONTRIBUTING.md) before creating this request.
- [`x` ] New extensions are written in Swift 5.0.
- [`x` ] New extensions support iOS 10.0+ / tvOS 9.0+ / macOS 10.10+ / watchOS 2.0+, or use `@available` if not.
- [`x` ] All extensions have a **clear** comments explaining their functionality, all parameters and return type in English.
- [`x` ] All extensions are declared as **public**.

### DIFF
--- a/Sources/SwifterSwift/Foundation/DateExtensions.swift
+++ b/Sources/SwifterSwift/Foundation/DateExtensions.swift
@@ -491,6 +491,30 @@ public extension Date {
     var unixTimestamp: Double {
         return timeIntervalSince1970
     }
+    
+    /// SwifterSwift: Which day of the year
+    ///
+    ///     Date().dayInYear -> 276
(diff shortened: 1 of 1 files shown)


guykogus:
Don't we already have these functions? And you're missing the CHANGELOG entry and tests.

guykogus:
@crpatil1901 can you please add tests and a CHANGELOG entry? Also, you can't call between functions, so you're going to have to copy-paste code. It's part of our project style, since people mostly copy-paste code rather than include it as a dependency.

PR #1050, opened 2022-10-03
:sparkles: Add `scan` function to Sequence

cHaLkdusT:
## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 - **UIButton**
   - Added `setBackgroundColor(_:for:)` method for setting background color for the specified UI state. [#1041](https://github.com/SwifterSwift/SwifterSwift/pull/1041) by [TwizzyIndy](https://github.com/TwizzyIndy)
 - **Sequence**
-  - Added `scan(initial:combine:)` method [#1050](https://github.com/SwifterSwift/SwifterSwift/pull/1050) by [cHaLkdusT](https://github.com/cHaLkdusT)
+  - Added `cumulate(::)` method. [#1050](https://github.com/SwifterSwift/SwifterSwift/pull/1050) by [cHaLkdusT](https://github.com/cHaLkdusT)
 
 ### Fixed
  - **Digest**
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 ### Added
 - **UIButton**
   - Added `setBackgroundColor(_:for:)` method for setting background color for the specified UI state. [#1041](https://github.com/SwifterSwift/SwifterSwift/pull/1041) by [TwizzyIndy](https://github.com/TwizzyIndy)
+- **Sequence**
+  - Added `scan(initial:combine:)` method [#1050](https://github.com/SwifterSwift/SwifterSwift/pull/1050) by [cHaLkdusT](https://github.com/cHaLkdusT)
 
 ### Fixed
  - **Digest**
--- a/Sources/SwifterSwift/SwiftStdlib/SequenceExtensions.swift
+++ b/Sources/SwifterSwift/SwiftStdlib/SequenceExtensions.swift
@@ -230,6 +230,22 @@ public extension Sequence {
     func first<T: Equatable>(where keyPath: KeyPath<Element, T>, equals value: T) -> Element? {
         return first { $0[keyPath: keyPath] == value }
     }
+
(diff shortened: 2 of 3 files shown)


viktart:
I think functionality of your `scan` function is already covered by `reduce` form standard library https://developer.apple.com/documentation/swift/sequence/reduce(_:_:)

cHaLkdusT:
Hi @viktart,
* `reduce` returns `Element`
* `scan` returns `[Element]`

Sources/SwifterSwift/SwiftStdlib/SequenceExtensions.swift
```diff
@@ -230,6 +230,22 @@ public extension Sequence {
...
+
+    /// SwifterSwift: Return a sequence where every value is equal to combine called on the previous element, and the current element.
+    ///
+    ///        [1, 2, 3].scan(0, combine: +) -> [1, 3, 6]
+    ///        ["a", "b", "c"].scan("", combine: +) -> ["a", "ab", "abc"]
+    ///
+    /// - Parameter initial: Initial result
+    /// - Parameter combine: A closure that takes as its arguments the previous value returned by the closure and the next element
+    /// - Returns: A sequence of the intermediate results calling `combine` on successive elements of self and an acuumulator
+    func scan(initial: Element, combine: (Element, Element) throws -> Element) rethrows -> [Element] {
```
guykogus, line 242: `scan` isn't an appropriate name. I think something like `cumulate` makes more sense, since the results are cumulative on each call to `combine`.
Actually even `combine` doesn't fit since you're not really combining the values, you're essentially adding them.

If you want it to be like a related function to `reduce` then maybe it should have almost the same interface.
```swift
func cumulate<Result>(
    _ initialResult: Result,
    _ nextPartialResult: (Result, Self.Element) throws -> Result
) rethrows -> [Result]
```

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 ### Added
 - **UIButton**
   - Added `setBackgroundColor(_:for:)` method for setting background color for the specified UI state. [#1041](https://github.com/SwifterSwift/SwifterSwift/pull/1041) by [TwizzyIndy](https://github.com/TwizzyIndy)
+- **Sequence**
+  - Added `scan(initial:combine:)` method [#1050](https://github.com/SwifterSwift/SwifterSwift/pull/1050) by [cHaLkdusT](https://github.com/cHaLkdusT)
```
guykogus, line 11:
```suggestion
  - Added `scan(initial:combine:)` method. [#1050](https://github.com/SwifterSwift/SwifterSwift/pull/1050) by [cHaLkdusT](https://github.com/cHaLkdusT)
```

guykogus (changes requested):

Sources/SwifterSwift/SwiftStdlib/SequenceExtensions.swift
LucianoPAlmeida: I would still go with `scan` because that is common naming for this operation across many languages and also apple's `Combine` framework https://developer.apple.com/documentation/combine/fail/scan(_:_:), so follow the pattern is easier to find and use.

cHaLkdusT: I actually based the function from ReactiveX https://reactivex.io/documentation/operators/scan.html

guykogus: Hey, sorry I've been MIA @cHaLkdusT. If `scan` is already a common naming pattern then please change it back :)

Sources/SwifterSwift/SwiftStdlib/SequenceExtensions.swift
```diff
@@ -230,6 +230,25 @@ public extension Sequence {
...
+
+    /// SwifterSwift: Return a sequence where every value is equal to combine called on the previous element, and the current element.
+    ///
+    ///        [1, 2, 3].cumulate(0, +) -> [1, 3, 6]
+    ///        ["a", "b", "c"].cumulate("", +) -> ["a", "ab", "abc"]
+    ///
+    /// - Parameter initialResult: Initial result
+    /// - Parameter nextPartialResult: A closure that takes as its arguments the previous value returned by the closure and the next element
+    /// - Returns: A sequence of the intermediate results calling `combine` on successive elements of self and an acuumulator
+    func cumulate<Result>(
```
guykogus, line 242:
```suggestion
    func scan<Result>(
```

guykogus (changes requested):
Apologies, didn't realise `scan` was a common name.

guykogus:
@cHaLkdusT closing due to inactivity, but feel free to reopen

PR #1055, opened 2022-10-29
Rename Color,EdgeInset and Font typealiases to fix namespace conflicts with swiftUI Types 🚀

MussaCharles:
## Checklist

## Breaking Changes
Note: **SF** below stands for SwifterSwift

Renamed: 
**Color**
 - `public typealias Color = UIColor` ->  `public typealias SFColor = UIColor`
 - `public typealias Color = NSColor` ->  `public typealias SFColor = NSColor`
 
**EdgeInsets**
 -  `public typealias EdgeInsets = UIEdgeInsets` -> `public typealias SFEdgeInsets = UIEdgeInsets`
 - `public typealias EdgeInsets = NSEdgeInsets` ->   `public typealias SFEdgeInsets = NSEdgeInsets`
 
**Font**
 - `public typealias Font = UIFont` -> `public typealias SFFont = UIFont`
 -  `public typealias Font = NSFont` -> `public typealias SFFont = NSFont`
 
 ### Reasons for changes
 ##### Summary
 The above typealiases are renamed so that to fix namespace conflict with SwiftUI's related types. Specifically SwiftUI uses Color,EdgeInset and Font types however SwifterSwift uses the same names on it's typealiases thus resulted in build errors. 
 
 #### Associated build errors
 Importing SwifterSwift in files where Color,EdgeInsets or Font is specified as concrete types for example
 ```swift
  init(color: Color, font: Font, insets: EdgeInsets) {...}
 ```
 Causes the following Compile time build errors
 ```
 text
 'Color' is ambiguous for type lookup in this context
 'EdgeInsets' is ambiguous for type lookup in this context
 'Font' is ambiguous for type lookup in this context
 ```
 
 Sample screenshot
Screen Shot 2022-10-29 at 7 53 13 AM

The above errors were displayed on my current project. To fix those errors I had to fork SwifterSwift and do the changes specified in this PR. 

Since these might be breaking changes to users who are using the renamed public typealiases for example  Color instead of UIColor or Font instead of UIFont on legacy UIKit based apps, a major release might be needed if this PR is approved.


LucianoPAlmeida (dismissed):
This makes sense to me
Can you add a changelog entry?

MussaCharles:
Hi, @LucianoPAlmeida I just added change log entry and update PR comment. 
Thanks :)

LucianoPAlmeida (approved):

guykogus:
Why SF and not SS?
Also, this is a general issue we have with a project like SwifterSwift - there are a lot of namespace clashes, although usually with other 3rd party libraries, so I get fixing it in this particular case?
Also, why can't you just use the library name to fix the clash? E.g.
```Swift
let suiColor = SwiftUI.Color.red
let ssColor = SwifterSwift.Color.red
```

MussaCharles:
> 

About Why SF and not SS?
 - I picked SF instead of SS because easy of readability/typing, but I am open to make naming modifications if needed. 


About directly using `SwiftUI.Color` or `SwifterSwift.Color`
  -  While that could fix the issue in relation to my current project and SwifterSwift, but I thought about what could happen in the future if more SwiftUI related extensions using the similar Type declarations are added in **SwifterSwift** internally or publicly. 
  
This will require all **SwiftUI** related extensions to explicitly use  `SwiftUI.Color` or `SwiftUI.Font` or `SwiftUI.EdgeInsets` because of the collision caused by global scoped typealiases which uses the same name. 

Consider the following simple example as a new extension added to SwifterSwift. The goal is to add a new property color of type `SwiftUI.Color` but it's common to simply specify our new property is of type `Color` since we are in the same module and already imported `SwiftUI`. 

```swift
import SwiftUI

public struct Foo {}

public typealias Color = Foo

public extension View {
    static var color: Color {
        // Compiler erorr:  Type 'Color' (aka 'Foo') has no member 'blue'
        return .blue
    }
}

```

As shown in the comment above a new SwiftUI related extension will have to explicity use `SwiftUI.Color` in order to fix compiler error above. So fixing this collision before more SwiftUI related extensions are added will make it more easy to add those extensions without having to type `SwiftUI.something` just like the way most of the `UIKit` components are handled. 
Instead of saying `UIKit.UIColor` we simply use `UIColor` to denote most of the related extensions residing the same SwifterSwift module. 

@guykogus

LucianoPAlmeida:
> Why SF and not SS?

I know is totally unrelated but to me it feels a bit weird [historically](https://en.wikipedia.org/wiki/Waffen-SS)

> Also, why can't you just use the library name to fix the clash? E.g.
 
IMO since it such those alias are so common names it makes sense to rename then. Because people wouldn't want to rename all references to `Color` in a `SwiftUI` file for example just because they imported `SwifterSwift`. So that is why I think in this case renaming is fine.

guykogus:
> I know is totally unrelated but to me it feels a bit weird [historically](https://en.wikipedia.org/wiki/Waffen-SS)

I thought that might worry people but as a Jew who lost family in the holocaust I promise you it's definitely not an issue. I had a school teacher (also Jewish) with the initials SS and she had it on her license plate and it made us giggle.

> IMO since it such those alias are so common names it makes sense to rename then. Because people wouldn't want to rename all references to `Color` in a `SwiftUI` file for example just because they imported `SwifterSwift`. So that is why I think in this case renaming is fine.

Yeah, it makes sense. I just find the old two-letter prefix thing so Objective-C, which feels backwards. But it makes sense. I remember when we discussed all the other clashes that we have (especially functions like `cornerRadius`) there was a crazy discussion about how to resolve it that I think I'd rather these 2-letter prefixes 😂.

So yeah, go for it @MussaCharles, and use SF if you like.

guykogus (approved):

SD10:

PR #1064, opened 2022-12-11
add a more simple conversion between string and double.

westfort4July:
## Checklist


ShivaHuang:
The problem in your example is the decimal separator of locale "km_KH" is `,`. Thus `8.32` is not a valid number in "km_KH". The behavior of current implementation has no problem at all. However, the proposed new implementation can't handle `8,32` at locales like  "km_KH", "fr_FR".

You can add following code to specify the locale in unit test

```swift
extension NSLocale {
    @objc
    static let currentLocale = NSLocale(localeIdentifier: "km_KH") // Set a needed locale
}
```

and the test `XCTAssertNotNil("8.23".doubleValue())` should fail.

PR #1074, opened 2023-01-23
Add `DefaultStringInterpolation.appendInterpolation(_:placeholder:predicate:)` method for providing placeholder string if interpolated value is `nil` or the optional predicate closure returns `false`

ShivaHuang:
## Checklist


ShivaHuang:
Can anyone in @SwifterSwift/core check this PR?

LucianoPAlmeida:
> Can anyone in @SwifterSwift/core check this PR?

I'll try to take a look as soon as possible.

Sources/SwifterSwift/SwiftStdlib/DefaultStringInterpolationExtensions.swift
```diff
@@ -0,0 +1,46 @@
...
+    ///   - predicate: A closure that takes unwrapped `value` as its argument and returns `true` if the `value`'s
+    ///                textual representation  should be interpolated into the string literal or `false` if the
+    ///                `placeholder` should be interpolated into the string literal. You can use a `nil` value for
+    ///                this parameter.
+    mutating func appendInterpolation<T>(_ value: T?,
+                                         placeholder: @autoclosure () -> String,
+                                         predicate: ((T) throws -> Bool)? = nil) rethrows
+    {
+        switch value {
+        case .some(let value) where try predicate?(value) != false:
```
LucianoPAlmeida, line 39: Why not true evaluation?
```suggestion
        case .some(let value) where try predicate?(value) == true:
```

Sources/SwifterSwift/SwiftStdlib/DefaultStringInterpolationExtensions.swift
```diff
@@ -0,0 +1,46 @@
+// DefaultStringInterpolationExtensions.swift - Copyright 2023 SwifterSwift
```
LucianoPAlmeida, line 1: Since there is a new type extensions can you add it to the readme?

Sources/SwifterSwift/SwiftStdlib/DefaultStringInterpolationExtensions.swift
```diff
@@ -0,0 +1,46 @@
...
+    /// - Parameters:
+    ///   - value: The values, literals, variables, or expressions to be interpolated.
+    ///   - placeholder: The string that displays when `value` is `nil`.
+    ///   - predicate: A closure that takes unwrapped `value` as its argument and returns `true` if the `value`'s
+    ///                textual representation  should be interpolated into the string literal or `false` if the
+    ///                `placeholder` should be interpolated into the string literal. You can use a `nil` value for
+    ///                this parameter.
+    mutating func appendInterpolation<T>(_ value: T?,
+                                         placeholder: @autoclosure () -> String,
+                                         predicate: ((T) throws -> Bool)? = nil) rethrows
```
LucianoPAlmeida, line 36:
```suggestion
                                         where predicate: ((T) throws -> Bool)? = nil) rethrows
```

Sources/SwifterSwift/SwiftStdlib/DefaultStringInterpolationExtensions.swift
```diff
@@ -0,0 +1,46 @@
+// DefaultStringInterpolationExtensions.swift - Copyright 2023 SwifterSwift
+
+#if canImport(Foundation)
+import Foundation
```
LucianoPAlmeida, line 4: Why does it requires `Foundation`? `DefaultStringInterpolation` is defined on stdlib

Sources/SwifterSwift/SwiftStdlib/DefaultStringInterpolationExtensions.swift
ShivaHuang: It may be `nil`, it's a little trick to avoid writing `predicate?(value) == nil || predicate?(value) == true`.

LucianoPAlmeida: But If is `nil` it wouldn't be also `true` right?

LucianoPAlmeida (approved):

Sources/SwifterSwift/SwiftStdlib/DefaultStringInterpolationExtensions.swift
ShivaHuang: Updated.

Sources/SwifterSwift/SwiftStdlib/DefaultStringInterpolationExtensions.swift
ShivaHuang: Updated, it make the signature more clear, thanks!

Sources/SwifterSwift/SwiftStdlib/DefaultStringInterpolationExtensions.swift
ShivaHuang: Updated, I also added the missing item for `DecodableExtensions.swift`.

Sources/SwifterSwift/SwiftStdlib/DefaultStringInterpolationExtensions.swift
ShivaHuang: My intention is append the literal of value if
1. the given `value` is not `nil` and `predicate` is not provided
or
2. the given `value` is not `nil` and `predicate` is provided  and the test result of the value is `true`

The following code presents the logic:

```swift
case .some(let value) where (predicate == nil || try predicate?(value) == true):
            appendLiteral("\(value)")
```

`predicate?(value) != false` is just an equivalence to `predicate == nil || try predicate?(value) == true`

ShivaHuang: Another equivalence is

```swift
        case .some(let value) where try predicate?(value) ?? true:
```

do you think it's more easy to read?

LucianoPAlmeida: Ah no makes sense

PR #1110, opened 2023-05-20
Minor changes in Optional Collection extensions. :rocket:

Mayank-84:
## Checklist

- [x] I have tested new extensions with existing test cases, and they passed.

## Description
- Updated the `isNilOrEmpty` and `nonEmpty` boolean computed variables of Optional Collections.

- In existing implementation, it was using guard statements to unwrap optionals and check for nil. However, this is not efficient, considering the Collection (on which this is applied) can itself be very huge in terms of space. And assigning again the large amount of data in some temporary variable is an overhead and consumes extra O(n) space complexity. 

- Also, the collection could be multi-dimensional.

- My approach to this is simple, we do some checks directly on the collection not on temporary variables, considering we aren't using the temporary variable except for checking the nil and empty scenarios.

- I believe newer implementation is more readable and concise.

## Performance Impact

I created one playground page to test the performance gains comparing the older implementation and my newer implementation using the continuous clock ([reference](https://developer.apple.com/documentation/swift/clock/measure(_:)-2emvx?changes=_3)). On my performance tests, there was nearly an improvement of more than **90%**. I tested and calculated the average using various sample collection data.

[Code measuring the performance](https://github.com/SwifterSwift/SwifterSwift/files/11522139/TestPerformance.xcplaygroundpage.zip)

## Tests

Since I just updated the extensions so I used the existing test cases which were already implemented. All test cases were passed to 100%

## NOTE

- I was amazed to see such performance gains just by changing some checks. I hope this gets merged, so that anyone who is using those extensions can be benefited.

- Kindly, give some feedback, if we can improve it further.

- Just one request, please merge this PR, instead of committing as Co-author. This really helps me build my github profile.

-  Let's build SwifterSwift more efficient and better!  :rocket:

### DIFF
--- a/Sources/SwifterSwift/SwiftStdlib/OptionalExtensions.swift
+++ b/Sources/SwifterSwift/SwiftStdlib/OptionalExtensions.swift
@@ -91,7 +91,7 @@ public extension Optional {
 public extension Optional where Wrapped: Collection {
     /// SwifterSwift: Check if optional is nil or empty collection.
     var isNilOrEmpty: Bool {
-        return (self == nil || self!.isEmpty)
+        return (self == nil || (self?.isEmpty ?? true))
     }
 
     /// SwifterSwift: Returns the collection only if it is not nil and not empty.
--- a/Sources/SwifterSwift/SwiftStdlib/OptionalExtensions.swift
+++ b/Sources/SwifterSwift/SwiftStdlib/OptionalExtensions.swift
@@ -96,9 +96,7 @@ public extension Optional where Wrapped: Collection {
 
     /// SwifterSwift: Returns the collection only if it is not nil and not empty.
     var nonEmpty: Wrapped? {
-        guard let collection = self else { return nil }
-        guard !collection.isEmpty else { return nil }
-        return collection
+        return (self?.isEmpty ?? true) ? nil : self
     }
 }
 
--- a/Sources/SwifterSwift/SwiftStdlib/OptionalExtensions.swift
+++ b/Sources/SwifterSwift/SwiftStdlib/OptionalExtensions.swift
@@ -91,7 +91,7 @@ public extension Optional {
 public extension Optional where Wrapped: Collection {
     /// SwifterSwift: Check if optional is nil or empty collection.
     var isNilOrEmpty: Bool {
-        return (self == nil || (self?.isEmpty ?? true))
+        return self?.isEmpty ?? true
     }
 
     /// SwifterSwift: Returns the collection only if it is not nil and not empty.
--- a/Sources/SwifterSwift/SwiftStdlib/OptionalExtensions.swift
+++ b/Sources/SwifterSwift/SwiftStdlib/OptionalExtensions.swift
@@ -91,8 +91,7 @@ public extension Optional {
 public extension Optional where Wrapped: Collection {
     /// SwifterSwift: Check if optional is nil or empty collection.
     var isNilOrEmpty: Bool {
-        guard let collection = self else { return true }
-        return collection.isEmpty
+        return (self == nil || self!.isEmpty)
     }
 
     /// SwifterSwift: Returns the collection only if it is not nil and not empty.


LucianoPAlmeida (commented):
Thanks for the PR,

> On my performance tests, there was nearly an improvement of more than 90%. I tested and calculated the average using various sample collection data.

Did you make sure to run your benchmarks in release mode? Also benchmarks are tricky so results may vary from a run to another for various reasons so it a common practice in benchmarks to perform a bunch of runs and compute averages. 

It did look very odd to see that difference specially seeing that new version seems to generate as slightly worst code than older one (force unwrap may cause a extra check and jump to `ud2` instruction to terminate the program). So in all it appeared as there was no sizable difference between the two looking at the [emitted code](https://godbolt.org/z/f9b6shrbW)

So running my own benchmarks I found that there is indeed not that much difference, and actually old one is slightly better still even though is negligible. So my take is that the observed difference may be from something else other that the pure code. 

chart 

Results are similar for `[String]?`

<details>

<summary>Full code for benchmarks:</summary>

```swift
import CollectionsBenchmark

var benchmark = Benchmark(title: "NilOrEmpty")
let inputGeneratorCommon: (Int) -> String? = { size in
  return String(repeating: "a", count: size)
}

benchmark.registerInputGenerator(inputGeneratorCommon)
benchmark.add(
  title: "isNilOrEmpty old",
  input: String?.self,
  maxSize: 10000
) { input in
  return { timer in
    let d = input.isNilOrEmpty
    blackHole(d)
  }
}

benchmark.add(
  title: "isNilOrEmpty new",
  input: String?.self,
  maxSize: 10000
) { input in
  return { timer in
    let d = input.isNilOrEmptyNew
    blackHole(d)
  }
}

benchmark.add(
  title: "nonEmpty old",
  input: String?.self,
  maxSize: 10000
) { input in
  return { timer in
    let d = input.nonEmpty
    blackHole(d)
  }
}

benchmark.add(
  title: "nonEmpty new",
  input: String?.self,
  maxSize: 10000
) { input in
  return { timer in
    let d = input.nonEmptyNew
    blackHole(d)
  }
}

benchmark.main()
```

</details>

>  And assigning again the large amount of data in some temporary variable is an overhead and consumes extra O(n) space complexity.

Due to COW, we probably won't have copy on assignment here so that statement is not correct because not all assignments imply copy. Not even sure this because an assignment since is a binding on the guard...

> I believe newer implementation is more readable and concise.

That is a more personal preference take, some people may prefer using guard statements as I think is even on language guidelines for such cases. I personally would prefer the old one. 

With all that said, I'm not in favor of this change because to me it doesn't change to much in terms of perf(diff is negligible but still slightly worse) and I actually prefer the old code. So deferring this to others if they want to merge this in.

omaralbeik (dismissed):
**L**ets **G**et **T**his **M**erged ™

Sources/SwifterSwift/SwiftStdlib/OptionalExtensions.swift
```diff
@@ -91,15 +91,12 @@ public extension Optional {
 public extension Optional where Wrapped: Collection {
     /// SwifterSwift: Check if optional is nil or empty collection.
     var isNilOrEmpty: Bool {
-        guard let collection = self else { return true }
-        return collection.isEmpty
+        return (self == nil || self!.isEmpty)
```
guykogus, line 94: Why not do `self?.isEmpty ?? true` like below?

guykogus (changes requested):

Sources/SwifterSwift/SwiftStdlib/OptionalExtensions.swift
Mayank-84: Yeah we can do that. But initially I thought anyways in second OR condition it would not be nil anyways. So that's why I was force unwrapping. I will do this change. Thanks for the feedback.

Mayank-84:
@guykogus I have updated the implementation of `isNilOrEmpty`.

Sources/SwifterSwift/SwiftStdlib/OptionalExtensions.swift
```diff
@@ -91,15 +91,12 @@ public extension Optional {
 public extension Optional where Wrapped: Collection {
     /// SwifterSwift: Check if optional is nil or empty collection.
     var isNilOrEmpty: Bool {
-        guard let collection = self else { return true }
-        return collection.isEmpty
+        return (self == nil || (self?.isEmpty ?? true))
```
LucianoPAlmeida, line 94: I think this was the suggestion
```suggestion
        return self?.isEmpty ?? true
```

Mayank-84: Oh yes, sorry my bad. Thanks for pointing out.

LucianoPAlmeida (approved):

Mayank-84:
@guykogus changes are done. Can you review it? And Can we get this merged?

PR #1113, opened 2023-06-19
Reverse the parameters for the ~= String operator to not break the default implementation

### LINKED ISSUE #1109: String-based switch matching can lead to bugs
```
switch "0" {
case "101":
    print("101") // match this case.
case "1001":
    print("1001")
case "0":
    print("0")
default:
    break;
}
```
String extension **static func  ~=** overload the original implementation.
StringExtensions.swift - SwifterSwift
```
    static func ~= (lhs: String, rhs: String) -> Bool {
        return lhs.range(of: rhs, options: .regularExpression) != nil
    }
```

[StringComparable.swift](https://github.com/apple/swift/blob/448c41185d7bc391fa209cada2e142b90dbeb930/stdlib/public/core/StringComparable.swift)
```
  public static func ~= (lhs: String, rhs: Substring) -> Bool {
    return lhs == rhs
  }
```

guykogus:
🚀
Fixes #1109, where performing `switch` on a `String` was broken because the order used in `~=` was reversed.

## Checklist


LucianoPAlmeida (dismissed):
Let's go with that, but I would rather have a `matches` method in `NSRegularExpression` which it would be much clear than use this operator.

LucianoPAlmeida (approved):

PR #1119, opened 2023-07-02
Date Extension

ArvindRajpurohit81:
[✓] Date format enum added so anyone can add his own format and getFormatDate will behave like that [✓] Here we have achieved 1 solid principle as Open for extension and close for modification, so our function will never modify only we need update enum.

## Checklist

### DIFF
--- a/Sources/SwifterSwift/Foundation/DateExtensions.swift
+++ b/Sources/SwifterSwift/Foundation/DateExtensions.swift
@@ -43,6 +43,27 @@ public extension Date {
         /// SwifterSwift: Full month name.
         case full
     }
+    
+    /// SwifterSwift: we have three predefined cases (fullDateTime, shortDate, time) that represent commonly used date formats.
+    enum DateFormat {
+        case fullDateTime
+        case shortDate
+        case time
+        case custom(String)
+        
+        var getString: String {
+            switch self {
+            case .fullDateTime:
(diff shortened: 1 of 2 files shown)


Sources/SwifterSwift/Foundation/DateExtensions.swift
```diff
@@ -43,6 +43,27 @@ public extension Date {
...
+    enum DateFormat {
+        case fullDateTime
+        case shortDate
+        case time
+        case custom(String)
+        
+        var getString: String {
+            switch self {
+            case .fullDateTime:
+                return "yyyy-MM-dd HH:mm:ss"
```
viktart, line 57: It's location-specific, so I this code isn't universal. As so I don't think it's a very good idea for an extension for people around the world to use. The only good way to make this extension universal is to rename cases of the enum to something like `USFullDateTime`, `europeanFullDateTime` etc.

viktart (changes requested):

guykogus:
I'm sorry @ArvindRajpurohit81 but I don't think this is a good idea. It doesn't handle localisations at all, and it hides creating a `DateFormatter` which is actually a fairly heavy object to create, so if somebody were to call this function on a large list of dates it would cause them performance issues because they should have reused a single `DateFormatter` instance.

PR #1127, opened 2023-08-21
Feature Extend a property backgroundViewColor for UIStackView to set its background color

WZBbiao:
In versions below iOS 14, UIStackView is a non-rendering component, that is, the background color cannot be set through the backgroundColor property. Extend a property backgroundViewColor for UIStackView to set its background color.

## Checklist

### DIFF
--- a/Sources/SwifterSwift/UIKit/UIStackViewExtensions.swift
+++ b/Sources/SwifterSwift/UIKit/UIStackViewExtensions.swift
@@ -14,7 +14,7 @@ public extension UIStackView {
             if #available(iOS 14.0, *) {
                 return backgroundColor
             } else {
-               return subviews.first(where: { $0 is BackgroundView })?.backgroundColor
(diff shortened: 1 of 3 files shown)


Sources/SwifterSwift/UIKit/UIStackViewExtensions.swift
```diff
@@ -6,6 +6,44 @@ import UIKit
...
+        get {
+            if #available(iOS 14.0, *) {
+                return backgroundColor
+            } else {
+                for subview in subviews {
+                    if subview is BackgroundView {
+                        return subview.backgroundColor
+                    }
+                }
+                return nil
```
LucianoPAlmeida, line 22:
```suggestion
               return subviews.first(where: { $0 is BackgroundView })?.backgroundColor
```

LucianoPAlmeida (approved):

PR #1133, opened 2023-09-05
Add `method` operator to URLRequest

ricardorauber:
I have a bunch of operators to handle `URLRequest` creation in an easy way. This one is the first of them, to set the `httpMethod`. My idea is to make a PR for each of the operators so in the end we could create requests like this:

```swift
let request = URLRequest(url: url)
    .method("post")
    .authorization(token: token)
    .header(name: "something", value: "interesting")
    .jsonBody(parameters)
```

## Checklist

### DIFF
--- a/Sources/SwifterSwift/Foundation/URLRequestExtensions.swift
+++ b/Sources/SwifterSwift/Foundation/URLRequestExtensions.swift
@@ -47,4 +47,21 @@ public extension URLRequest {
     }
 }
 
+// MARK: - Operators
+
+public extension URLRequest {
+    /// SwifterSwift: Duplicates the request and modifies the HTTP method (verb) for the request (i.e.: GET, POST, PUT)
+    ///
+    ///     let request = URLRequest(url: url)
+    ///         .method("post")
+    ///
+    /// - Parameter methodString: The method as a String value
+    /// - Returns: The modified request
+    func method(_ methodString: String) -> Self {
+        var request = self
+        request.httpMethod = methodString.uppercased()
+        return request
+    }
+}
+
 #endif
--- a/Sources/SwifterSwift/Foundation/URLRequestExtensions.swift
+++ b/Sources/SwifterSwift/Foundation/URLRequestExtensions.swift
@@ -47,7 +47,7 @@ public extension URLRequest {
     }
 }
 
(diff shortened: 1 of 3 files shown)


Sources/SwifterSwift/Foundation/URLRequestExtensions.swift
```diff
@@ -47,4 +47,21 @@ public extension URLRequest {
     }
 }
 
+// MARK: - Operators
```
LucianoPAlmeida, line 50: Is this mark correct? The extension bellow is not an operaror

LucianoPAlmeida (commented):
A minor comment, otherwise looks good :)

ricardorauber:
Hey, thanks a lot for replying! I have followed Apple’s convention on Combine, they call methods such as flat and filter like that. But I am happy to change it to something else!https://developer.apple.com/documentation/combine/publishers-merge-publisher-operatorsEm 5 de set. de 2023, à(s) 19:43, Luciano Almeida ***@***.***> escreveu:﻿
@LucianoPAlmeida commented on this pull request.

A minor comment, otherwise looks good :)

In Sources/SwifterSwift/Foundation/URLRequestExtensions.swift:
> @@ -47,4 +47,21 @@ public extension URLRequest {
     }
 }
 
+// MARK: - Operators

Is this mark correct? The extension bellow is not an operaror

LucianoPAlmeida:
> Hey, thanks a lot for replying! I have followed Apple’s convention on Combine, they call methods such as flat and filter like that. But I am happy to change it to something else!

Is just that in our context we use `// MARK: - Operators` for extensions that are unary or binary operators, so marking a method extension with that makes things a bit confusing. Can you change it to `// MARK: - Methods`?

ricardorauber-singular:
> > Hey, thanks a lot for replying! I have followed Apple’s convention on Combine, they call methods such as flat and filter like that. But I am happy to change it to something else!
> 
> Is just that in our context we use `// MARK: - Operators` for extensions that are unary or binary operators, so marking a method extension with that makes things a bit confusing. Can you change it to `// MARK: - Methods`? Just to stay consistent across the project.

Sure, will do!

LucianoPAlmeida (approved):

PR #1134, opened 2023-09-06
Add `header(name:, value:)` method to URLRequest

ricardorauber:
Following the idea of the [#1133](https://github.com/SwifterSwift/SwifterSwift/pull/1133), this PR adds a method to URLRequest to set a header.

## Checklist

### DIFF
--- a/Sources/SwifterSwift/Foundation/URLRequestExtensions.swift
+++ b/Sources/SwifterSwift/Foundation/URLRequestExtensions.swift
@@ -62,6 +62,21 @@ public extension URLRequest {
         request.httpMethod = methodString.uppercased()
         return request
     }
+    
+    /// SwifterSwift: Duplicates the request and set a header with key and value. It will use String(describing: value) as the value should be a string
+    ///
+    ///     let request = URLRequest(url: url)
+    ///         .header(name: "something", value: "interesting")
+    ///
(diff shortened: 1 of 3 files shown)


Sources/SwifterSwift/Foundation/URLRequestExtensions.swift
```diff
@@ -62,6 +62,21 @@ public extension URLRequest {
...
+    /// SwifterSwift: Duplicates the request and set a header with key and value. It will use String(describing: value) as the value should be a string
+    ///
+    ///     let request = URLRequest(url: url)
+    ///         .header(name: "something", value: "interesting")
+    ///
+    /// - Parameters:
+    ///   - name: The name of the header
+    ///   - value: The value of the header
+    /// - Returns: The modified request
+    func header(name: String, value: Any) -> Self {
```
LucianoPAlmeida, line 75: Let's keep string here, if users want to use that with other types is easy enough to add this to the call e.g. `.header(name: "Number", value: "\(10)")`
```suggestion
    func header(name: String, value: String) -> Self {
```

ricardorauber: Done!

LucianoPAlmeida (approved):

PR #1135, opened 2023-09-07
Add `headers()` method to URLRequest

ricardorauber:
Following the idea of the [#1133](https://github.com/SwifterSwift/SwifterSwift/pull/1133), this PR adds a method to URLRequest to set headers with a dictionary.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,7 +37,7 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
   - Added `allQueryParameters`, `appendingQueryParameters(_:)` and `appendQueryParameters(_:)` for using `URLQueryItem`, as an addition to the `[String: String]` variants, to handle `nil`-value query parameters. [#1116](https://github.com/SwifterSwift/SwifterSwift/issues/1116) by [guykogus](https://github.com/guykogus)
 - **URLRequest**
   - Added `method()` to duplicate the request and modify the HTTP method (verb) for the request (i.e.: GET, POST, PUT) [#1133](https://github.com/SwifterSwift/SwifterSwift/pull/1133) by [Ricardo Rauber](https://github.com/ricardorauber)
-  - Added `header(name:, value:)` to duplicate the request and set a header with key and value. It will use String(describing: value) as the value should be a string
+  - Added `header(name:, value:)` to duplicate the request and set a header with key and value. It will use String(describing: value) as the value should be a string [#1134](https://github.com/SwifterSwift/SwifterSwift/pull/1134) by [Ricardo Rauber](https://github.com/ricardorauber)
 - **URLSession**
   - Added `dataSync(for:)` to make requests synchronously. [#1076](https://github.com/SwifterSwift/SwifterSwift/pull/1076) by [Roman Podymov](https://github.com/RomanPodymov)
 - **UIStackView**
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
   - Added `allQueryParameters`, `appendingQueryParameters(_:)` and `appendQueryParameters(_:)` for using `URLQueryItem`, as an addition to the `[String: String]` variants, to handle `nil`-value query parameters. [#1116](https://github.com/SwifterSwift/SwifterSwift/issues/1116) by [guykogus](https://github.com/guykogus)
 - **URLRequest**
   - Added `method()` to duplicate the request and modify the HTTP method (verb) for the request (i.e.: GET, POST, PUT) [#1133](https://github.com/SwifterSwift/SwifterSwift/pull/1133) by [Ricardo Rauber](https://github.com/ricardorauber)
+  - Added `header(name:, value:)` to duplicate the request and set a header with key and value. It will use String(describing: value) as the value should be a string
 - **URLSession**
   - Added `dataSync(for:)` to make requests synchronously. [#1076](https://github.com/SwifterSwift/SwifterSwift/pull/1076) by [Roman Podymov](https://github.com/RomanPodymov)
 - **UIStackView**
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -38,6 +38,7 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 - **URLRequest**
   - Added `method()` to duplicate the request and modify the HTTP method (verb) for the request (i.e.: GET, POST, PUT) [#1133](https://github.com/SwifterSwift/SwifterSwift/pull/1133) by [Ricardo Rauber](https://github.com/ricardorauber)
   - Added `header(name:, value:)` to duplicate the request and set a header with key and value. [#1134](https://github.com/SwifterSwift/SwifterSwift/pull/1134) by [Ricardo Rauber](https://github.com/ricardorauber)
+  - Added `headers()` to duplicate the request and set headers with key and value from a dictionary. [#1135](https://github.com/SwifterSwift/SwifterSwift/pull/1135) by [Ricardo Rauber](https://github.com/ricardorauber)
```
guykogus, line 41:
```suggestion
  - Added `adding(headers:)` to duplicate the request and set headers with key and value from a dictionary. [#1135](https://github.com/SwifterSwift/SwifterSwift/pull/1135) by [Ricardo Rauber](https://github.com/ricardorauber)
```

Sources/SwifterSwift/Foundation/URLRequestExtensions.swift
```diff
@@ -77,6 +77,24 @@ public extension URLRequest {
...
+    ///
+    ///     let request = URLRequest(url: url)
+    ///         .headers([
+    ///             "Content-Type": "application/json",
+    ///             "Authorization": "Bear abcd"
+    ///         ])
+    ///
+    /// - Parameter headers: A dictionary with headers
+    /// - Returns: The modified request
+    func headers(_ headers: [String: String]) -> Self {
```
guykogus, line 91:
```suggestion
    func adding(headers: [String: String]) -> Self {
```

Tests/FoundationTests/URLRequestExtensionsTests.swift
```diff
@@ -82,6 +82,17 @@ final class URLRequestExtensionsTests: XCTestCase {
         XCTAssertEqual(request.allHTTPHeaderFields!["Content-Type"]!, "application/json")
         XCTAssertEqual(Int(request.allHTTPHeaderFields!["Number"]!), 10)
     }
+    
+    func testHeaders() throws {
```
guykogus, line 86: Missing tests for headers that already exist or an empty dictionary.

guykogus (changes requested):
I can see that you added a few other related functions recently, and honestly I don't see the point of this. In the tests we see the example
```swift
let request = URLRequest(url: url)
    .headers([
        "Content-Type": "application/json",
        "Authorization": "Bear abcd"
    ])
```
but you could have just as easily done
```swift
var request = URLRequest(url: url)
request.setValue("application/json", forHTTPHeaderField: "Content-Type")
request.setValue("Bear abcd", forHTTPHeaderField: "Authorization")
```
Not only is this code shorter and cleaner, it's also more performant as you don't have to create a copy as part of the operation.

The naming convention is also wrong, as calling `headers` should return the headers, not create a copy with extra headers.

Perhaps what you were really looking for, in order to make something better, is to create an initialiser that adds everything you want at once. E.g.
```swift
init(url: URL,
     cachePolicy: URLRequest.CachePolicy = .useProtocolCachePolicy,
     timeoutInterval: TimeInterval = 60.0,
     method: String,
     headers: [String: String]) {
    self.init(url: url, cachePolicy: cachePolicy, timeoutInterval: timeoutInterval)
    httpMethod = method
    for (field, value) in headers {
        setValue(value, forHTTPHeaderField: field)
    }
}
```

Would you like to reconsider your strategy here?

ricardorauber:
Hey @guykogus thanks for all your comments! 

Your idea of having an initializer is great but, as I mentioned on [#1133](https://github.com/SwifterSwift/SwifterSwift/pull/1133), my idea is to have something like this:
```swift
let request = URLRequest(url: url)
    .method("post")
    .authorization(token: token)
    .header(name: "something", value: "interesting")
    .jsonBody(parameters)
```
Both ideas are great and can work seamless within this package, I believe we can add them all. 

About the naming convention, I did like Apple on `SwiftUI`:
```swift
Text("Something")
    .bold()
    .frame(maxWidth: .infinity)
    .padding(12)
    .background(.blue)
    .foregroundColor(.white)
    .cornerRadius(12)
```
I understand your point, but this way it looks and feels familiar to anyone who works with `SwiftUI` or even `Combine`.

guykogus:
@ricardorauber you're totally right, I got stuck in the old way of thinking. I'll update my review comments (some still stand).

CHANGELOG.md
guykogus, line 41:
```suggestion
  - Added `method(_:)` to duplicate the request and modify the HTTP method (verb) for the request (i.e.: GET, POST, PUT) [#1133](https://github.com/SwifterSwift/SwifterSwift/pull/1133) by [Ricardo Rauber](https://github.com/ricardorauber)
  - Added `header(name:value:)` to duplicate the request and set a header with key and value. [#1134](https://github.com/SwifterSwift/SwifterSwift/pull/1134) by [Ricardo Rauber](https://github.com/ricardorauber)
  - Added `headers(_:)` to duplicate the request and set headers with key and value from a dictionary. [#1135](https://github.com/SwifterSwift/SwifterSwift/pull/1135) by [Ricardo Rauber](https://github.com/ricardorauber)
```

guykogus (changes requested):

guykogus:
@ricardorauber ok, rethinking this a bit more, sorry 😅
The reason for SwiftUI to use that style is because of `ViewModifier`s, which are designed to be generic and work with any `View`. Similar to Combine, where it deals a lot with generics. You've taken a generic approach and applied its _style_ to a non-generic problem that has a pretty heavy downside, being performance. I think a single initialiser suits your problem much, much better.

ricardorauber:
@guykogus We could have both of them, I can create a new PR with the init as well, it will be great! 

I believe we can keep both approaches and each developer could choose the way they want. Also, there is a possibility in the chained approach to make the request execution even simpler:
```swift
func someService() async throws {
    ...
    let (data, response) = try await URLRequest(url: url)
        .method("post")
        .authorization(token: token)
        .jsonBody(parameters)
        .send(on: session)
    ...
}
```

guykogus:
@ricardorauber it doesn't make the approach simpler, it's the exact same:
```swift
func someService() async throws {
>     ...
    let (data, response) = try await URLRequest(url: url,
        method: "post",
        authorizationToken: token,
        jsonBody: parameters)
        .send(on: session)
    ...
}
```
Again, the only difference here being that in one instance you make several copies of the object and in the other you make just the 1.

Nevertheless, since we've already started tumbling down this rabbit hole I'll accept the PR, you just need to fix the comments.

PR #1136, opened 2023-09-17
Fixes tests for iOS 17

guykogus:
🚀
Some tests fail in iOS 17 and need to be updated. E.g. `URL(string:)` has changed the way that it parses strings.
> For apps linked on or after iOS 17 and aligned OS versions, [NSURL](https://developer.apple.com/documentation/foundation/nsurl) parsing has updated from the obsolete RFC 1738/1808 parsing to the same [RFC 3986](https://www.ietf.org/rfc/rfc3986.txt) parsing as [NSURLComponents](https://developer.apple.com/documentation/foundation/nsurlcomponents). This unifies the parsing behaviors of the NSURL and NSURLComponents APIs. Now, NSURL automatically percent- and IDNA-encodes invalid characters to help create a valid URL.

## Checklist


guykogus:
@LucianoPAlmeida I fixed most of the tests, but some are not working. Wanna have a look and play with it? It looks like in iOS 17 the way images are created using the renderer have extra data somewhere inside that creates mismatches.

Sources/SwifterSwift/UIKit/UIImageExtensions.swift
```diff
@@ -102,11 +102,23 @@ public extension UIImage {
...
-        UIGraphicsBeginImageContextWithOptions(CGSize(width: newWidth, height: toHeight), opaque, self.scale)
-        draw(in: CGRect(x: 0, y: 0, width: newWidth, height: toHeight))
-        let newImage = UIGraphicsGetImageFromCurrentImageContext()
-        UIGraphicsEndImageContext()
-        return newImage
+        let size = CGSize(width: newWidth, height: toHeight)
+        let rect = CGRect(origin: .zero, size: size)
+
+        #if !os(watchOS)
+        if #available(tvOS 10.0, *) {
```
LucianoPAlmeida, line 109: We have iOS 10 as minimum target deployment for the project, which personally think is too old, but I think we can just drop the old code and remove the availability checks across. Maybe even move tvOS to 10 as well. But that is another discussion

guykogus:
@LucianoPAlmeida looks like I finally fixed all the tests and compilations and swiftlint and whatnot. Will you have a check?

And what about deployment versions? Should we use the minimum recommended by Xcode 15? E.g. iOS/tvOS 12? We can do that in a separate PR.

LucianoPAlmeida:
> Should we use the minimum recommended by Xcode 15? E.g. iOS/tvOS 12? We can do that in a separate PR.

I think yes, we can definitely experiment with that in another PR :)

PR #1143, opened 2023-10-09
Added method for toggling between values for Equatable and CaseIterable types 🚀

wiencheck:
The method below allows to easily change value of the caller to the next available case.

Example usage would look like this:
```
         enum ShuffleMode: CaseIterable {
            case off, songs, albums
         }
    
         var mode: ShuffleMode = .off
         mode.toggle() // .songs

         var mode: ShuffleMode = .albums
         mode.toggle(between: [.songs, .albums]) // .songs
```

The only requirement for this method is that the caller conforms to `Equatable`. When conforming to `CaseIterable` another method is available that passes `allCases` as the parameter for `values`.

## Checklist

### DIFF
--- a/Sources/SwifterSwift/SwiftStdlib/CaseIterableExtensions.swift
+++ b/Sources/SwifterSwift/SwiftStdlib/CaseIterableExtensions.swift
@@ -1,0 +1,47 @@
+// CaseIterableExtensions.swift - Copyright 2023 SwifterSwift
+
+import Foundation
+
+public extension CaseIterable where Self: Equatable, AllCases == [Self] {
+    
+    /// SwifterSwift: Sets the value to next available case from given array.
+    ///
+    ///     enum ShuffleMode: CaseIterable {
+    ///        case off, songs, albums
+    ///     }
+    ///
+    ///     var mode: ShuffleMode = .off
+    ///     mode.toggle() // .songs
+    ///
+    ///     var mode: ShuffleMode = .albums
+    ///     mode.toggle() // .off
+    ///
+    ///     var mode: ShuffleMode = .albums
+    ///     mode.toggle(cycle: false) // .albums
+    ///
+    ///     var mode: ShuffleMode = .albums
+    ///     mode.toggle(between: [.songs, .albums]) // .songs
+    ///
+    /// - Parameters:
+    ///    - cases: Array of values that can be assigned to the caller.
+    ///    If empty array is passed the possible values will be taken from `allCases` array.
+    ///    - cycle: Indicates whether the value should be modified if it's already at the end of `values` array.
+    ///
+    mutating func toggle(
+        between values: [Self] = Self.allCases,
+        cycle: Bool = false
+    ) {
+        guard var index = values.firstIndex(of: self) else {
+            fatalError("Case \(self) must be included in `values` array.")
+        }
+        if index < (values.count - 1) {
+            index += 1
+        }
+        else {
+            if !cycle { return }
+            index = 0
+        }
+        self = values[index]
+    }
+    
+}
--- a/Sources/SwifterSwift/SwiftStdlib/EquatableExtensions.swift
+++ b/Sources/SwifterSwift/SwiftStdlib/EquatableExtensions.swift
@@ -1,36 +1,39 @@
-// CaseIterableExtensions.swift - Copyright 2023 SwifterSwift
+// EquatableExtensions.swift - Copyright 2023 SwifterSwift
 
 import Foundation
 
-public extension CaseIterable where Self: Equatable, AllCases == [Self] {
+public extension Equatable {
     
-    /// SwifterSwift: Sets the value to next available case from `allCases` array.
+    /// SwifterSwift: Sets the value to next available value from given array.
     ///
-    ///     enum ShuffleMode: CaseIterable {
+    ///     enum ShuffleMode {
     ///        case off, songs, albums
     ///     }
     ///
+    ///     let values: [ShuffleMode] = [.off, .songs, .albums]
+    ///
     ///     var mode: ShuffleMode = .off
-    ///     mode.toggle() // .songs
+    ///     mode.toggle(between: values) // .songs
     ///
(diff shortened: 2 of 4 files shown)


Sources/SwifterSwift/SwiftStdlib/CaseIterableExtensions.swift
```diff
@@ -0,0 +1,45 @@
...
+    ///     var mode: ShuffleMode = .albums
+    ///     mode.toggle(cycle: false) // .albums
+    ///
+    ///     var mode: ShuffleMode = .albums
+    ///     mode.toggle(between: [.songs, .albums]) // .songs
+    ///
+    /// - Parameters:
+    ///    - cycle: Indicates whether the value should be modified if it's already at the end of `values` array.
+    ///
+    mutating func toggle(
```
LucianoPAlmeida, line 28: I understand the reason for this but that seems too specific for extensions on a generic type like `CaseIterable`. This to me seems the kind of problem we solve with a different abstraction e.g. `CircularList`
```swift
let cl = CircularList(sequence: ShuffleMode.allCases)
cl.advanceTo(element: .off)
cl.next() // .songs
```
The meaning of exactly what this code is doing is very clear just reading it whether:
```swift
var mode: ShuffleMode = .off
mode.toggle() // Reading this code, what does toggle mean? 
```
It is just not clear at point of use. So IMO I would go into a different abstraction for this use case.

LucianoPAlmeida (commented):
Thank you for investing time into this PR, but to me I don't think this extension is a good fit for reasons explained before. But can be a matter of personal opinion and I'll let other give their thoughts. Maybe is still valid if we can be more verbose on naming for it to be more clear at point of use but to me I would still go with another approach for this kind of problem =]

Sources/SwifterSwift/SwiftStdlib/CaseIterableExtensions.swift
guykogus: I think that the issue is mostly in the naming, as "toggling" a value is for switching a value on or off, like you have with `Bool`, or you could use it with `UISwitch`, if you wanted to add an extension there.
If you simply renamed this to `nextCase` then I think it could make sense for a `CaseIterable` value to have this function.

guykogus: You can be more generic with this solution (no need for `Array`)
```swift
public extension CaseIterable where Self: Equatable {
    func nextCase(cycle: Bool = false) -> Self {
        let values = Self.allCases
        guard let currentIndex = values.firstIndex(of: self) else {
            fatalError("Case \(self) must be included in `values` array.")
        }
        var index = values.index(after: currentIndex)
        if index == values.endIndex {
            if cycle {
                index = values.startIndex
            } else {
                index = currentIndex
            }
        }
        return values[index]
    }
}
```

Tests/SwiftStdlibTests/CaseIterableExtensionsTests.swift
```diff
@@ -0,0 +1,51 @@
...
+@testable import SwifterSwift
+import XCTest
+
+final class CaseIterableExtensionsTests: XCTestCase {
+    
+    private enum Test: CaseIterable {
+        case test1, test2, test3
+    }
+    
+    func testNextCase() {
```
guykogus, line 12: Only write 1 test case for each extension.

Tests/SwiftStdlibTests/CaseIterableExtensionsTests.swift
```diff
@@ -0,0 +1,51 @@
+// CaseIterableExtensionsTests.swift - Copyright 2023 SwifterSwift
+
+@testable import SwifterSwift
+import XCTest
+
+final class CaseIterableExtensionsTests: XCTestCase {
+    
+    private enum Test: CaseIterable {
```
guykogus, line 8: Can you perform this without making the `enum` conform to `Equatable`?
```suggestion
    private enum Test: CaseIterable, Equatable {
```

Sources/SwifterSwift/SwiftStdlib/EquatableExtensions.swift
```diff
@@ -0,0 +1,48 @@
...
+    ///     mode.toggle(between: values, cycle: false) // .albums
+    ///
+    ///     var mode: ShuffleMode = .albums
+    ///     mode.toggle(between: [.songs, .albums]) // .songs
+    ///
+    /// - Parameters:
+    ///    - cases: Array of values that can be assigned to the caller.
+    ///    - cycle: Indicates whether the value should be modified if it's already at the end of `values` array.
+    ///
+    mutating func toggle(
```
guykogus, line 31: Remove this extension function. It doesn't make sense, as @LucianoPAlmeida explained and I responded. It only belongs in `CaseIterable`.

guykogus (changes requested):
Please make sure that you read the CONTRIBUTING guidelines. You're still missing a CHANGELOG entry and please update the READMEs to include this new class.

wiencheck:
Thank you @guykogus @LucianoPAlmeida for looking at the PR. I agree with your comments about naming and making this extension available only for CaseIterable types. I will update the PR and add changelog entry

Sources/SwifterSwift/SwiftStdlib/CaseIterableExtensions.swift
LucianoPAlmeida: Sound fair enough, 
Only one other point I would add is for cycle `false` is a bit odd for calling next on last element and for that operation to be no-op without caller receiving a result. Meaning calling that api would advance or not but there is no way for me as a caller to know that other than kind of do compare after. 
Maybe a case to have two functions one for cycle that returns non-optional because it would always have a value and the non-cycle returns nil on next for the last element?

guykogus: Good point about the no-op. So maybe we should have `nextCase` returning `Self?` and `nextCaseCycled` returning `Self`. WDYT @wiencheck?

PR #1167, opened 2024-01-25
Improved extension methods for Sequence

MartonioJunior:
:rocket:

Was packaging up some utility methods and discovered this package, so I decided to send them here as contributions.
I may have some more later down the line, but for now here's the summary:
- Added `product` methods for multiplying elements and properties in a sequence
- Added `contains` that matches and compares the fields.
- Since Swift 5.2, you can pass `KeyPath` expressions as functions, so methods in Sequence now are more flexible to allow for any mapping function to be used.
- Changed `contains` parameter to be generic for any type conforming to `Sequence` 

## Checklist

- [X] New extensions are written in Swift 5.6.
- [ ] New extensions support iOS 12.0+ / tvOS 12.0+ / macOS 10.13+ / watchOS 4.0+, or use `@available` if not.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,15 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 - **UIView**
   - Added `removeBlur()` method for removing the applied blur effect from the view. [#1159](https://github.com/SwifterSwift/SwifterSwift/pull/1159) by [regi93](https://github.com/regi93)
   - Added `makeCircle(diameter:)` method to make the view circular. [#1165](https://github.com/SwifterSwift/SwifterSwift/pull/1165) by [happyduck-git](https://github.com/happyduck-git)
+- **Sequence**
+  - Added `product()` for calculating the product of all `Numeric` elements. [#1167](https://github.com/SwifterSwift/SwifterSwift/pull/1167)[MartonioJunior](https://github.com/MartonioJunior)
+  - Added `product(for:)` for calculating the product of the `Numeric` property for all elements in `Sequence`. [#1167](https://github.com/SwifterSwift/SwifterSwift/pull/1167)[MartonioJunior](https://github.com/MartonioJunior)
+  - Added `contains(_:matchedBy:)` to allow checking for values with a matcher that compares values.
+
+### Changed
+- **Sequence**
+    - `sorted(by:)`, `sorted(by:with:)`, `sorted(by:and:)`, `sorted(by:and:and:)`, `sum(for:)`, `product(for:)`, `first(where:equals:)` now receive functions as parameters. This change maintains compatibility with KeyPath while making the methods more flexible. [#1167](https://github.com/SwifterSwift/SwifterSwift/pull/1167)[MartonioJunior](https://github.com/MartonioJunior)
+    - `contains(_:)` for both `Element: Hashable` and `Element: Equatable` now can receive any type that conforms to `Sequence`, not just `Array`. [#1167](https://github.com/SwifterSwift/SwifterSwift/pull/1167)[MartonioJunior](https://github.com/MartonioJunior)
 
 ### Fixed
 - **UIImageView**
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,7 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 ### Changed
 - **Sequence**
     - `sorted(by:)`, `sorted(by:with:)`, `sorted(by:and:)`, `sorted(by:and:and:)`, `sum(for:)`, `product(for:)`, `first(where:equals:)` now receive functions as parameters. This change maintains compatibility with KeyPath while making the methods more flexible. [#1167](https://github.com/SwifterSwift/SwifterSwift/pull/1167)[MartonioJunior](https://github.com/MartonioJunior)
-    - `contains(_:)` for both `Element: Hashable` and `Element: Equatable` now can receive any type that conforms to `Sequence`, not just `Array`. [#1167](https://github.com/SwifterSwift/SwifterSwift/pull/1167)[MartonioJunior](https://github.com/MartonioJunior)
+    - `contains(_:)` for `Element: Hashable` now can receive any type that conforms to `Sequence`, not just `Array`. [#1167](https://github.com/SwifterSwift/SwifterSwift/pull/1167)[MartonioJunior](https://github.com/MartonioJunior)
 
 ### Fixed
 - **UIImageView**
--- a/Sources/SwifterSwift/SwiftStdlib/SequenceExtensions.swift
+++ b/Sources/SwifterSwift/SwiftStdlib/SequenceExtensions.swift
@@ -240,31 +240,31 @@ public extension Sequence {
 }
 
 public extension Sequence where Element: Equatable {
-    /// SwifterSwift: Check if array contains an array of elements.
+    /// SwifterSwift: Check if sequence contains elements of another sequence.
     ///
     ///        [1, 2, 3, 4, 5].contains([1, 2]) -> true
     ///        [1.2, 2.3, 4.5, 3.4, 4.5].contains([2, 6]) -> false
     ///        ["h", "e", "l", "l", "o"].contains(["l", "o"]) -> true
     ///
-    /// - Parameter elements: array of elements to check.
+    /// - Parameter elements: sequence of elements to check.
     /// - Returns: true if array contains all given items.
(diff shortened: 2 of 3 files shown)


Sources/SwifterSwift/SwiftStdlib/SequenceExtensions.swift
```diff
@@ -308,3 +335,14 @@ public extension Sequence where Element: AdditiveArithmetic {
         return reduce(.zero, +)
     }
 }
+
+// MARK: - Methods (Numeric)
+
+public extension Sequence where Element: Numeric {
+    /// SwifterSwift: Product of all elements in array.
```
LucianoPAlmeida, line 342:
```suggestion
    /// SwifterSwift: Product of all elements in sequence.
```
Probably worth changing sum as well.

Sources/SwifterSwift/SwiftStdlib/SequenceExtensions.swift
```diff
@@ -105,6 +105,23 @@ public extension Sequence {
     func filtered<T>(_ isIncluded: (Element) throws -> Bool, map transform: (Element) throws -> T) rethrows -> [T] {
         return try lazy.filter(isIncluded).map(transform)
     }
+    
+    /// SwifterSwift: Contains operations based on matching function
+    ///
+    ///      ["James", "Wade", "Bryant"].contains([4,5], for: { $0.count == $1 }) -> true
```
LucianoPAlmeida, line 111: I know there are other convenience functions like that already here but I think this is something that can be already achieved by `["James", "Wade", "Bryant"].lazy.map(\.count).contains([4,5])` so I don't see a big win with this one. But if others are fine with including this I don't see a problem.

Sources/SwifterSwift/SwiftStdlib/SequenceExtensions.swift
```diff
@@ -161,110 +178,120 @@ public extension Sequence {
...
     ///        [1, 2, 3, 4, 5].contains([1, 2]) -> true
     ///        [1.2, 2.3, 4.5, 3.4, 4.5].contains([2, 6]) -> false
     ///        ["h", "e", "l", "l", "o"].contains(["l", "o"]) -> true
     ///
-    /// - Parameter elements: array of elements to check.
+    /// - Parameter elements: sequence of elements to check.
     /// - Returns: true if array contains all given items.
     /// - Complexity: _O(m·n)_, where _m_ is the length of `elements` and _n_ is the length of this sequence.
-    func contains(_ elements: [Element]) -> Bool {
+    func contains<T: Sequence>(_ elements: T) -> Bool where Element == T.Element {
```
LucianoPAlmeida, line 279: Although small, this could be a source breaking change, as `[Element]` is more specialized, in a place where this is being choose by overload resolution, with this change other method can be considered ambiguous or be choose instead. Let's keep it as is, and if we would remove, let's do in a separate PR where we deprecate the old one and then introduce the new generic signature.

Sources/SwifterSwift/SwiftStdlib/SequenceExtensions.swift
MartonioJunior: I'll wait for input then to see what others think.

Sources/SwifterSwift/SwiftStdlib/SequenceExtensions.swift
MartonioJunior: Alright, I'll work on that as soon as I can.

Sources/SwifterSwift/SwiftStdlib/SequenceExtensions.swift
guykogus: I don't really see the purpose of this function. I read the example as "the array of names contains an array of numbers for a given comparison between the length of the names and each number". It would be very unclear to read this code in a real-world project. The example given by Luciano, however, is much clearer and just as simple to write.

Sources/SwifterSwift/SwiftStdlib/SequenceExtensions.swift
```diff
@@ -161,86 +178,96 @@ public extension Sequence {
...
-    /// SwifterSwift: Return a sorted array based on a key path and a compare function.
+    /// SwifterSwift: Return a sorted array based on a map function and a compare function.
     ///
-    /// - Parameter keyPath: Key path to sort by.
+    /// - Parameter map: Function that defines the property to sort by.
     /// - Parameter compare: Comparison function that will determine the ordering.
     /// - Returns: The sorted array.
-    func sorted<T>(by keyPath: KeyPath<Element, T>, with compare: (T, T) -> Bool) -> [Element] {
-        return sorted { compare($0[keyPath: keyPath], $1[keyPath: keyPath]) }
+    func sorted<T>(by map: (Element) -> T, with compare: (T, T) -> Bool) -> [Element] {
```
guykogus, line 186: This is a breaking change as you're changing the API of these functions. Even though direct usages should remain usable, any function that accepts a `KeyPath` and then passes it on to this function will fail to compile.
```swift
func mySorted<T>(by keyPath: KeyPath<Element, T>, with compare: (T, T) -> Bool) -> [Element] {
    return sorted(by: keyPath, with: compare) // error, as the `by` parameter type has changed
}
```
You can just add your version alongside it, to avoid the breaking change.

CHANGELOG.md
```diff
@@ -13,6 +13,15 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
...
   - Added `makeCircle(diameter:)` method to make the view circular. [#1165](https://github.com/SwifterSwift/SwifterSwift/pull/1165) by [happyduck-git](https://github.com/happyduck-git)
+- **Sequence**
+  - Added `product()` for calculating the product of all `Numeric` elements. [#1167](https://github.com/SwifterSwift/SwifterSwift/pull/1167)[MartonioJunior](https://github.com/MartonioJunior)
+  - Added `product(for:)` for calculating the product of the `Numeric` property for all elements in `Sequence`. [#1167](https://github.com/SwifterSwift/SwifterSwift/pull/1167)[MartonioJunior](https://github.com/MartonioJunior)
+  - Added `contains(_:matchedBy:)` to allow checking for values with a matcher that compares values.
+
+### Changed
+- **Sequence**
+    - `sorted(by:)`, `sorted(by:with:)`, `sorted(by:and:)`, `sorted(by:and:and:)`, `sum(for:)`, `product(for:)`, `first(where:equals:)` now receive functions as parameters. This change maintains compatibility with KeyPath while making the methods more flexible. [#1167](https://github.com/SwifterSwift/SwifterSwift/pull/1167)[MartonioJunior](https://github.com/MartonioJunior)
```
guykogus, line 23: As commented below, these are breaking changes, which we should avoid. They should be separate additions.
```suggestion
- **Sequence**
  - `sorted(by:)`, `sorted(by:with:)`, `sorted(by:and:)`, `sorted(by:and:and:)`, `sum(for:)`, `product(for:)`, `first(where:equals:)` now receive functions as parameters. This change maintains compatibility with KeyPath while making the methods more flexible. [#1167](https://github.com/SwifterSwift/SwifterSwift/pull/1167)[MartonioJunior](https://github.com/MartonioJunior)
```

CHANGELOG.md
```diff
@@ -13,6 +13,15 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
...
+    - `contains(_:)` for `Element: Hashable` now can receive any type that conforms to `Sequence`, not just `Array`. [#1167](https://github.com/SwifterSwift/SwifterSwift/pull/1167)[MartonioJunior](https://github.com/MartonioJunior)
```
guykogus, line 24:
```suggestion
  - `contains(_:)` for `Element: Hashable` now can receive any type that conforms to `Sequence`, not just `Array`. [#1167](https://github.com/SwifterSwift/SwifterSwift/pull/1167)[MartonioJunior](https://github.com/MartonioJunior)
```

guykogus (changes requested):
Some great changes here. Can we please split this into separate PRs? Each PR should be specifically for its own change, e.g.
- one for `product()` and `product(for:)`
- one for adding alternatives for the KeyPaths
- one for updating `contains(_:)`

MartonioJunior:
Okay, I'll rework this PR to be based on the alternatives to KeyPaths and send the other two as separate PRs

MartonioJunior:
> Okay, I'll rework this PR to be based on the alternatives to KeyPath and send the other two (product and contains) as separate PRs

Decided it's best to keep everything as new, separate PRs (#1168, #1169, #1170). As such, I'll be closing this PR.

PR #1187, opened 2024-04-11
Create separate SPM packages

### LINKED ISSUE #1023: Refactor Package.swift to support sub modules
Introduce sub modules in Package.swift, similar to how pod subspecs are organized

guykogus:
🚀

- Creates a separate SPM package for each library extension.
- Maintains the original `import SwifterSwift` usage so that there weren't be issues with backwards compatibility.
- Removes the old Xcode project and workspace and change CI tests to use the Swift packages, since they can now run tests on all platforms (including watchOS!)

Fixes #1023

## Checklist


guykogus:
@LucianoPAlmeida what do you think? I was wondering if maybe we need to completely refactor it out so that each extension package exists in a separate folder. I.e. at the root of the project have `SwifterSwift-AppKit/Sources/*`. This seems to work, too, but it's another option.

LucianoPAlmeida:
> @LucianoPAlmeida what do you think? I was wondering if maybe we need to completely refactor it out so that each extension package exists in a separate folder. I.e. at the root of the project have `SwifterSwift-AppKit/Sources/*`. This seems to work, too, but it's another option.

I think this is fine if that works and such a refactor will be much work for no much value if the way it is serves the purpose.

PR #1191, opened 2024-04-26
fixed appendingQueryParameters and add deletingAllQueryParameters

tospery:
1. Modify appendingQueryParameters method, if the parameter already exists, replace it. 
2. Add deletingAllQueryParameters/deleteAllQueryParameters methods support.

### DIFF
--- a/Sources/SwifterSwift/Foundation/URLExtensions.swift
+++ b/Sources/SwifterSwift/Foundation/URLExtensions.swift
@@ -45,9 +45,15 @@ public extension URL {
     /// - Returns: URL with appending given query parameters.
     func appendingQueryParameters(_ parameters: [String: String]) -> URL {
         var urlComponents = URLComponents(url: self, resolvingAgainstBaseURL: true)!
-        var items = urlComponents.queryItems ?? []
-        items += parameters.map({ URLQueryItem(name: $0, value: $1) })
-        urlComponents.queryItems = items
+        var queryItems = urlComponents.queryItems ?? []
+        for param in parameters {
+            if let index = queryItems.firstIndex(where: { $0.name == param.key }) {
+                queryItems[index] = URLQueryItem(name: param.key, value: param.value)
+            } else {
+                queryItems.append(URLQueryItem(name: param.key, value: param.value))
+            }
+        }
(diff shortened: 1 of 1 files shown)


Sources/SwifterSwift/Foundation/URLExtensions.swift
```diff
@@ -45,9 +45,15 @@ public extension URL {
...
-        urlComponents.queryItems = items
+        var queryItems = urlComponents.queryItems ?? []
+        for param in parameters {
+            if let index = queryItems.firstIndex(where: { $0.name == param.key }) {
+                queryItems[index] = URLQueryItem(name: param.key, value: param.value)
+            } else {
+                queryItems.append(URLQueryItem(name: param.key, value: param.value))
+            }
+        }
+        urlComponents.queryItems = queryItems
```
guykogus, line 56: This isn't correct, it's a separate function `settingQueryParameters`. Or maybe you want to add another parameter, `replaceExisting: Bool = false`. But changing this functionality is not correct, query parameters can have duplicate entries.

Sources/SwifterSwift/Foundation/URLExtensions.swift
```diff
@@ -101,6 +107,27 @@ public extension URL {
...
+        return urlComponents.url!
+    }
+    
+	/// SwifterSwift: Remove all the query parameters from the URL.
+	///
+	///        var url = URL(string: "https://google.com?code=12345")!
+	///        url.deleteAllQueryParameters()
+	///        print(url) // prints "https://google.com"
+    mutating func deleteAllQueryParameters() {
+        self = deletingAllQueryParameters(parameters)
```
guykogus, line 128: One of the conditions of SwifterSwift is that you can't use call our own functions, they should all be easily copy-pasted individually. It's not the most efficient way, but it suits how most people use this library.
```suggestion
        var urlComponents = URLComponents(url: self, resolvingAgainstBaseURL: true)!
        urlComponents.queryItems = nil
        self = urlComponents.url!
```

guykogus (changes requested):
Thanks for your PR. Please add tests and a CHANGELOG entry.

PR #1197, opened 2024-06-11
Fix compilation for Xcode 16

guykogus:
Also includes SwiftLint fixes

## Checklist


guykogus:
@LucianoPAlmeida maybe I'll separate out the SwiftLint fixes. Will get to it when I can.

guykogus:
@LucianoPAlmeida WDYT? I'll leave Swift 6 fixes for the next PR

SwifterSwift.xcodeproj/project.pbxproj
```diff
@@ -3572,7 +3566,6 @@
 				GCC_WARN_UNUSED_VARIABLE = YES;
 				INFOPLIST_FILE = "$(SRCROOT)/Sources/Info.plist";
 				INSTALL_PATH = "$(LOCAL_LIBRARY_DIR)/Frameworks";
-				IPHONEOS_DEPLOYMENT_TARGET = 10.0;
```
LucianoPAlmeida, line 3575: I think it may be time to up the minimum to 12 or 13 maybe? But that is something for another PR.

LucianoPAlmeida (approved):

SwifterSwift.xcodeproj/project.pbxproj
guykogus: Is there any reason to do it? I don't think the new Xcode complains about the minimum being 12.

LucianoPAlmeida: It doesn't complain but I think right now our minimum is 10, so it feels a good time up it to 12 or 13

guykogus: We're on iOS 12, which I think is fine for now.

PR #1199, opened 2024-06-24
Add new toggle Function to Set Extension

weihas:
## Checklist

### DIFF
--- a/Sources/SwifterSwift/SwiftStdlib/SetExtensions.swift
+++ b/Sources/SwifterSwift/SwiftStdlib/SetExtensions.swift
@@ -1,0 +1,20 @@
+// SetExtensions.swift - Copyright 2024 SwifterSwift
+
+public extension Set {
+    
+    /// SwifterSwift: Toggles an element in the set. If the element is present, it removes it; otherwise, it inserts it.
+    ///
+    ///     var set: Set<T> = [T1, T2, T3]
+    ///     set.toggle(T1) -> [T2, T3]
+    ///     set.toggle(T4) -> [T2, T3, T4]
+    ///     set.toggle(T1) -> [T1, T2, T3, T4]
+    ///
+    /// - Parameter element: The element to toggle in or out of the set.
+    mutating func toggle(_ element: Element) {
+        if contains(element) {
+            self.remove(element)
+        } else {
+            self.insert(element)
+        }
+    }
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,9 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 
 ## Upcoming Release
 ### Added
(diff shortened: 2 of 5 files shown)


Sources/SwifterSwift/SwiftStdlib/SetExtensions.swift
```diff
@@ -0,0 +1,20 @@
+// SetExtensions.swift - Copyright 2024 SwifterSwift
+
+public extension Set {
+    
+    /// SwifterSwift: Toggles an element in the set. If the element is present, it removes it; otherwise, it inserts it.
```
LucianoPAlmeida, line 5: Can you elaborate on a practical use of such method? I am not sure most people would have a valid case for using such a method. And I don't think this is a common operation on  sets such as [union, intersection, disjoint, ...](https://developer.apple.com/documentation/swift/setalgebra).
Also, the naming `toggle` does not communicate well what the method does, I had to read the documentation.

LucianoPAlmeida (commented):
Hey @weihas 
Thank you for taking the time for making a contribution, I left a comment and IMO this would not be a fit for introducing into the library. But up to discussion with others cc @guykogus 
But if you still interested in contributing see the list of issues to find areas where the project needs contributions :)

Sources/SwifterSwift/SwiftStdlib/SetExtensions.swift
weihas: Sometimes there is a collection to collect certain values, when the user clicks, if it is not added in it, it will be added, otherwise it will be deleted, just like a switch or toggle. For example, in many scenes of adding tags.
 I admit that it is indeed a little far-fetched and the name is a little difficult to understand. Maybe it's just that only I have used this logic many times. But thanks for your review!  :)

guykogus: I've never felt a need to use this myself, but it kinda makes sense to have. Although normally, especially if you're referring to responding to a switch, you would check the state of the UI and then insert/remove the element. E.g.
```swift
@objc func toggle(_ sender: UISwitch) {
    let element = mappedElement(for: sender.tag)
    if sender.isOn {
        selectedElements.insert(element)
    } else {
        selectedElements.remove(element)
    }
}
```

The only thing I'm wondering is if this should be added to `SetAlgebra` instead of just `Set`?

LucianoPAlmeida: I don't think it would make sense in 'SetAlgebra' as I think it represents more the mathematical properties (operations) of a Set which is not the case for this one IMO.

LucianoPAlmeida: I would also like a better naming, but don't have any suggestion at the moment.

PR #1206, opened 2024-08-19
Added an initializer o create a `UIColor` instance from a hexadecimal…

herlandro:
… string representation

## Checklist


Sources/SwifterSwift/UIKit/UIColorExtensions.swift
```diff
@@ -18,6 +18,64 @@ public extension UIColor {
...
+    /// in the format `#RRGGBB`. The alpha component is ignored.
+    ///
+    /// - Returns: A string representing the color in hexadecimal format, prefixed with `#`.
+    ///
+    /// Example:
+    /// ```
+    /// let color = UIColor.red
+    /// print(color.hexString) // Output: "#ff0000"
+    /// ```
+    var hexString: String {
```
LucianoPAlmeida, line 34: Doesn't this already exists on the project?https://github.com/SwifterSwift/SwifterSwift/blob/3b2a1725e967498edadb84e6271bb85430c1961c/Sources/SwifterSwift/Shared/ColorExtensions.swift#L91

LucianoPAlmeida (commented):
Also, can Unit tests be added as well?

Sources/SwifterSwift/UIKit/UIColorExtensions.swift
herlandro: Your are right. When I was checking on the project I only rechearched inside the UIColor extension, but his class that you mention is a ealias `public typealias SFColor = UIColor`. So in this project has 2 places to manage UIColor extension:
- Source/SwifterSwift/UIkit/UIColorExtensions
- Source/SwifterSwift/Shared/ColorExtensions

In this case, the var that I added is not necessary, I will remove.

herlandro: Also I will add tests cases

herlandro:
Hey Luciano,
The code has been refactored and the test cases are also ready.
Can you review?

Sources/SwifterSwift/UIKit/UIColorExtensions.swift
```diff
@@ -3,6 +3,30 @@
...
+/// SwifterSwift: Is an enumeration that represents potential errors related to converting hexadecimal strings to colors.
+///
+/// This enumeration is used to capture and report errors that may occur when attempting to initialize a color using a hexadecimal string.
+/// It handles cases where the provided string either does not have the correct length or contains invalid characters.
+///
+/// - Case `invalidHexLength`: Indicates that the provided hexadecimal string does not have exactly 6 characters,
+///   which is required for a valid conversion.
+/// - Case `invalidHexFormat`: Indicates that the provided hexadecimal string contains characters that are not valid
+///   hexadecimal digits (0-9, A-F).
+enum ColorError: Error {
```
LucianoPAlmeida, line 15: I know it is against the norm, but here on this project we try to keep extensions self contained, so we don't introduce new types.

Sources/SwifterSwift/UIKit/UIColorExtensions.swift
```diff
@@ -18,6 +42,53 @@ public extension UIColor {
...
+    /// ```
+    /// let color = UIColor(hex: "#ff0000")
+    /// // Creates a red color
+    /// ```
+    ///
+    /// If the input string does not contain exactly 6 characters, the initializer will trigger a runtime
+    /// error with a precondition failure.
+    ///
+    /// - Note: The alpha component is set to 1.0 (fully opaque) by default.
+    convenience init(hex: String) throws {
```
LucianoPAlmeida, line 64: We can't make it throwing, so let's make it a failable init and return `nil` in case of error
```suggestion
    convenience init?(hex: String) {
```

herlandro: Thanks for your review. I refactored the function and the test cases

guykogus:
@herlandro [this initialiser already exists](https://github.com/SwifterSwift/SwifterSwift/blob/master/Sources/SwifterSwift/Shared/ColorExtensions.swift#L308), so we don't need it. I'll close this PR, but if you see a reason to reopen it feel free to do so.

Thank you for your effort, and please share any more extensions you'd like to add 🚀

LucianoPAlmeida:
> @herlandro [this initialiser already exists](https://github.com/SwifterSwift/SwifterSwift/blob/master/Sources/SwifterSwift/Shared/ColorExtensions.swift#L308), so we don't need it. I'll close this PR, but if you see a reason to reopen it feel free to do so.
> 
> 
> 
> Thank you for your effort, and please share any more extensions you'd like to add 🚀

Oh we have that already as well? I think I may have missed it. But @herlandro there is a lot of other good extensions and maybe improvements on existing ones that can be done. Take a look at our issues and feel free to contribute to this project :)

PR #1209, opened 2024-09-06
Added extension that converts an array into a dictionary using a specified key

herlandro:
## Checklist

- [ ] New extensions are written in Swift 5.6.
- [ ] New extensions support iOS 12.0+ / tvOS 12.0+ / macOS 10.13+ / watchOS 4.0+, or use `@available` if not.


Tests/SwiftStdlibTests/ArrayExtensionsTests.swift
```diff
@@ -144,4 +144,39 @@ final class ArrayExtensionsTests: XCTestCase {
...
+    }
+
+    // Test converting an array of integers into a dictionary using modulo as the key.
+    func testToDictionaryByModulo() {
+        let numbers = [1, 2, 3, 4, 5]
+        let result = numbers.toDictionary { $0 % 2 == 0 ? "even" : "odd" }
+
+        let expected: [String: Int] = ["odd": 5, "even": 4]
+        XCTAssertEqual(result, expected)
+    }
```
guykogus, line 164: This example should show you what's wrong with your approach. There are several odd and even  numbers, and you can get the whole list of them using `Dictionary.init(grouping:by:)`. I.e.
```swift
// Returns `["odd": [1, 3, 5], "even": [2, 4]]`
[String: [Int]](grouping: numbers, by: { $0 % 2 == 0 ? "even" : "odd" })
```
I think that the currently-existing initialiser is all that we need, and not the version that you've created. Do you have a better reason to only keep the _last_ element of these arrays?

guykogus (changes requested):
Please clean up your PR. It contains code from other PRs, plus it doesn't look like you've read the contribution guideline.

PR #1217, opened 2024-10-17
StringExtensions // Add all available parameters to String.localized().

ShikiSuen:
Closing #1217 :

The String localization APIs in this package needs new parameters to specify the bundle and the tableName.

https://github.com/SwifterSwift/SwifterSwift/blob/8ea787afccfe431c81e0c3fa5f1f304e5dff831a/Sources/SwifterSwift/SwiftStdlib/StringExtensions.swift#L524-L527

Otherwise, these APIs will be totally useless if for accessing String catalogs bundled inside an SPM.

**Describe the solution you'd like**
Add new parameters to these APIs to allow users specify the bundle and the tableName.

## Checklist

- [ ] I have added tests for new extensions, and they passed. (Test not needed in this case.)


ShikiSuen:
I rebased the previous commit with some forced pushes. Now it is at `4029740`.

Further changes will be pushed as new dedicated commits instead.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 
 ## Upcoming Release
 ### Added
+  - `localized()` now supports all parameters provided by `NSLocalizedString()`. [#1217](https://github.com/SwifterSwift/SwifterSwift/pull/1217) by [Shiki Suen](https://github.com/ShikiSuen)
```
guykogus, line 7: This isn't `Added`, it's `Changed`.

Sources/SwifterSwift/SwiftStdlib/StringExtensions.swift
```diff
@@ -514,15 +514,52 @@ public extension String {
...
+    /// found in the table.
+    ///   - comment: The comment to place above the key-value pair in the strings file. This parameter provides
+    /// the translator with some context about the localized string’s presentation to the user.
+    /// - Returns: Localized string. Please refer to the Xcode documentation of `NSLocalizedString()` API for details.
+    func localized(
+        tableName: String? = nil,
+        bundle: Bundle = Bundle.main,
+        value: String = "",
+        comment: String = "") -> String {
+        /// TODO: Use the following API instead (plus the `locale` parameter )
```
guykogus, line 547: Remove this TODO. If you want to add another function for this please create a new PR, and you can use `@available` to support it properly.

Sources/SwifterSwift/SwiftStdlib/StringExtensions.swift
```diff
@@ -514,15 +514,52 @@ public extension String {
     }
     #endif
 
+    #if canImport(Foundation)
+    /// SwifterSwift: Returns a localized string.
+    ///
+    ///        "Hello world".localized -> Hallo Welt
+    ///
+    var localized: String {
```
guykogus, line 522: Please remove this, it's unnecessary. Since you provide default parameter values callers can just run `.localized()` directly.

guykogus (changes requested):

Sources/SwifterSwift/SwiftStdlib/StringExtensions.swift
ShikiSuen: If removing it, Swift will force users to add the `()`.

LucianoPAlmeida (approved):

CHANGELOG.md
```diff
@@ -5,6 +5,9 @@ The changelog for **SwifterSwift**. Also see the [releases](https://github.com/S
 ## Upcoming Release
 ### Added
 
+### Changed
+  - `localized()` now supports all parameters provided by `NSLocalizedString()`. [#1217](https://github.com/SwifterSwift/SwifterSwift/pull/1217) by [Shiki Suen](https://github.com/ShikiSuen)
```
guykogus, line 9:
```suggestion
  - `localized(comment:)` now supports all parameters provided by `NSLocalizedString(tableName:bundle:value:comment:)`. [#1217](https://github.com/SwifterSwift/SwifterSwift/pull/1217) by [Shiki Suen](https://github.com/ShikiSuen)
```

guykogus (changes requested):
One last suggestion

ShikiSuen:
Feel free to ping me to update the API as soon as when this project drops support for macOS 12 and earlier.

guykogus (changes requested):
@ShikiSuen apologies, I hadn't updated the CHANGELOG for the new release. Will you please update to it.

guykogus:
> Feel free to ping me to update the API as soon as when this project drops support for macOS 12 and earlier.

Why do we need to drop support for macOS 12? What will this give you?

ShikiSuen:
@guykogus Your modification `5a6de91` hinders this repo from being mergable into the main.
This is troublesome especially the reviewing process is completed.
I'm not the trouble maker so I'll leave this PR here as is. You are free to make further changes if you like.

ShikiSuen:
> Why do we need to drop support for macOS 12? What will this give you?

image

The TODO I added previously is to not use the ObjC-backed-up NSLocalizedString() but a newer Swift-native API.

guykogus (approved):

guykogus:
> > Why do we need to drop support for macOS 12? What will this give you?
> 
> The TODO I added previously is to not use the ObjC-backed-up NSLocalizedString() but a newer Swift-native API.
> 
> This API does not support macOS 12. Also, it has a new parameter `locale` which NSLocalizedString does not have.

@ShikiSuen if you want to make functions available for versions greater than the minimum, you can just do it like this.
```swift
public extension String {
    @available(macOS 13, iOS 16, tvOS 16, watchOS 9, *)
    func localized(locale: Locale = .current,
                   options: LocalizationOptions = LocalizationOptions()) -> String {
        var localizable = LocalizedStringResource(stringLiteral: self)
        localizable.locale = locale
        return String(localized: localizable, options: options)
    }
}

@available(iOS 15, tvOS 15, watchOS 8, macOS 12, *)
public extension String.LocalizationValue {
    func localized(table: String? = nil,
                   bundle: Bundle? = nil,
                   locale: Locale = .current,
                   comment: StaticString? = nil) -> String {
        String(localized: self,
               table: table,
               bundle: bundle,
               locale: .current,
               comment: comment)
    }
}

public extension StaticString {
    @available(iOS 15, tvOS 15, watchOS 8, macOS 12, *)
    func localized(table: String? = nil,
                   bundle: Bundle? = nil,
                   locale: Locale = .current,
                   comment: StaticString? = nil) -> String {
        String(localized: self,
               defaultValue: "",
               table: table,
               bundle: bundle,
               locale: .current,
               comment: comment)
    }
}
```
I personally don't see much value in the `String.LocalizationValue` and `StaticString`, since the functions have default values anyway and you can just call them directly. But maybe the one using `LocalizedStringResource` could be useful... although probably not, as I think Xcode 16 can use these calls to automatically update the String Catalog. In fact I don't even like the original `localized()` function we have because `genstrings` looks for calls to `NSLocalizedString`, and by not using it directly the callers lose that tool.

PR #1227, opened 2024-12-09
fix datetime return nil without Locale and Timezone

ws00801526:
## Checklist

### DIFF
--- a/Sources/SwifterSwift/SwiftStdlib/StringExtensions.swift
+++ b/Sources/SwifterSwift/SwiftStdlib/StringExtensions.swift
@@ -327,7 +327,8 @@ public extension String {
     var date: Date? {
         let selfLowercased = trimmingCharacters(in: .whitespacesAndNewlines).lowercased()
         let formatter = DateFormatter()
-        formatter.timeZone = TimeZone.current
+        formatter.locale = .init(identifier: Locale.current.identifier)
+        formatter.timeZone = .init(identifier: Locale.current.identifier)
         formatter.dateFormat = "yyyy-MM-dd"
         return formatter.date(from: selfLowercased)
     }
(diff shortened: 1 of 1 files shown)


guykogus:
@ws00801526 what scenarios fail? Can you please add tests that show the current implementation failing but the new implementation working?

ws00801526:
> @ws00801526 what scenarios fail? Can you please add tests that show the current implementation failing but the new implementation working?

This bug can be reproduced  on the mac with Intel clip
Xcode: 16.1
macOS: Sonoma 14.7.1 
Simlutor: iPhone 16 Pro with iOS 18.1 

screen_shot_2
screen_shot_1

guykogus:
@ws00801526 what timezone are you in that causes this bug? I just ran the test on all known identifiers by changing the function:
```swift
func dateTime(timeZone: TimeZone) -> Date? {
    let selfLowercased = trimmingCharacters(in: .whitespacesAndNewlines).lowercased()
    let formatter = DateFormatter()
    formatter.timeZone = timeZone
    formatter.dateFormat = "yyyy-MM-dd HH:mm:ss"
    return formatter.date(from: selfLowercased)
}
```
and then updated the test like this
```swift
func testDateTime() {
    for timeZone in TimeZone.knownTimeZoneIdentifiers {
        let dateFromStr = "2015-06-01 14:23:09".dateTime(timeZone: TimeZone(identifier: timeZone)!)
        XCTAssertNotNil(dateFromStr, "Failed to get date time for time zone \(timeZone)")
        XCTAssertEqual(dateFromStr?.year, 2015)
        XCTAssertEqual(dateFromStr?.month, 6)
        XCTAssertEqual(dateFromStr?.day, 1, "Incorrect day for time zone \(timeZone)")
        XCTAssertEqual(dateFromStr?.hour, 14, "Incorrect hour for time zone \(timeZone)")
        XCTAssertEqual(dateFromStr?.minute, 23, "Incorrect minute for time zone \(timeZone)")
        XCTAssertEqual(dateFromStr?.second, 9)
    }
}
```
and I got a lot of errors for day/hour/minute (which is to be expected as the test values are expecting the current local timezone) but none of the responses came back `nil`.

Also, your code is wrong, calling `formatter.timeZone = .init(identifier: Locale.current.identifier)` doesn't make sense as you're creating a `TimeZone` object using a locale identifier. I just tested this myself and it trying to do this just returns the local current timezone, not whatever you were trying to get.

The truth is that these functions are not correct, I wouldn't recommend for anyone to really use them. Playing with calendars, dates, locales, timezones, formatters... it's not easy, and clearly using functions like this that mask that complexity leads to problems because there are so many edge cases.

